PR #22, opened 2019-06-12
update GA based on upstream PR

xlc:


xlc:
@ianhe8x there are some cennzx changes, can you have a look

crml/cennzx-spot/src/lib.rs
```diff
@@ -675,7 +677,11 @@ impl<T: Trait> Module<T> {
 		let denominator = output_reserve - output_amount;
 		let output: T::Balance = numerator / denominator + One::one();
 
-		(FeeRate::one() + fee_rate) * output
+		let output: T::AsBalance = output.into();
+		let output: u128 = output.as_();
+		let result = (FeeRate::one() + fee_rate) * output;
+		let result: T::AsBalance = As::sa(result);
+		Into::<T::Balance>::into(result)
```
ianhe8x: is this like a safe mul? not understand the hardcode u128 part.

xlc: This is same behavior as the old code, but because we are dropping `As<u128>` from `GA::Balance`, we need some workarounds here.
The proper solution is make FeeRate not needing `As<u128>` anymore but it is too much for this PR.

xlc:
This shouldn't change metadata (other the doc string part) as no externally used types are changed.

ianhe8x:
> This shouldn't change metadata (other the doc string part) as no externally used types are changed.

sometimes the way to define types changed, typename in metadata won't be the same as before

PR #32, opened 2019-06-24
Build and Publish Docker Image

cam-cen:
A few changes
- builds our now set to yesterday's rust-nightly (to ensure there'd always be a rust nightly version)
- docker images will be built on any commit to master
- docker images now use the `cennznet` user instead of root


cam-cen:
ready for merge @xlc

xlc:
do we always want to use latest nightly or stick with a known working one?

.circleci/config.yml
```diff
@@ -18,21 +23,23 @@ jobs:
               curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
             fi
             source $HOME/.cargo/env
-            rustup default nightly
-            rustup update nightly
-            rustup target add wasm32-unknown-unknown --toolchain nightly
+            rustup toolchain update nightly
+            rustup toolchain install nightly-"$(cat NIGHTLY_DATE)"
```
xlc: why update and then install?
can you `rustup toolchain update nightly` and figure out the latest nightly got installed from update command?

.circleci/config.yml
```diff
@@ -18,21 +23,23 @@ jobs:
...
-            rustup component add rustfmt --toolchain=nightly
+            rustup target add x86_64-unknown-linux-musl --toolchain=nightly-"$(cat NIGHTLY_DATE)"
+            rustup component add rustfmt --toolchain=nightly-"$(cat NIGHTLY_DATE)"
             rustc --version; cargo --version; rustup --version
       - run:
           name: Check format
           command: |
             source $HOME/.cargo/env
-            cargo +nightly fmt
+            cargo fmt
```
xlc:
```suggestion
            cargo fmt --all -- --check
```

.circleci/config.yml
cam-cen: good find. I thought there was a bug in rustup where it was missing part of the toolchain, but upon reflection this extra update is unnecessary

cam-cen:
> do we always want to use latest nightly or stick with a known working one?
until we can default to 1.36 we have to use nightly
```
   Compiling cennznet-primitives v0.1.0 (/home/circleci/project/primitives)
warning: the feature `alloc` has been stable since 1.36.0 and no longer requires an attribute to enable
  --> /home/circleci/project/primitives/src/lib.rs:20:43
   |
20 | #![cfg_attr(not(feature = "std"), feature(alloc))]
   |                                           ^^^^^
   |
   = note: #[warn(stable_features)] on by default
```

xlc:
nightly is required only for wasm build. can you hardcoded a nightly version and we can manually bump it as needed

.circleci/config.yml
```diff
@@ -3,12 +3,14 @@ jobs:
...
       image: ubuntu-1604:201903-01
+    environment:
+      RUST_VERSION: nightly-2019-07-14
     steps:
       - checkout
       - restore_cache:
           keys:
-            - v1-cargo-global-cache-{{ arch }}-{{ checksum "Cargo.lock" }}
-            - v1-cargo-global-cache-
+            - v1-cargo-global-cache-{{ arch }}-{{ checksum "Cargo.lock" }}-{{ checksum ".circleci/RUST_VERSION" }}
```
xlc: is it possible to have this using environment variable `$RUST_VERSION` to avoid duplicated nightly version?

cam-cen: unfortunately not, I tried that

cam-cen:
thanks, yeah unfortunately its not possible, I could set it as a job environment variable in the circleci dashboard, but then there'd be no visibility at all where it is. If it was part of the Cargo file then we could just cache it on that https://circleci.com/docs/2.0/caching/#using-keys-and-templates

.circleci/config.yml
xlc: also not possible to load env from file?

PR #50, opened 2019-07-19
Handle overflow error in add/remove liquidity

KarishmaBothara:


ianhe8x:
this u128 -> u256 treatment bloat codes, need some way to keep readability.
how about create some safe math util functions (solidity styled)? @xlc

xlc:
There is `from_rational_approximation` introduced in substrate 2.0
We can refactor all those code once substrate 2.0 is merged in

https://github.com/paritytech/substrate/blob/722e6d93125fa12813bd4bb767a653d20eaf2c87/core/sr-primitives/src/lib.rs#L196

PR #73, opened 2019-11-12
Fix cennznet primitive types

cowboy-bebug:
- use consistent primitive types for transaction index
- add AssetId type to be used in generic-asset module

(to be updated with Wilfred's response)

### DIFF
--- a/executor/src/lib.rs
+++ b/executor/src/lib.rs
@@ -36,7 +36,7 @@ native_executor_instance!(
 #[cfg(test)]
 mod tests {
 	use super::Executor;
-	use cennznet_primitives::{Balance, BlockNumber, Hash};
+	use cennznet_primitives::{Balance, BlockNumber, Hash, Index, Timestamp};
 	use cennznet_runtime::{
 		constants::currency::*, impls::WeightToFee, Balances, Block, BuildStorage, Call, CheckedExtrinsic, Event,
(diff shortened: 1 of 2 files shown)


jordy25519 (approved):
Re the offline discussion of `u64` vs `u32` If we had 100,000,000 accounts * 4 bytes.
The space savings are quite marginal
so LGTM!

executor/src/lib.rs
```diff
@@ -463,7 +463,7 @@ mod tests {
 		(block1, block2)
 	}
 
-	fn block_with_size(time: u64, nonce: u32, size: usize) -> (Vec<u8>, Hash) {
+	fn block_with_size(time: u64, nonce: u64, size: usize) -> (Vec<u8>, Hash) {
```
aliXsed: how about? 
```suggestion
        use cennznet_primitives::{Index, Timestamp};
	fn block_with_size(time: Timestamp, nonce: Index, size: usize) -> (Vec<u8>, Hash) {
```
I'm suggesting it so from now on if there would be any need to change the type of Index, nonce follows it automatically.

cowboy-bebug: Ah yep sounds good! I'll make the change

cowboy-bebug:
Sorry I screwed up with interactive rebase lol I'll merge this as soon as it's ready

mzxyz:
LGTM

PR #96, opened 2019-12-10
Optimise circleci

cowboy-bebug:
Changes:
- divide building binary (only runs `cargo check`) and tests into jobs (so that they can run in parallel in a workflow)
- refactor inline `run` into `commands`
- added `build-docker-and-publish` job to deploy with `dev` tag for now
- added `build` job to make ci pass (circleci requires `build` job to pass)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,7 +144,10 @@ workflows:
     jobs:
       - build-bin
       - build-test-and-run
-      - build-docker-and-publish
+      - build-docker-and-publish:
+          filters:
+            branches:
+              only: master
       - build:
           requires:
             - build-bin
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ jobs:
     machine:
       image: ubuntu-1604:201903-01
     steps:
-      - run: echo "successfully build and tested"
+      - run: echo "successfully built and tested"
   build-bin:
     machine:
       image: ubuntu-1604:201903-01
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,11 @@ commands:
           no_output_timeout: 60m
 
 jobs:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -59,9 +59,42 @@ commands:
...
+          command: |
+            cargo test --all
+  docker-build-and-publish:
+    steps:
+      - run:
+          name: Build and publish Docker image
+          command: |
+            docker login -u $DOCKER_LOGIN -p $DOCKER_PASSWORD
+            docker build --pull -t cennznet/cennznet:dev -f docker/Dockerfile .
+            docker push cennznet/cennznet:dev
```
jordy25519: Is it possible to use `{{ .Environment.CIRCLE_BRANCH }}` as the tag?

.circleci/config.yml
```diff
@@ -59,9 +59,42 @@ commands:
...
+          no_output_timeout: 60m
 
 jobs:
   build:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - run: echo "successfully build and tested"
```
jordy25519: successfully buil**t**

jordy25519 (approved):
TY this is great :tada: 
Some small things but TIOLI

.circleci/config.yml
cowboy-bebug: yeah actually let me add back DOCKER_TAG stuff with branch name

.circleci/config.yml
```diff
@@ -147,7 +147,9 @@ workflows:
       - build-docker-and-publish:
           filters:
             branches:
-              only: master
+              only:
+                - develop
+                - /^[0-9]+[.][0-9]+[.][0-9](-rc.+)?$/
```
jordy25519, line 156: 🔥

PR #101, opened 2019-12-16
Use standard types and checked methods for large integer calcs in Cennzx

### LINKED ISSUE #59: Use substrate support of large integers for CENNZX and refine calculations
tracker for: https://github.com/cennznet/cennznet/blob/ec299bfb0d5230ac2388ef021ee129205f8b8d4a/crml/cennzx-spot/src/lib.rs#L37

arithmetic crate: https://github.com/plugblockchain/plug-blockchain/tree/1.0.0-rc1/core/sr-arithmetic

Depends on: #60

aliXsed:
Closes #59


crml/cennzx-spot/src/tests.rs
```diff
@@ -834,15 +834,15 @@ fn get_input_price_for_max_reserve_balance() {
...
 			CennzXSpot::get_input_price(
-				u128::max_value(),
-				u128::max_value(),
-				u128::max_value(),
+				LowPrecisionUnsigned::max_value(),
+				LowPrecisionUnsigned::max_value(),
+				LowPrecisionUnsigned::max_value(),
 				DefaultFeeRate::get()
 			),
 			"Overflow error"
```
aliXsed: @holygits @elim0322 looks like after refining the calculations we can, in fact, avoid the overflow in this test and return Ok(169886353929574869427545984738775941814) which is probably correct. So I suspect this test itself is not valid. This is the only test that fails at the moment.

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,157 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
-//!
-//! CENNZX-SPOT Types
-//!
```
cowboy-bebug: could u add back this module doc comment?

crml/cennzx-spot/src/tests.rs
aliXsed: Another possible scenario that I'm considering is that we have a silent saturation happening since I have used saturating_mul for high precision calculations. The reason for doing that was high precision integers have already got double space, so I assumed u128 numbers are not capable of saturating u256 results! it could then simplify the code to use saturating operations as they have a simpler interface. However, I may need to re-evaluate this assumption. What do you think?

aliXsed: Ok, back to my first comment. My assumption that saturating calculations for u128 values in the space of u256 is safe must be correct.

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,157 @@
...
+	fn scale() -> LowPrecisionUnsigned {
+		1_000
+	}
+}
 
-/// FeeRate S.F precision
-const SCALE_FACTOR: u128 = 1_000_000;
+/// Per hundredth of unit price
+#[derive(Debug, Clone, Copy)]
+pub enum Percent {}
```
jordy25519: It's not pretty but I'd recommend `PerCent` for consistency

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,157 @@
...
+use core::{
+	convert::{From, Into, TryFrom},
+	marker::PhantomData,
+};
+
+pub use primitive_types::U256 as HighPrecisionUnsigned;
+pub use u128 as LowPrecisionUnsigned;
+
+pub trait Scaled {
+	fn scale() -> LowPrecisionUnsigned;
```
jordy25519: Could use an associated constant here: `const SCALE: LowPrecisionUnsigned;`

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,157 @@
...
-		let rhs = U256::from(rhs.into());
-		let scale_factor = U256::from(SCALE_FACTOR);
-		let lhs = U256::from(lhs.0);
-		let res: u128 = (lhs * rhs / scale_factor).try_into().map_err(|_| "Overflow error")?;
+impl<Unit> Eq for FeeRate<Unit> {}
 
-		Ok(res)
+impl<Unit> TryFrom<HighPrecisionUnsigned> for FeeRate<Unit> {
+	type Error = Error;
+	fn try_from(h: HighPrecisionUnsigned) -> Result<Self, Error> {
```
jordy25519: `Result<Self, Self::Error>`, same for the other declarations

jordy25519 (commented):
There's lots here so this is an initial first pass review

`FeeRate` with generic scale is a nice improvement 👍 

Suggestions:
- Should use`decl_error!` to make the error an enum: https://github.com/plugblockchain/plug-blockchain/blob/0966e9d414e018f7a1ab85b6ed145eb78af590a6/frame/support/src/error.rs#L34-L43
- Recommend implementing `::new` for `FeeRate` to avoid initialization with `PhantomData`
- No need to name unit tests with `test_<something>`
- Scale / Unit seem to be used interchangeably one term could be preferred e.g. `impl<S: Scale> for FeeRate<S>`

crml/cennzx-spot/src/lib.rs
```diff
@@ -657,59 +657,73 @@ impl<T: Trait> Module<T> {
...
 
-		let input_amount_less_fee_scaled = FeeRate::safe_div(lhs, div_rate)?;
-		let input_reserve: u128 = T::BalanceToU128::from(input_reserve).into();
-		let output_reserve = U256::from(T::BalanceToU128::from(output_reserve).into());
-		let input_amount = U256::from(input_amount_less_fee_scaled);
+		let input_amount_scaled = FeeRate::<PerMill>::from(T::BalanceToUnsignedInt::from(input_amount).into())
+			.checked_div(div_rate)
+			.ok_or::<&'static str>(Error::DivideByZero.into())?;
 
-		let denominator: u128 = (input_amount + U256::from(input_reserve))
```
jordy25519, line 704: This would overflow previously as we try to convert `denominator` back into `u128`. Your changes keep the high precision up until the final price calculation which I think is a good improvement 🥇

crml/cennzx-spot/src/tests.rs
jordy25519: LGTM

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,157 @@
...
+pub trait Scaled {
+	fn scale() -> LowPrecisionUnsigned;
 }
 
-construct_uint! {
-	/// 256-bit unsigned integer.
-	pub struct U256(4);
+/// Per millionth of unit price
+#[derive(Debug, Clone, Copy)]
+pub enum PerMill {}
```
jordy25519, line 36: 💭 was there a reason to prefer `enum PerMill {}` over a unit struct `struct PerMill{}`?

crml/cennzx-spot/src/lib.rs
```diff
@@ -51,8 +51,8 @@ pub trait Trait: system::Trait + generic_asset::Trait {
 	type Event: From<Event<Self>> + Into<<Self as system::Trait>::Event>;
 	/// A function type to get an exchange address given the asset ID pair.
 	type ExchangeAddressGenerator: ExchangeAddressFor<Self::AssetId, Self::AccountId>;
-	type BalanceToU128: From<<Self as generic_asset::Trait>::Balance> + Into<u128>;
-	type U128ToBalance: From<u128> + Into<<Self as generic_asset::Trait>::Balance>;
+	type BalanceToUnsignedInt: From<<Self as generic_asset::Trait>::Balance> + Into<LowPrecisionUnsigned>;
```
jordy25519, line 54: It still bugs me that we need to have two types when they both bind to `Balance` ¯\_(ツ)_/¯

jordy25519 (commented):
Final review pass. The changes are really good TY.
Just the small style / quality things then we're good to go

crml/cennzx-spot/src/lib.rs
aliXsed: Me too. I had a go at it. But it was at the time I had lots of compile errors and I needed to withdraw from some of my changes. I will have a second look at it to check if the solution is really simple. Otherwise, we may need to stand it until a later opportunity.

crml/cennzx-spot/src/types.rs
aliXsed: I took the solution from the Rust by examples. Not sure. Just guessing under the hood it may be more efficient.

aliXsed:
> There's lots here so this is an initial first pass review
> 
> `FeeRate` with generic scale is a nice improvement 👍
> 
> Suggestions:
> 
> * Should use`decl_error!` to make the error an enum: https://github.com/plugblockchain/plug-blockchain/blob/0966e9d414e018f7a1ab85b6ed145eb78af590a6/frame/support/src/error.rs#L34-L43
> * Recommend implementing `::new` for `FeeRate` to avoid initialization with `PhantomData`
> * No need to name unit tests with `test_<something>`
> * Scale / Unit seem to be used interchangeably one term could be preferred e.g. `impl<S: Scale> for FeeRate<S>`

FeeRate::<Unit>::from is doing exactly the same job as what we may define for "new". Should I instead use from in more places then?

jordy25519:
> FeeRate::::from is doing exactly the same job as what we may define for "new". Should I instead use from in more places then?

That works, no need to pass `PhantomData` each time is all

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
-	pub struct U128(2);
+use codec::{Decode, Encode, HasCompact};
+use core::{
+	convert::{From, Into, TryFrom},
+	marker::PhantomData,
+};
+use support::decl_error;
+
+pub use primitive_types::U256 as HighPrecisionUnsigned;
+pub use u128 as LowPrecisionUnsigned;
```
cowboy-bebug, line 27: These aliases are confusing to me because there's no reference to the actual types u128 and U256 being used. Is there a reason why we can't use the original names?

cowboy-bebug (changes requested):
LGTM 🥇I have a couple of comments for changes.
I'm sure it's formatted but could u make sure `cargo fmt`.

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
+/// Per millionth of unit price
+#[derive(Debug, Clone, Copy)]
+pub enum PerMill {}
+impl Scaled for PerMill {
+	const SCALE: LowPrecisionUnsigned = 1_000_000;
 }
 
-use codec::{Compact, CompactAs, Decode, Encode, HasCompact};
+/// Per thousandth of unit price
+#[derive(Debug, Clone, Copy)]
```
cowboy-bebug: could u change this to `#[derive(Debug, Clone, Copy, PartialEq)]`

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
+pub enum PerMilli {}
+impl Scaled for PerMilli {
+	const SCALE: LowPrecisionUnsigned = 1_000;
+}
+
+/// Per hundredth of unit price
+#[derive(Debug, Clone, Copy)]
```
cowboy-bebug: could u change this to `#[derive(Debug, Clone, Copy, PartialEq)]`

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
+/// A trait for values which hold an implicit scale factor that needs to be taken into account in calculations
+pub trait Scaled {
+	const SCALE: LowPrecisionUnsigned;
 }
 
-construct_uint! {
-	/// 256-bit unsigned integer.
-	pub struct U256(4);
+/// Per millionth of unit price
+#[derive(Debug, Clone, Copy)]
```
cowboy-bebug: could u change this to `#[derive(Debug, Clone, Copy, PartialEq)]`

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
+	}
+}
 
-/// FeeRate (based on Permill), uses a scale factor
-/// Inner type is `u128` in order to support compatibility with `generic_asset::Balance` type
-#[derive(Encode, Decode, Default, Copy, Clone, PartialEq, Eq, Debug)]
+/// Inner type is `LowPrecisionUnsigned` in order to support compatibility with `generic_asset::Balance` type
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct FeeRate(u128);
+#[derive(Encode, Decode, Copy, Clone, Debug)]
```
cowboy-bebug: could u change this to `#[derive(Encode, Decode, Copy, Clone, Debug, PartialEq)]`

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
+}
 
-	/// Create a FeeRate from a % i.e. `x / 100`
-	pub fn from_percent(x: u128) -> FeeRate {
-		FeeRate(x * SCALE_FACTOR / 100)
+impl<S: Scaled> PartialEq for FeeRate<S> {
+	fn eq(&self, other: &Self) -> bool {
+		self.0 == other.0
 	}
+}
```
cowboy-bebug: These lines can be removed because we derived `PartialEq` above

crml/cennzx-spot/src/types.rs
```diff
@@ -13,91 +13,144 @@
...
 
-	/// Divide a `u128` supported numeric by a FeeRate
-	pub fn safe_div<N: Into<u128>>(lhs: N, rhs: FeeRate) -> rstd::result::Result<u128, &'static str> {
-		let lhs = lhs.into();
-		let lhs = U256::from(lhs);
-		let scale_factor = U256::from(SCALE_FACTOR);
-		let rhs = U256::from(rhs.0);
-		let res: u128 = (lhs * scale_factor / rhs).try_into().map_err(|_| "Overflow error")?;
+impl<S: Scaled> Eq for FeeRate<S> {}
```
cowboy-bebug: This line should be removed because it turns out we don't need  / use`Eq` 😄

cowboy-bebug (changes requested):

crml/cennzx-spot/src/types.rs
jordy25519: Just found it's at the start of the rust book 🤦‍♂ https://doc.rust-lang.org/nomicon/exotic-sizes.html

crml/cennzx-spot/src/types.rs
aliXsed: Sure

crml/cennzx-spot/src/types.rs
aliXsed: If we decide in the future that our low precision threshold would be 256 which is the case for Etherium and a likely change for us, we shouldn't need to go and change many places of the code.

crml/cennzx-spot/src/types.rs
aliXsed: Why we would need PartialEq here?

crml/cennzx-spot/src/types.rs
aliXsed: I remember I tried deriving but couldn't pacify the compiler without explicitly implementing PartialEq

crml/cennzx-spot/src/types.rs
aliXsed: Nice notice!

crml/cennzx-spot/src/types.rs
cowboy-bebug: That's because `PartialEq` was not derived for the underlying structs (PerMill, PerMilli and PerCent). The error message actually tells u about this if I try to derive PartialEq just for FeeRate:
```
error[E0369]: binary operation `==` cannot be applied to type `(&types::FeeRate<types::PerMill>,)`
   --> crml/cennzx-spot/src/lib.rs:58:1
    |
58  |   decl_module! {
    |  _^
    | |_|
    | |
59  | |     pub struct Module<T: Trait> for enum Call where origin: T::Origin {
60  | |         fn deposit_event() = default;
61  | |
...   |
229 | |     }
230 | | }
    | | ^
    | |_|
    | |_(&types::FeeRate<types::PerMill>,)
    |   (&types::FeeRate<types::PerMill>,)
    |
    = note: an implementation of `std::cmp::PartialEq` might be missing for `(&types::FeeRate<types::PerMill>,)````

File: crml/cennzx-spot/src/types.rs (hunk 10 above)
Comment (cowboy-bebug) [reply crml/cennzx-spot/src/types.rs]: So we can derive it for `FeeRate`

File: crml/cennzx-spot/src/types.rs (hunk 14 above)
Comment (aliXsed) [reply crml/cennzx-spot/src/types.rs]: PerMill, PerMilli, ... are empty types. There can't be any instantiation from them. That's why I don't think deriving PartialEq would do the thing we think it should. However I will test it again.

Comment (cowboy-bebug) [reply crml/cennzx-spot/src/types.rs]: It builds and tests fine on my machine. For more info on derive attribute: https://doc.rust-lang.org/reference/attributes/derive.html

File: crml/cennzx-spot/src/lib.rs (hunk 8 above)
Comment (aliXsed) [reply crml/cennzx-spot/src/lib.rs:54]: As we discussed, this issues requires changes beyond this task and will be revisited later when we bring in our currency module and make cennzx depend on, instead of generic-assets.

**Comment by aliXsed** [comment]
@holygits @cowboy-bebug If no more comments, would you please approve?

**Review by cowboy-bebug — APPROVED** [review]
Nice work!! 🎉

File: crml/cennzx-spot/src/types.rs [hunk 16]
```diff
@@ -16,87 +16,136 @@
...
+	const SCALE: LowPrecisionUnsigned = 1_000_000;
 }
 
-use codec::{Compact, CompactAs, Decode, Encode};
+/// Per thousandth of unit price
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum PerMilli {}
+impl Scaled for PerMilli {
+	const SCALE: LowPrecisionUnsigned = 1_000;
+}
```
hoani, line 46: I know I'm not strictly a reviewer on this, but I would suggest making `PerMill` and `PerMilli` more distinct from eachother.
It isn't obvious that Mill is abbreviating a Million. Also the two words are effectively "PerM" plus a bunch of lines - I remember reading a good writeup in Code Complete in what naming anti-patterns to avoid.

Anyway, I would suggest renaming these to `PerMilli` and `PerMillion` to avoid accidental mixing of the two scales =)

hoani: I don't mean to push the point too much, but a quick look online for the meaning of the word Mill gives:

image

jordy25519 (approved):
LGTM
I'd prefer to merge this as is, can address concerns in a follow up PR

aliXsed:
> LGTM
> I'd prefer to merge this as is, can address concerns in a follow up PR

I just noticed that unfortunately, sourcetree hasn't been configured quite the same as my command line git and thus my commits are not signed. I need to fix that before merging this branch :(

crml/cennzx-spot/src/types.rs
aliXsed: Sounds a good suggestion to me. However, because we have had the notions of Permill, Percent and even Perbill not limited to this place in the code, I would like to know if @holygits would prefer this change or not.

jordy25519:
squash and force push, the commit history is not required

PR #103, opened 2019-12-18
tidy up circleci

cowboy-bebug:
Changes:
- removed `wasm-gc ` stuff because rustc natively supports it now
- variablise install-sccache step so we can easily bump versions later, with reference to [this](https://github.com/mozilla/application-services/blob/master/.circleci/config.yml) ci config
---
Notes:
- steps cannot run in parallel, only jobs can:
  + tried to run checkout, install-rust and install-sccache steps in parallel, which wasn't possible
  + a solution to reduce time installing rust, etc. is to (1) create a base image building repo, (2) publish to docker hub with our specific needs and (3) use that image in this repo as base

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,19 +13,21 @@ commands:
             source ~/.cargo/env
             rustup install $RUST_VERSION
             rustup default $RUST_VERSION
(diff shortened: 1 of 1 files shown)


jordy25519 (approved):
LGTM

Issue #107, opened 2019-12-19
Update CENNZnut module matching


jordy25519:
Crate naming convention has changed upstream. It is no longer `srml-`
This should be updated to be more flexible e.g. split on first `-`
https://github.com/cennznet/cennznet/blob/87f52db68909e1eca582f11dab3e74f803edc537/runtime/src/lib.rs#L523

roy-powerfinance:
For internal Tracking: https://centralitydev.atlassian.net/browse/BT-103

roy-powerfinance:
PR is now completed: https://github.com/cennznet/cennznet/pull/108

PR #108, opened 2020-01-08
Changed the way module names prefixes are removed

roy-powerfinance:
#107 Changed the way module names prefixes are removed

We now allow module names prefixes with arbitrary length, instead of
hardcoding it to a length of 4, as long as the module input is in the format:
\<prefix\>-<module_name>

Also fixed some formatting issue with "cargo fmt"

### DIFF
--- a/runtime/src/doughnut.rs
+++ b/runtime/src/doughnut.rs
@@ -109,7 +109,7 @@ mod test {
 		let doughnut = make_doughnut("cennznet", cennznut.encode());
 		assert_err!(
 			verify_dispatch(&doughnut, "trmlattestation", "attest"),
-			"error during Module name segmentation"
+			"error during module name segmentation"
 		);
 	}
 }
--- a/runtime/src/doughnut.rs
+++ b/runtime/src/doughnut.rs
@@ -52,6 +52,20 @@ mod test {
 	}
 
 	#[test]
+	fn it_works_with_arbitrary_prefix_short() {
+		let cennznut = make_cennznut("attestation", "attest");
+		let doughnut = make_doughnut("cennznet", cennznut.encode());
+		assert_ok!(verify_dispatch(&doughnut, "t-attestation", "attest"));
+	}
+
+	#[test]
+	fn it_works_with_arbitrary_prefix_long() {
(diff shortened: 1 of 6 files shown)


runtime/src/doughnut.rs
```diff
@@ -88,4 +102,14 @@ mod test {
...
 		);
 	}
+
+	#[test]
+	fn it_fails_when_module_name_is_invalid() {
+		let cennznut = make_cennznut("attestation", "attest");
+		let doughnut = make_doughnut("cennznet", cennznut.encode());
+		assert_err!(
+			verify_dispatch(&doughnut, "trmlattestation", "attest"),
+			"error during Module name segmentation"
```
jordy25519: 👍

runtime/src/lib.rs
```diff
@@ -519,8 +519,15 @@ impl additional_traits::DelegatedDispatchVerifier<CennznetDoughnut> for Runtime
 			.ok_or("CENNZnut does not grant permission for cennznet domain")?;
 		let cennznut: CENNZnut = Decode::decode(&mut domain).map_err(|_| "Bad CENNZnut encoding")?;
 
-		// Strips [c|p|s]rml- prefix
-		match cennznut.validate(&module[5..], method, &[]) {
+		// Extract Module name from <prefix>-<Module_name>
+		let dash_pos;
+		match module.find('-') {
```
jordy25519: Can make it more idiomatic like this:
```
let module_offset = module.find('_').ok_or("error during module name segmentation")? + 1;
```

jordy25519 (changes requested):
Nice one. Can be tidied up but otherwise LGTM

runtime/src/lib.rs
cowboy-bebug: Can be `let _ = module.find...` to avoid unused variable warning as well

roy-powerfinance: shortened the code as Jordy suggested, it get rids of the warning at the same time

cowboy-bebug (approved):
Nice work!

jordy25519 (approved):

PR #110, opened 2020-01-12
Bugfix/109 module name

### LINKED ISSUE #109: Fix out-of-bounds in module segmentation
Could get an "out of bounds" here if a module name like `-` or `something-` is used
https://github.com/cennznet/cennznet/blob/e489a29dab0935ad2f357f254aa90055b25bae78/runtime/src/lib.rs#L524

(Issue existed before #108)

roy-powerfinance:
#109 Module name segmentation with invalid input

Added logic to handle cases when the module name is not valid
Added tests for when the module name/prefix are empty

Closes #109

### DIFF
--- a/runtime/src/doughnut.rs
+++ b/runtime/src/doughnut.rs
@@ -112,4 +112,34 @@ mod test {
 			"error during module name segmentation"
 		);
 	}
+
+	#[test]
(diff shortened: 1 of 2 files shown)


jordy25519 (approved):
TY for this

cowboy-bebug (approved):
Nice!

PR #111, opened 2020-01-31
Point deps to plug-rc2, and update crate name changes

cowboy-bebug:
Crate aliasing is still left as is (except for crml/*) and correct crate names are used to resolve "crate not found" issues. The project doesn't yet build as a whole so ci won't pass, but once merged others can start working on the task.

All `crml/*` modules are migrated to use plug-rc2 so these don't need to be migrated:
```
├── cli                       =>  
├── crml                      =>  
│   ├── cennzx-spot           =>  build & pass tests, crate alias removed
│   ├── sylo                  =>  build & pass tests, crate alias removed
│   └── transaction-payment   =>  build & pass tests, crate alias removed, updated
├── executor                  =>  
├── primitives                =>  
├── rpc                       =>  
├── rpc-client                =>  
├── runtime                   =>  
└── testing                   =>  
```

Ci had to be bumped to use `stable-2019-12-19` as one of the dependencies (yamux) uses experimental `#[non_exhaustive]`, which is now included in stable rust.


crml/cennzx-spot/src/types.rs
```diff
@@ -52,15 +51,24 @@ impl Scaled for PerCent {
...
-	pub enum Error {
-		Overflow,
-		DivideByZero,
-		EmptyPool,
+#[derive(Debug)]
+pub enum Error {
+	Overflow,
+	DivideByZero,
+	EmptyPool,
+}
```
cowboy-bebug, line 60: Decided not to use `decl_error!` as it now becomes `Error<T>` and `T: Trait` bound needs to be added to every function that returns `Error::<T>::Overflow` etc. Instead, `Into` trait is implemented below to handle `&'static str` return type

cli/src/factory_impl.rs
```diff
@@ -29,14 +29,14 @@ use codec::{Decode, Encode};
 use finality_tracker;
 use inherents::InherentData;
 use keyring::sr25519::Keyring;
+use node_transaction_factory::modes::Mode;
```
jordy25519:
```suggestion
use node_transaction_factory::{modes::Mode, RuntimeAdapter};
```

crml/cennzx-spot/src/types.rs
jordy25519: It would be good to keep `decl_error!` so the module can return error codes.
The internal conversion error types in this file should have some other internal error type which can convert into one of the `decl_error` variants.
But could leave a comment and do in another PR

crml/transaction-payment/Cargo.toml
```diff
@@ -5,25 +5,27 @@ authors = ["Centrality Developers <support@centrality.ai>"]
...
 [features]
 default = ["std"]
 std = [
-	"cennznet-primitives/std",
 	"codec/std",
-	"rstd/std",
-	"sr-primitives/std",
-	"support/std",
-	"system/std",
+	"cennznet-primitives/std",
```
jordy25519, line 25: nitpick: this was alphabetically ordered

mzxyz:
Maybe we could remove `RUST_VERSION` file under `.circleci` as we have config it in `environment` directly

jordy25519 (approved):
LGTM 🎉  TY @cowboy-bebug

cowboy-bebug:
> Maybe we could remove `RUST_VERSION` file under `.circleci` as we have config it in `environment` directly

I think it's used here: https://github.com/cennznet/cennznet/blob/50242691c47b6c4b1f155db2801872465f90616e/.circleci/config.yml#L58
Still need to be updated

crml/cennzx-spot/src/types.rs
cowboy-bebug: Yep I'll add a TODO

mzxyz (approved):

crml/transaction-payment/Cargo.toml
cowboy-bebug: Thanks for spotting this!

mzxyz:
Just test, could also set `cennznet-rpc-client` and `primitives` as `build & pass tests, crate alias removed, updated`

PR #112, opened 2020-02-02
Migrate the primitives module to the latest plug

roy-powerfinance:
*renamed some modules

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -5,10 +5,10 @@ authors = ["Centrality Developers <support@centrality.ai>"]
 edition = "2018"
 
 [dependencies]
-codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
-primitives = { package = "sp-core", git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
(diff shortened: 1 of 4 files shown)


primitives/Cargo.toml
```diff
@@ -5,10 +5,10 @@ authors = ["Centrality Developers <support@centrality.ai>"]
 edition = "2018"
 
 [dependencies]
-codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }
-primitives = { package = "sp-core", git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
+parity-scale-codec = { version = "1.0.0", default-features = false, features = ["derive"] }
```
cowboy-bebug: My bad for not clarifying this but could we keep the old `codec = { package = "parity-scale-codec", version = "1.0.0", default-features = false, features = ["derive"] }` <- this is used throughout plug

cowboy-bebug (changes requested):
Looks good. One small change requested. And can we order them alphabetically so it's easier to see (third-party crates should be at the top, followed by others)

cowboy-bebug (approved):
LGTM thanks!

PR #114, opened 2020-02-03
update runmtime/lib with plugblockchain rc2

mzxyz:
update  cennznet_runtime with plugblockchain rc2

* Change `type AccountId`
* Change `SessionKeys`
* Introduce 3 new runtime modules: `Identity`, `Society`, `Recovery`
* Update unit tests

Will update all the cates dependencies in another PR


runtime/src/lib.rs
```diff
@@ -542,30 +649,33 @@ construct_runtime!(
...
-		Treasury: treasury::{Module, Call, Storage, Event<T>},
+		Treasury: treasury::{Module, Call, Storage, Config, Event<T>},
 		Contracts: contracts,
 		Sudo: sudo,
 		ImOnline: im_online::{Module, Call, Storage, Event<T>, ValidateUnsigned, Config<T>},
-		AuthorityDiscovery: authority_discovery::{Module, Call, Config<T>},
+		AuthorityDiscovery: authority_discovery::{Module, Call, Config},
 		Offences: offences::{Module, Call, Storage, Event},
 		RandomnessCollectiveFlip: randomness_collective_flip::{Module, Call, Storage},
+		Identity: pallet_identity::{Module, Call, Storage, Event<T>},
```
jordy25519, line 676: Would prefer not to include new modules if they are not required as dependencies (Identity, Society, Recovery). Can always add them in later if needed

cowboy-bebug (approved):
Very nice! LGTM 😄

jordy25519 (changes requested):

runtime/src/lib.rs
cowboy-bebug: Yep this will be addressed in another PR soon

PR #117, opened 2020-02-03
Merge rpc module from plug into cennznet

roy-powerfinance:


mzxyz (approved):
Better to do rebase on the latest `1.0.0-rc2` branch

aliXsed:
@holygits Couldn't we somehow just use plug/bin/node/rpc/src/lib.rs instead of copying it across?

PR #123, opened 2020-02-05
Use decl_error in cennzx-spot

hoani:
- Change all Cennzx spot module errors into decl_error enum
- Decouple Cennzx Spot FeeRate Error type from Cennzx spot module

### DIFF
--- a/crml/cennzx-spot/src/lib.rs
+++ b/crml/cennzx-spot/src/lib.rs
@@ -81,10 +81,48 @@ decl_error! {
 		AssetSaleValueNotAboveZero,
 		// Asset to core sale price should be greater than zero
 		AssetToCorePriceNotAboveZero,
+		// Insufficient core asset balance in exchange account
+		InsufficientCoreAssetInExchangeBalance,
 		// Asset to core sale price exceeds the specified max. limit
 		AssetToCorePriceAboveMaxLimit,
+		// Core to asset sale price should be greater than zero
+		CoreToAssetPriceNotAboveZero,
+		// Core to asset sale price exceeds the specified max. limit
+		CoreToAssetPriceAboveMaxLimit,
 		// Asset to asset sale price exceeds the specified max. limit
 		AssetToAssetPriceAboveMaxLimit,
+		// Tried to overdraw liquidity
+		LiquidityTooLow,
+		// Minimum trade asset is required
+		MinimumTradeAssetIsRequired,
+		// Minimum core asset is required
+		MinimumCoreAssetIsRequired,
+		// Assets withdrawn to be greater than zero
+		AssetToWithdrawNotAboveZero,
+		// Amount of exchange asset to burn should exist
+		LiquidityToWithdrawNotAboveZero,
+		// Liquidity should exist
+		NoLiquidityToRemove,
+		// trade asset amount must be greater than zero
+		TradeAssetToAddLiquidityNotAboveZero,
+		// core asset amount must be greater than zero
+		CoreAssetToAddLiquidityNotAboveZero,
+		// not enough core asset in balance
+		CoreAssetBalanceToAddLiquidityTooLow,
+		// not enough trade asset balance
+		TradeAssetBalanceToAddLiquidityTooLow,
+		// Minimum liquidity is required
+		LiquidityMintableLowerThanRequired,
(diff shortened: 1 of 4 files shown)


cowboy-bebug (approved):
LGTM nice work!

crml/cennzx-spot/src/lib.rs
```diff
@@ -55,8 +56,83 @@ pub trait Trait: frame_system::Trait + pallet_generic_asset::Trait {
 	type UnsignedIntToBalance: From<LowPrecisionUnsigned> + Into<<Self as pallet_generic_asset::Trait>::Balance>;
 }
 
+decl_error! {
+	/// Error for the generic-asset module.
```
jordy25519: Update module name

crml/cennzx-spot/src/types.rs
```diff
@@ -53,18 +53,14 @@ impl Scaled for PerCent {
 
 // TODO: refactor below to use frame_support::decl_error!
```
jordy25519: Can remove the TODO

crml/cennzx-spot/src/lib.rs
```diff
@@ -55,8 +56,83 @@ pub trait Trait: frame_system::Trait + pallet_generic_asset::Trait {
...
+	pub enum Error for Module<T: Trait> {
+		/// Exchange pool is empty.
+		EmptyExchangePool,
+		// Insufficient trade asset reserve in exchange
```
aliXsed, line 63: `///` for all enum entries

crml/cennzx-spot/src/lib.rs
```diff
@@ -55,8 +56,83 @@ pub trait Trait: frame_system::Trait + pallet_generic_asset::Trait {
...
+		// Minimum core asset is required
+		MinimumCoreAssetIsRequired,
+		// Assets withdrawn to be greater than zero
+		AssetToWithdrawNotAboveZero,
+		// Amount of exchange asset to burn should exist
+		LiquidityToWithdrawNotAboveZero,
+		// Liquidity should exist
+		NoLiquidityToRemove,
+		// trade asset amount must be greater than zero
+		TradeAssetToAddLiquidityNotAboveZero,
```
alexsednz: Seems like we can merge a few of such errors in one enum entry as "AssetNotAboveZero". I know we want to generate a different error message, we should also take the fact that the type of the error is the same into account.

crml/cennzx-spot/src/lib.rs
```diff
@@ -55,8 +56,83 @@ pub trait Trait: frame_system::Trait + pallet_generic_asset::Trait {
...
+		// core asset amount must be greater than zero
+		CoreAssetToAddLiquidityNotAboveZero,
+		// not enough core asset in balance
+		CoreAssetBalanceToAddLiquidityTooLow,
```
alexsednz: The same here for "LiquidityTooLow"

jordy25519 (commented):
Quick first pass review

crml/cennzx-spot/src/lib.rs
```diff
@@ -55,8 +56,83 @@ pub trait Trait: frame_system::Trait + pallet_generic_asset::Trait {
...
+	/// Error for the generic-asset module.
+	pub enum Error for Module<T: Trait> {
+		/// Exchange pool is empty.
+		EmptyExchangePool,
+		// Insufficient trade asset reserve in exchange
+		InsufficientTradeAssetReserve,
+		// Insufficient core asset reserve in exchange
+		InsufficientCoreAssetReserve,
+		// Insufficient asset balance in buyer account
+		InsufficientBuyerTradeAssetBalance,
```
aliXsed, line 68: I think we probably don't need that many error messages as the real type of errors is the same in many places. Similar to the Overflow error that we have used in many places regardless of where it has happened, we could have a few error types such as InsufficientAssetBalance, AssetNotAboveZero, AmountNotAboveZero,... 
However, as this is a refactor which is not pursuing any significant output change, I can see you may not want to go down that route.

hoani: Yeah, there are as many error types as there were unique messages in the past. Totally agree that we should consider reducing the number of errors going forward

jordy25519 (approved):
LGTM!

PR #124, opened 2020-02-05
Migrate/zh/cli

mzxyz:
Migrate all the updates for cli from latest plugBlockchain


cli/build.rs
```diff
@@ -49,5 +51,5 @@ fn build_completion(shell: &Shell) {
 
 	fs::create_dir(&path).ok();
 
-	CoreParams::<NoCustom, NoCustom>::clap().gen_completions("cennznet-node", *shell, &path);
+	CoreParams::<NoCustom, NoCustom>::clap().gen_completions("substrate-node", *shell, &path);
```
roy-powerfinance: do we want to start using substrate-node now?

jordy25519 (approved):

cli/Cargo.toml
```diff
@@ -7,34 +7,32 @@ edition = "2018"
...
+futures = { version = "0.3.1", features = ["compat"] }
 hex-literal = "0.2.1"
-jsonrpc-core = "13.2.0"
+jsonrpc-core = "14.0.3"
 log = "0.4.8"
 rand = "0.7.2"
-serde = { version = "1.0.101", features = [ "derive" ] }
-structopt = "0.3.7"
-tokio = "0.1.22"
+structopt = "=0.3.7"
```
aliXsed, line 17: Shouldn't it be `structopt = "0.3.7"`?

cli/src/service.rs
```diff
@@ -62,22 +59,18 @@ macro_rules! new_full_start {
...
+			let maintainer = sc_transaction_pool::FullBasicPoolMaintainer::new(pool.pool().clone(), client);
+			let maintainable_pool = sp_transaction_pool::MaintainableTransactionPool::new(pool, maintainer);
+			Ok(maintainable_pool)
 		})?
 		.with_import_queue(|_config, client, mut select_chain, _transaction_pool| {
 			let select_chain = select_chain
 				.take()
 				.ok_or_else(|| sc_service::Error::SelectChainRequired)?;
-			let (grandpa_block_import, grandpa_link) =
-				sc_finality_grandpa::block_import::<_, _, _, cennznet_runtime::RuntimeApi, _, _>(
```
aliXsed, line 76: Do we really want to remove cennznet_runtime::RuntimeApi? which is specific to cennznet?

cli/src/service.rs
```diff
@@ -49,7 +47,6 @@ construct_simple_protocol! {
 ///
 /// Use this macro if you don't actually need the full service, but just the builder in order to
 /// be able to perform chain operations.
-#[macro_export]
```
aliXsed, line 52: This seems to be necessary

cowboy-bebug (approved):

cowboy-bebug:
I'll merge this one!

cli/Cargo.toml
cowboy-bebug: We want to use the same version as in plug-rc2
https://github.com/plugblockchain/plug-blockchain/blob/1.0.0-rc2/bin/node/cli/Cargo.toml#L33

cli/src/service.rs
cowboy-bebug: I think it's able to infer the type now

cli/src/service.rs
cowboy-bebug: Even without `#[macro_export]`, `new_full_start!` is in scope of `cli/src/cli.rs` for some reason. I think that's why parity dropped the attribute

PR #125, opened 2020-02-05
Fix/build issues

cowboy-bebug:


mzxyz:
RUST_VERSION file also need to be changed

executor/src/lib.rs
```diff
@@ -94,12 +94,13 @@ mod tests {
 
 	/// Default transfer fee
 	fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64) -> Balance {
-		let length_fee = TransactionBaseFee::get() + TransactionByteFee::get() * (extrinsic.encode().len() as Balance);
+		let length_fee = TransactionByteFee::get() * (extrinsic.encode().len() as Balance);
```
aliXsed, line 97: How this could have happened in the merge?

cowboy-bebug: The change is from `plug-blockchain/1.0.0-rc2`. I think the formula is correct now

cli/src/cli.rs
```diff
@@ -198,7 +198,7 @@ where
 
 	let _ = exit_send.send(());
 
-	runtime.block_on(handle);
+	let _ = runtime.block_on(handle);
```
cowboy-bebug, line 201: For the `#[warn(unused_must_use)]` warning

runtime/src/impls.rs
```diff
@@ -47,24 +46,16 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
-/// By default, substrate node will have a weight range of [0, 1_000_000_000].
-pub struct WeightToFee;
-impl Convert<Weight, Balance> for WeightToFee {
-	fn convert(x: Weight) -> Balance {
+/// Convert from weight to balance via a simple coefficient multiplication
+/// The associated type C encapsulates a constant in units of balance per weight
+pub struct LinearWeightToFee<C>(sp_std::marker::PhantomData<C>);
+
+impl<C: Get<Balance>> Convert<Weight, Balance> for LinearWeightToFee<C> {
+	fn convert(w: Weight) -> Balance {
```
cowboy-bebug, line 54: @holygits actually all this refactoring doesn't introduce any changes. Shall we just keep the original `WeightToFee`?

executor/src/lib.rs
aliXsed: My question was rather about how we missed bringing this in, the first attempt? Is this an alert that the merge approach was not sound and so there could be more such issues or it was just a mistake?

executor/src/lib.rs
```diff
@@ -581,7 +580,7 @@ mod tests {
 				EventRecord {
 					phase: Phase::ApplyExtrinsic(0),
 					event: Event::system(frame_system::Event::ExtrinsicSuccess(DispatchInfo {
-						weight: 10000,
+						weight: 10_000,
```
jordy25519, line 583: So the difference is balance.transfer vs. ga.transfer weight?

jordy25519 (approved):
LGTM

runtime/src/impls.rs
jordy25519: I think keep whichever is closest to Plug rc2

executor/src/lib.rs
jordy25519: Oh nvm 😆

executor/src/lib.rs
cowboy-bebug: It was my mistake to not bringing in the change. I did this initial migration (https://github.com/cennznet/cennznet/pull/111) as quickly as I could so that other people can start working on it. And it makes sense that the change was picked up here

aliXsed (approved):

PR #128, opened 2020-02-12
Add post contract execution logic to handle gas payment

cowboy-bebug:
Implement `GasHandler` trait for Runtime as the post contract execution. It tries to access storage with `cennzx-spot` as key to get FeeExchange, which is consumed at `buy_fee_exchange`.

### DIFF
--- a/crml/cennzx-spot/src/lib.rs
+++ b/crml/cennzx-spot/src/lib.rs
@@ -23,7 +23,7 @@ mod mock;
 mod tests;
 
 mod impls;
-mod types;
+pub mod types;
 pub use impls::{ExchangeAddressFor, ExchangeAddressGenerator};
 pub use types::{FeeRate, HighPrecisionUnsigned, LowPrecisionUnsigned, PerMilli, PerMillion};
 
@@ -729,7 +729,7 @@ impl<T: Trait> Module<T> {
 	}
 
 	/// `asset_id` - Trade asset
-	/// `amount_sold` - Amount of trade assets sold
+	/// `amount_sold` - Amount of the trade asset to sell
 	/// Returns amount of core that can be bought with input assets.
 	pub fn get_asset_to_core_input_price(
 		asset_id: &T::AssetId,
@@ -898,11 +898,12 @@ impl<T: Trait> Module<T> {
 		buy_amount: T::Balance,
 		max_paying_amount: T::Balance,
 		fee_rate: FeeRate<PerMillion>,
-	) -> DispatchResult {
+	) -> sp_std::result::Result<T::Balance, DispatchError> {
 		let core_asset = Self::core_asset_id();
 		ensure!(asset_sold != asset_bought, Error::<T>::AssetCannotSwapForItself);
+		let sold_amount: T::Balance;
 		if *asset_sold == core_asset {
-			let _ = Self::make_core_to_asset_output(
+			sold_amount = Self::make_core_to_asset_output(
 				buyer,
 				recipient,
 				asset_bought,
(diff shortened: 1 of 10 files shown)


runtime/src/lib.rs
```diff
@@ -585,6 +589,32 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
 	}
 }
 
+/// Handles gas payment post contract execution (before deferring runtime calls) via CENNZX-Spot exchange.
+impl<T> pallet_contracts::GasHandler<T> for Runtime
+where
+	T: pallet_contracts::Trait + pallet_generic_asset::Trait + crml_cennzx_spot::Trait + Decode,
+{
+	fn empty_unused_gas(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
+		let key = &b"cennzx-spot"[..];
```
jordy25519: Could throw this into constants.rs and would prefer a name related to fee exchange / gas

runtime/src/lib.rs
```diff
@@ -585,6 +589,32 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
...
+
+		// The take() function ensures the entry is "killed" after access.
+		// Note: if there's no entry (ie, None), it means we handled an edge case of
+		// CPAY in FeeExchange, ie, gas is already paid in CPAY.
+		if let Some(exchange_op) = unhashed::take::<FeeExchange>(&key) {
+			let gas_spent = gas_meter.spent();
+
+			// Fee exchange can never fail as conditions such as having enough liquidity
+			// are checked early (before FeeExchange is put into database)
```
jordy25519: for consistency `database` -> `storage`

runtime/src/lib.rs
```diff
@@ -585,6 +589,32 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
...
 
+/// Handles gas payment post contract execution (before deferring runtime calls) via CENNZX-Spot exchange.
+impl<T> pallet_contracts::GasHandler<T> for Runtime
+where
+	T: pallet_contracts::Trait + pallet_generic_asset::Trait + crml_cennzx_spot::Trait + Decode,
+{
+	fn empty_unused_gas(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
+		let key = &b"cennzx-spot"[..];
+
+		// The take() function ensures the entry is "killed" after access.
```
jordy25519: Don't forget the back ticks ` around keywords

runtime/src/lib.rs
cowboy-bebug: We will use something like this: `const GAS_FEE_EXCHANGE = "cennzx-spot"`

jordy25519: can you change the actual key byte string to match the variable name also?

runtime/src/lib.rs
```diff
@@ -585,6 +589,30 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
 	}
 }
 
+/// Handles gas payment post contract execution (before deferring runtime calls) via CENNZX-Spot exchange.
+impl<T> pallet_contracts::GasHandler<T> for Runtime
+where
+	T: pallet_contracts::Trait + pallet_generic_asset::Trait + crml_cennzx_spot::Trait + Decode,
+{
+	fn empty_unused_gas(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
```
aliXsed: How have you made sure that this function would be called for a contract?

cowboy-bebug: Thanks for spotting this!

crml/cennzx-spot/src/lib.rs
```diff
@@ -729,7 +729,7 @@ impl<T: Trait> Module<T> {
 	}
 
 	/// `asset_id` - Trade asset
-	/// `amount_sold` - Amount of trade assets sold
+	/// `amount_sold` - Amount of the trade asset to sell
```
aliXsed, line 732: @hoani @royyang-centrality @holygits Do you think I should revert this change? I now think I didn't get what that sentence meant (still not getting it to be honest)

roy-powerfinance: I think your comment is correct. Here we use Trade asset to buy Core, and we specify the amount of trade asset we are willing to use, and expect the function to return the amount of Core we can buy.
I.e. I have 100 sylo, tell me how much Cpay I can get out of this

roy-powerfinance: the parameter name in quote should say `sell_amount`

jordy25519:
Superseded by #133

PR #129, opened 2020-02-12
Use the correct DelegatedDispatchVerifier type

jordy25519:
Use the CENNZnet validation impl


cowboy-bebug (approved):
Nice thanks for spotting this!

PR #131, opened 2020-02-14
Remove indices module

jordy25519:
Address type is equivalent to AccountId


cowboy-bebug (approved):
LGTM thanks for tidying up the code as well 👍

PR #132, opened 2020-02-18
Integration test setup

cowboy-bebug:
Changes:
- Add optional doughnut and fee_exchange param for signed_extra() in `testing/src/keyring.rs`
- Add 2 tests for simple GA transfer with & without FeeExchange
- Add empty tests so they can be implemented in the gas payment PR

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -111,7 +111,7 @@ parameter_types! {
 	pub const Version: RuntimeVersion = VERSION;
 }
 
-type CennznetDoughnut = prml_doughnut::PlugDoughnut<Runtime>;
+pub type CennznetDoughnut = prml_doughnut::PlugDoughnut<Runtime>;
 
 impl frame_system::Trait for Runtime {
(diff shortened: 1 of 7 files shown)


runtime/src/lib.rs
```diff
@@ -111,7 +111,7 @@ parameter_types! {
 	pub const Version: RuntimeVersion = VERSION;
 }
 
-type CennznetDoughnut = prml_doughnut::PlugDoughnut<Runtime>;
+pub type CennznetDoughnut = prml_doughnut::PlugDoughnut<Runtime>;
```
jordy25519, line 114: We should put this in primitives like all the other types

jordy25519 (approved):
Nice one!

runtime/src/lib.rs
jordy25519: can ignore this

cowboy-bebug: Yeah I wanted to preserve & use Runtime and stuff in it as much as possible

PR #133, opened 2020-02-20
91/tidy up

jordy25519:
The purpose of this PR is to provide a multi-currency gas payment implementation for CENNZnet

**Change Summary**

Adds:
- Custom gas handler implementation in `runtime/src/impls.rs`
- `FeeExchange` utility functions for accessing asset_id and max_payment fields
- 

Changes:
- `BuyFeeAsset` trait has added associated types for AccountId and Balance
- FeeExchange has re-introduced generic types for `AssetId` and `Balance`
Note: The purpose of these changes is incidental to help compiler type inference

- `ChargeTransactionPayment::validate` has been modified to store the fee exchange data for later retrieval by the custom GasHandler (this is a quick hack)
- Move transaction payment mocking into its own file `mock.rs`

- CENNZ-X `get_asset_to_core_input_price` has been modified to return the amount of assets sold.
This is so that the fee exchange max payment information can be updated accordingly

Key things to review are the custom GasHandler implementation and it's interaction with transaction payment `::validate()`


crml/cennzx-spot/src/impls.rs
```diff
@@ -92,7 +92,7 @@ pub(crate) mod impl_tests {
 	type CoreAssetCurrency = mock::CoreAssetCurrency<Test>;
 	type TradeAssetCurrencyA = mock::TradeAssetCurrencyA<Test>;
 	type FeeAssetCurrency = mock::FeeAssetCurrency<Test>;
-	type TestFeeExchange = FeeExchange<CennzXSpot>;
+	type TestFeeExchange = FeeExchange<u32, u128>;
```
aliXsed, line 94: The purpose of introducing a type which could get CennzXSpot as the generic type was that we only specify the type of asset id and balance in one place for CennzxSpot and FeeExchange uses the same types. It would prevent from specifying different types in different places incidentally. Aligned to the DRY principle.

crml/transaction-payment/src/lib.rs
```diff
@@ -243,48 +241,46 @@ where
...
-				));
+			if let Some(exchange) = &self.fee_exchange {
+				// Buy the CENNZnet fee currency paying with the user's nominated fee currency
+				let fees_purchased = T::BuyFeeAsset::buy_fee_asset(who, fee, &exchange)
+					.map_err(|_| TransactionValidityError::Invalid(InvalidTransaction::Payment))?;
+
+				// Temporarily store the `FeeExchange` info so that it can be used later by the-
+				// custom `GasHandler` impl see `runtime/src/impls.rs`
+				// This is a hack!
+				storage::unhashed::put(
```
aliXsed: This is now saying if the computed fee is zero, we shouldn't save the fee exchange for gas payment. Do we want to couple fee and gas this way?

crml/transaction-payment/src/lib.rs
```diff
@@ -243,48 +241,46 @@ where
...
+					&GAS_FEE_EXCHANGE_KEY,
+					&FeeExchange::new_v1(
+						exchange.asset_id(),
+						exchange.max_payment().checked_sub(&fees_purchased).unwrap_or(0.into()),
+					),
+				);
+			} else {
+				// Pre-caution to ensure `FeeExchange` data is cleared
+				storage::unhashed::kill(&GAS_FEE_EXCHANGE_KEY);
```
aliXsed: Because it was a cautionary measure to address possible past transaction's mess, the condition for doing that couldn't be subject to anything from the current transaction. Thus it was placed at the top with no conditions. We do that because the time for the previous transactions has come to an end and it's a new one.

crml/transaction-payment/src/lib.rs
```diff
@@ -445,7 +441,6 @@ mod tests {
 	type Balances = pallet_balances::Module<Runtime>;
 	type System = frame_system::Module<Runtime>;
 	type TransactionPayment = Module<Runtime>;
-	type TestFeeExchange = TransactionFeeExchange<Runtime>;
```
aliXsed: Compare this deleted line with the added one at 413. I would say this one is tidier and less error-prone. Any changes to Runtime would be followed automatically.

runtime/src/impls.rs
```diff
@@ -123,6 +128,122 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+	///    - User is paying in the native fee currency
+	///    - Deduct the 'fill meter cost' from the users balance and fill the gas meter
+	/// 2b) User has nominated to pay fees in another currency
+	///    - Calculate the 'fill gas cost' in terms of their nominated payment currency-
+	///      using the CENNZX spot exchange rate
+	///....- Check the user has liquid balance to pay the converted 'fill gas cost' and fill the gas meter
+	fn fill_gas(transactor: &T::AccountId, gas_limit: Gas) -> Result<GasMeter<T>, DispatchError> {
+		// Calculate the cost to fill the meter in the CENNZnet fee currency
+		let gas_price = Contracts::<T>::gas_price();
+		let fill_meter_cost = if gas_price.is_zero() {
```
aliXsed, line 159: Why we should evaluate zero separately? zero multiplied by anything would become zero.

runtime/src/impls.rs
```diff
@@ -123,6 +128,122 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+
+		// Calculate the `fill_meter_cost` in terms of the user's nominated payment asset
+		let converted_fill_meter_cost = CennzxSpot::<T>::get_asset_to_core_output_price(
+			&payment_asset,
+			T::Balance::unique_saturated_from(fill_meter_cost.saturated_into()),
+			CennzxSpot::<T>::fee_rate(),
+		)?;
+
+		// Respect the user's max. fee preference
+		if converted_fill_meter_cost > exchange_op.max_payment() {
```
aliXsed, line 195: This is against our conversation and decision here: 

>Question: if the gas_limit is greater than what fee_exchange.max_payment would allow I should take one of the following steps to resolve the case:
>  1. Correct the gas_limit to match max_payment and let the contract go ahead with this new limit which is below the original limit and thus not going to violate the user's permission but would give the contract a chance to succeed.
> 2. Reject the contract and return error
> Which of these two you think should be our approach? given that implementation wise, neither is harder than the other one?

>Answer: 1 is on the right track. We want to do the least suprising thing for the user. I think use max payment for transaction payment like normal then use gas_limit for contract ignoring max payment
> 
>

runtime/src/impls.rs
```diff
@@ -123,6 +128,122 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+		// Does the user have enough funds to pay the `converted_fill_meter_cost` with `payment_asset`
+		// also taking into consideration any liquidity restrictions
+		GenericAsset::<T>::ensure_can_withdraw(
+			&payment_asset,
+			transactor,
+			converted_fill_meter_cost,
+			WithdrawReason::Fee.into(),
+			// Calculate the expected user balance after paying the `converted_fill_meter_cost`
+			// This value is required to ensure liquidity restrictions are upheld
+			GenericAsset::<T>::free_balance(&payment_asset, transactor)
```
aliXsed: I don't see the value of mixing free_balance with ensure_can_withdraw from the readability's point of view. So one would ask what's the value of changing it from a work that was done?

runtime/src/impls.rs
```diff
@@ -123,6 +128,122 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+
+		// The `take()` function ensures the entry is killed after access
+		if let Some(exchange_op) = storage::unhashed::take::<FeeExchange<T::AssetId, T::Balance>>(&GAS_FEE_EXCHANGE_KEY)
+		{
+			// Pay for `gas_spent` in a user nominated currency using the CENNZX spot exchange
+			// Payment can never fail as liquidity is verified before filling the meter
+			if let Some(used_gas_cost) = gas_price.checked_mul(&gas_spent.saturated_into()) {
+				let _ = CennzxSpot::<T>::buy_fee_asset(
+					transactor,
+					T::Balance::unique_saturated_from(used_gas_cost.saturated_into()),
```
alexsednz: Good catch. Thank you for spotting it.

aliXsed:
> The purpose of this PR is to provide a multi-currency gas payment implementation for CENNZnet
> 
> **Change Summary**
> 
> Adds:
> 
> * Custom gas handler implementation in `runtime/src/impls.rs`
> * `FeeExchange` utility functions for accessing asset_id and max_payment fields
> 
> Changes:
> 
> * `BuyFeeAsset` trait has added associated types for AccountId and Balance
> * FeeExchange has re-introduced generic types for `AssetId` and `Balance`
>   Note: The purpose of these changes is incidental to help compiler type inference
> * `ChargeTransactionPayment::validate` has been modified to store the fee exchange data for later retrieval by the custom GasHandler (this is a quick hack)
> * CENNZ-X `get_asset_to_core_input_price` has been modified to return the amount of assets sold.
>   This is so that the fee exchange max payment information can be updated accordingly
> 
> Key things to review are the custom GasHandler implementation and it's interaction with transaction payment `::validate()`

Thank you for this very useful summary.

runtime/src/impls.rs
```diff
@@ -123,6 +128,122 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+		let gas_left = gas_meter.gas_left();
+		let gas_price = Contracts::<T>::gas_price();
+		let gas_spent = gas_meter.spent();
+
+		// The `take()` function ensures the entry is killed after access
+		if let Some(exchange_op) = storage::unhashed::take::<FeeExchange<T::AssetId, T::Balance>>(&GAS_FEE_EXCHANGE_KEY)
+		{
+			// Pay for `gas_spent` in a user nominated currency using the CENNZX spot exchange
+			// Payment can never fail as liquidity is verified before filling the meter
+			if let Some(used_gas_cost) = gas_price.checked_mul(&gas_spent.saturated_into()) {
```
aliXsed, line 237: Thank you for spotting that the gas spent hadn't been multiplied by the gas price here.

crml/cennzx-spot/src/impls.rs
jordy25519: Not essential for this PR

runtime/src/impls.rs
```diff
@@ -123,6 +128,139 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+
+impl IsGasMeteredCall for GasMeteredCallResolver {
+	/// The runtime extrinsic `Call` type
+	type Call = Call;
+	/// Return whether the given `call` is gas metered
+	fn is_gas_metered(call: &Self::Call) -> bool {
+		match call {
+			Call::Contracts(pallet_contracts::Call::call(_, _, _, _)) => true,
+			Call::Contracts(pallet_contracts::Call::instantiate(_, _, _, _)) => true,
+			_ => false,
```
aliXsed, line 265: put_code is also gas metered. Isn't it?

runtime/src/impls.rs
```diff
@@ -123,6 +128,139 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+		GenericAsset::<T>::ensure_can_withdraw(
+			&payment_asset,
+			transactor,
+			converted_fill_meter_cost,
+			WithdrawReason::Fee.into(),
+			// Calculate the expected user balance after paying the `converted_fill_meter_cost`
+			// This value is required to ensure liquidity restrictions are upheld
+			GenericAsset::<T>::free_balance(&payment_asset, transactor)
+				.checked_sub(&converted_fill_meter_cost)
+				.ok_or("Insufficient liquidity to fill gas meter")?,
```
cowboy-bebug: Do we want to check the liquidity of exchange as well? (please ignore me if we don't need to)
```rust
let ex_liquidity = CennzxSpot::<T>::get_liquidity(
    &(CENTRAPAY_ASSET_ID, payment_asset),
    transactor,
);

if converted_fill_meter_cost > ex_liquidity {
    return Err("Fee cost exceeds exchange liquidity pool".into());
}
```

aliXsed: Good point. @holygits what do you say?

jordy25519: This is accounted for when `converted_fill_meter_cost` is calculated by CENNZ-X

cowboy-bebug (approved):
LGTM !!

runtime/src/impls.rs
```diff
@@ -123,6 +131,142 @@ impl Convert<(Weight, Fixed64), Fixed64> for FeeMultiplierUpdateHandler {
...
+		return Ok(GasMeter::with_limit(gas_limit, gas_price));
+	}
+
+	/// Handle settlement of unused gas after contract execution
+	///
+	/// The process is as follows:
+	/// - Default case: refund unused gas tokens to the user (`transactor`) in CENNZnet's native fee currency as the current gas price
+	/// - FeeExchange case: Gas spent will be charged to the user in their nominated fee currency at the current gas price
+	fn empty_unused_gas(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
+		// TODO: Update `GasSpent` for the block
```
aliXsed, line 227: We can remove this line if we apply https://github.com/plugblockchain/plug-blockchain/pull/68 to plug rc2 as a hotfix.

jordy25519: Nice, will remove it

runtime/src/impls.rs
aliXsed: @holygits Can you please comment on this just to have a reference?

jordy25519: Final outcome is that `max_payment` is the maximum limit for total fees that will be paid including gas by a user for a transaction. So this implementation is corret

aliXsed: That must be the case anyhow. The question is under this condition and based on max_payment we can calculate a new gas_limit that can work. So instead of an error, we can execute the contract with a lower gas limit.

PR #140, opened 2020-02-25
Add gas payment integration tests

cowboy-bebug:
Added:
- more units for the currency in `constants.rs`
- `fn contract_fails`
- `fn contract_call_fails_with_insufficient_gas_without_fee_exchange`
- `fn contract_call_fails_with_insufficient_gas_with_fee_exchange`
- `fn contract_call_works_without_fee_exchange`
- `fn contract_call_works_with_fee_exchange`
- `fn contract_call_fails_when_fee_exchange_is_not_enough_for_gas`
- `fn contract_call_fails_when_exchange_liquidity_is_low`
- `fn contract_call_with_cpay_fee_exchange_does_not_use_exchange`

### DIFF
--- a/runtime/src/constants.rs
+++ b/runtime/src/constants.rs
@@ -36,7 +36,10 @@ pub mod asset {
 pub mod currency {
 	use cennznet_primitives::types::Balance;
 
-	pub const MILLICENTS: Balance = 1_000_000_000;
+	pub const PICOCENTS: Balance = 1;
+	pub const NANOCENTS: Balance = 1_000 * PICOCENTS;
+	pub const MICROCENTS: Balance = 1_000 * NANOCENTS;
+	pub const MILLICENTS: Balance = 1_000 * MICROCENTS;
 	pub const CENTS: Balance = 1_000 * MILLICENTS; // assume this is worth about a cent.
 	pub const DOLLARS: Balance = 100 * CENTS;
 }
--- a/runtime/src/impls.rs
+++ b/runtime/src/impls.rs
@@ -240,6 +240,14 @@ where
 					T::Balance::unique_saturated_from(used_gas_cost.saturated_into()),
 					&exchange_op,
 				);
+				let imbalance = T::Currency::withdraw(
+					transactor,
+					used_gas_cost,
+					WithdrawReason::Fee.into(),
+					ExistenceRequirement::KeepAlive,
+				)
+				.expect("Used gas cost could not be withdrawn");
+				T::GasPayment::on_unbalanced(imbalance);
 			}
 		} else {
 			// Refund remaining gas by minting it as CENNZnet fee currency
--- a/runtime/tests/mock.rs
+++ b/runtime/tests/mock.rs
@@ -19,6 +19,7 @@ use cennznet_runtime::{constants::asset::*, Runtime, VERSION};
 use cennznet_testing::keyring::*;
 use core::convert::TryFrom;
 use crml_cennzx_spot::{FeeRate, PerMilli, PerMillion};
+use pallet_contracts::{Gas, Schedule};
 
 pub const GENESIS_HASH: [u8; 32] = [69u8; 32];
 pub const SPEC_VERSION: u32 = VERSION.spec_version;
@@ -27,6 +28,9 @@ pub const SPEC_VERSION: u32 = VERSION.spec_version;
 pub struct ExtBuilder {
 	initial_balance: u128,
 	gas_price: u128,
+	// Configurable prices for certain gas metered operations
+	gas_sandbox_data_read_cost: Gas,
+	gas_regular_op_cost: Gas,
 }
 
 impl ExtBuilder {
@@ -36,6 +40,14 @@ impl ExtBuilder {
 	}
 	pub fn gas_price(mut self, gas_price: u128) -> Self {
 		self.gas_price = gas_price;
+		self
+	}
+	pub fn gas_sandbox_data_read_cost<T: Into<Gas>>(mut self, cost: T) -> Self {
+		self.gas_sandbox_data_read_cost = cost.into();
+		self
+	}
+	pub fn gas_regular_op_cost<T: Into<Gas>>(mut self, cost: T) -> Self {
+		self.gas_regular_op_cost = cost.into();
 		self
 	}
 	pub fn build(self) -> sp_io::TestExternalities {
@@ -48,8 +60,14 @@ impl ExtBuilder {
 		}
 		.assimilate_storage(&mut t)
 		.unwrap();
+
+		// Configure the gas schedule
+		let mut gas_price_schedule = Schedule::default();
+		gas_price_schedule.sandbox_data_read_cost = self.gas_sandbox_data_read_cost;
+		gas_price_schedule.regular_op_cost = self.gas_regular_op_cost;
+
 		pallet_contracts::GenesisConfig::<Runtime> {
-			current_schedule: Default::default(),
+			current_schedule: gas_price_schedule,
 			gas_price: self.gas_price,
 		}
 		.assimilate_storage(&mut t)
@@ -73,4 +91,54 @@ impl ExtBuilder {
 		.unwrap();
 		t.into()
 	}
+}
+
+/// Test contracts
+pub mod contracts {
+
+	/// Contract WABT for reading 32 bytes from memory
+	pub const CONTRACT_READ_32_BYTES: &str = r#"
+	(module
+		(import "env" "ext_scratch_read" (func $ext_scratch_read (param i32 i32 i32)))
+		(import "env" "memory" (memory 1 1))
+		(func (export "deploy"))
+		(func (export "call")
+			(call $ext_scratch_read
+				(i32.const 0)
+				(i32.const 0)
+				(i32.const 4)
+			)
+		)
+
+		;; 32 bytes for reading
+		(data (i32.const 4)
+			"\09\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00\00"
+		)
+	)"#;
+
+	/// Contract WABT for a contract which will fail during execution
+	pub const CONTRACT_WITH_TRAP: &str = r#"
+	(module
+		(import "env" "ext_scratch_read" (func $ext_scratch_read (param i32 i32 i32)))
+		(import "env" "memory" (memory 1 1))
+		(func (export "deploy"))
+		(func (export "call")
+			unreachable
+		)
+	)"#;
+
+	/// Contract WABT for a contract which dispatches a generic asset transfer of CENNZ to charlie
+	pub const CONTRACT_WITH_GA_TRANSFER: &str = r#"
+	(module
+		(import "env" "ext_dispatch_call" (func $ext_dispatch_call (param i32 i32)))
+		(import "env" "memory" (memory 1 1))
+		(func (export "call")
+			(call $ext_dispatch_call
+				(i32.const 8) ;; Pointer to the start of encoded call buffer
+				(i32.const 42) ;; Length of the buffer
+			)
+		)
+		(func (export "deploy"))
+		(data (i32.const 8) "\06\01\01\FA\90\B5\AB\20\5C\69\74\C9\EA\84\1B\E6\88\86\46\33\DC\9C\A8\A3\57\84\3E\EA\CF\23\14\64\99\65\FE\22\07\00\10\A5\D4\E8")
+	)"#;
 }
--- a/runtime/tests/tests.rs
+++ b/runtime/tests/tests.rs
@@ -14,17 +14,24 @@
 // You should have received a copy of the GNU General Public License
 // along with CENNZnet.  If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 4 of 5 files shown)


runtime/src/constants.rs
```diff
@@ -36,7 +36,10 @@ pub mod asset {
 pub mod currency {
 	use cennznet_primitives::types::Balance;
 
-	pub const MILLICENTS: Balance = 1_000_000_000;
+	pub const PICOCENTS: Balance = 1;
```
hoani, line 39: +1 +1 +1

cowboy-bebug: 👍 👍 👍 👍 👍 👍 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
 
 #[test]
 fn contract_fails() {
-	// Contract itself fails
+	ExtBuilder::default()
```
hoani, line 284: This is a great test, not at all related to this PR though.

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+fn contract_call_fails_with_insufficient_gas_without_fee_exchange() {
+	ExtBuilder::default()
+		.initial_balance(10_000 * TransactionBaseFee::get())
+		.gas_price(1)
+		.build()
+		.execute_with(|| {
+			initialize_block();
+			let xt = sign(CheckedExtrinsic {
+				signed: Some((alice(), signed_extra(0, 0, None, None))),
+				function: Call::Contracts(pallet_contracts::Call::call::<Runtime>(bob(), 10, 1000, vec![])),
```
hoani: This test fails because we don't provide a high enough gas limit
The purpose of this test is to check that our paying for gas fails because the user doesnt have enough fee assets in thier account - we should be setting `gas_limit` high and `initial_balance` low

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+			});
+
+			initialize_block();
+			let xt_at_base_call_fee = sign(CheckedExtrinsic {
+				signed: Some((alice(), signed_extra(0, 0, None, Some(fee_exchange.clone())))),
+				function: Call::Contracts(pallet_contracts::Call::call::<Runtime>(bob(), 10, 1000, vec![])),
+			});
+			assert_eq!(
+				Executive::apply_extrinsic(xt_at_base_call_fee),
+				Ok(Err(DispatchError::Other("not enough gas to pay transfer fee")))
```
hoani: Same issue as before, the user has enough CENNZ, but we set the `gas_limit` low - we need to give the user not enough CENNZ to pay for a high enough `gas_limit`, that way we are testing our `fill_gas` and `empty_gas` functions

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+				Ok(Err(DispatchError::Other("not enough gas to pay base call fee")))
+			);
+		});
 }
 
 #[test]
 fn contract_call_works_without_fee_exchange() {
-	// Happy case with no fee exchange
-	// Contract changes users account assets
+	let balance_amount = 10_000 * TransactionBaseFee::get();
```
hoani, line 520: 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+		});
 }
 
 #[test]
 fn contract_call_works_with_fee_exchange() {
-	// Happy case with fee exchange (with/without excess funds)
-	// Fee exchange is asking for CPay
-	// Contract makes an extrinsic to the exchange
-	// Contract changes users account assets
+	let balance_amount = 10_000 * TransactionBaseFee::get();
```
hoani, line 556: 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+				10, // min_liquidity
+				liquidity_asset_amount,
+				liquidity_core_amount,
+			);
+			let ex_key = (CENTRAPAY_ASSET_ID, CENNZ_ASSET_ID);
+			assert_eq!(CennzxSpot::get_liquidity(&ex_key, &charlie()), liquidity_core_amount);
+
+			let fee_exchange = FeeExchange::V1(FeeExchangeV1 {
+				asset_id: CENNZ_ASSET_ID,
+				max_payment: 1,
```
hoani, line 639: 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+		gas_limit_amount as u64,
+		vec![],
+	));
+
+	ExtBuilder::default()
+		.initial_balance(10_000 * TransactionBaseFee::get())
+		.gas_price(1)
+		.build()
+		.execute_with(|| {
+			let liquidity_core_amount = 100;
```
hoani, line 667: 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+	));
+
+	ExtBuilder::default()
+		.initial_balance(10_000 * TransactionBaseFee::get())
+		.gas_price(1)
+		.build()
+		.execute_with(|| {
+			let liquidity_core_amount = 100 * TransactionBaseFee::get();
+			let liquidity_asset_amount = 10 * TransactionBaseFee::get();
+			let _ = CennzxSpot::add_liquidity(
```
hoani: We don't need this setup since the call doesn't care about CENNZ

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+				max_payment: 100 * gas_limit_amount,
+			});
+
+			initialize_block();
+			let xt = sign(CheckedExtrinsic {
+				signed: Some((alice(), signed_extra(0, 0, None, Some(fee_exchange)))),
+				function: contract_call,
+			});
+			let r = Executive::apply_extrinsic(xt);
+			assert_eq!(r, Err(InvalidTransaction::Payment.into()));
```
hoani, line 720: 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +263,318 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+				function: contract_call,
+			});
+			initialize_block();
+			let r = Executive::apply_extrinsic(xt);
+			assert_eq!(r, Err(InvalidTransaction::Payment.into()));
+		});
+}
+
+#[test]
+fn contract_call_with_cpay_fee_exchange_does_not_use_exchange() {
```
hoani: Might need to rename this, originally we thought we could ignore the fee_exchange, but now we know the transaction will be rejected.
Maybe something like
`contract_call_with_cpay_fee_exchange_fails`

hoani (changes requested):
A couple of the tests need to be reworked to check when the user balance is too low to pay for contracts (currently they test if the gas_limit is too low).

~~I would like to see~~
~~`contract_call_works_with_fee_exchange_exact_amount`~~

Edit - after discussing with Eric, we realized that the `exact_amount` will always be moving target, so I don't think we should add it to our integration tests - it might be a good unit test for a later date

Let me know if I can assist with anything :)

runtime/tests/tests.rs
cowboy-bebug: Yep thanks for catching this 👍

runtime/tests/tests.rs
cowboy-bebug: How about? `contract_call_fails_when_cpay_is_used_for_fee_exchange`

runtime/tests/tests.rs
cowboy-bebug: Okay we trust plug for handling `gas_limit` too low case

runtime/tests/tests.rs
```diff
@@ -222,35 +280,441 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+}
+
+// Scenario:
+// - Extrinsic made with a contract call and fee payment in CENNZ
+// - Contract will dispatch a runtime call to move the callers CENNZ funds which should be used for payment
+// This must fail!
+#[test]
+fn contract_dispatches_runtime_call_funds_are_safu() {
+	ExtBuilder::default()
+		.initial_balance(100_000_000_000_000_000_000_000_000) // 1_000_000_000_000 * DOLLARS
```
jordy25519: This feels so wrong but don't have a better idea on how to test with sane numbers right now

runtime/tests/tests.rs
```diff
@@ -222,35 +280,441 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+				)),
+			});
+			assert_eq!(Executive::apply_extrinsic(xt), Err(InvalidTransaction::Payment.into()));
+		});
 }
 
 #[test]
-fn contract_fails_with_insufficient_gas() {
-	// Not enough gas to run contract
+fn contract_call_fails_with_insufficient_gas_with_fee_exchange() {
```
hoani, line 467: 👍

runtime/tests/tests.rs
```diff
@@ -222,35 +280,441 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+					topics: vec![],
+				},
+			];
+			assert_eq!(block_events[..7], events[..7]);
+		});
+}
+
+#[test]
+fn contract_call_fails_with_insufficient_gas_without_fee_exchange() {
+	ExtBuilder::default()
```
hoani, line 447: 👍

runtime/tests/tests.rs
cowboy-bebug: ~Let me try scale the numbers down a bit~ actually will fix in follow up PR

runtime/tests/tests.rs
```diff
@@ -222,35 +280,441 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+				CENNZ_ASSET_ID,
+				100_000_000_000_000_000_000_000,
+				100_000_000_000_000_000_000_000,
+				100_000_000_000_000_000_000_000,
+			)
+			.is_ok());
+
+			let bob_max_funds = 10_000_000_000_000;
+			// We use an encoded call in the contract
+			// if the test fails here the runtime encoding has changed so the contract WABT needs an update
```
hoani, line 331: Nice

runtime/tests/tests.rs
```diff
@@ -222,35 +280,441 @@ fn generic_asset_transfer_works_with_fee_exchange() {
...
+					phase: Phase::ApplyExtrinsic(2),
+					event: Event::system(frame_system::Event::ExtrinsicSuccess(DispatchInfo {
+						weight: 10000,
+						class: DispatchClass::Normal,
+						pays_fee: true,
+					})),
+					topics: vec![],
+				},
+			];
+			assert_eq!(block_events[..7], events[..7]);
```
jordy25519: 🔥

runtime/src/impls.rs
```diff
@@ -240,6 +240,14 @@ where
 					T::Balance::unique_saturated_from(used_gas_cost.saturated_into()),
 					&exchange_op,
 				);
+				let imbalance = T::Currency::withdraw(
+					transactor,
+					used_gas_cost,
+					WithdrawReason::Fee.into(),
+					ExistenceRequirement::KeepAlive,
+				)
+				.unwrap();
```
jordy25519: can you `expect` this so that it gives a meaningful error if it ever would fail

runtime/tests/tests.rs
cowboy-bebug: Oh I forgot to remove that lol thanks for spotting

runtime/tests/mock.rs
```diff
@@ -74,3 +92,53 @@ impl ExtBuilder {
...
+	(module
+		(import "env" "ext_scratch_read" (func $ext_scratch_read (param i32 i32 i32)))
+		(import "env" "memory" (memory 1 1))
+		(func (export "deploy"))
+		(func (export "call")
+			unreachable
+		)
+	)"#;
+
+	/// Contract WABT for a contract which dispatches a generic asset transfer of 1_000 CENNZ to charlie
```
jordy25519: Let's remove the value from this comment as it's out of date `1_000`

runtime/tests/tests.rs
hoani: It's not clear to me how these sets of events are showing the GA transfer has failed (assuming comments at the top of the test are still correct).
Suggest adding some comments to explain the assert logic

hoani (approved):
Suggest adding some comments to clarify GA contract test

jordy25519 (approved):
LGTM 🚀

hoani (approved):

runtime/tests/tests.rs
cowboy-bebug: This is somewhat addressed in the latest commit

PR #142, opened 2020-02-25
Better Transaction Payment Errors

mzxyz:
Better transaction payment errors:

- Catch the specific DispatchError from `cennzx-spot / buy_fee_asset` and map the error message to code number
- Check whether `fee_exchange_asset_id` exist at the top of `buy_fee_asset` fn
- Update test cases in `transaction_payment/lib.rs` and `executor/src/lib.rs`


crml/cennzx-spot/src/impls.rs
```diff
@@ -62,13 +62,20 @@ impl<T: Trait> BuyFeeAsset for Module<T> {
 		amount: Self::Balance,
 		exchange_op: &Self::FeeExchange,
 	) -> Result<Self::Balance, DispatchError> {
+		// check whether exchange asset id exist
+		let fee_exchange_asset_id = exchange_op.asset_id();
+		ensure!(
```
hoani, line 67: Nice addition

crml/transaction-payment/src/constants.rs
```diff
@@ -0,0 +1,53 @@
...
+	pub const ASSET_TO_CORE_PRICE_NOT_ABOVE_ZERO: u8 = 193;
+	pub const ASSET_TO_CORE_PRICE_ABOVE_MAX_LIMIT: u8 = 194;
+	pub const INSUFFICIENT_BUYER_TRADE_ASSET_BALANCE: u8 = 195;
+
+	pub const ASSET_SALE_VALUE_NOT_ABOVE_ZERO: u8 = 196;
+	pub const SALE_VALUE_BELOW_REQUIRED_MINIMUM: u8 = 197;
+	pub const INSUFFICIENT_SELLER_CORE_ASSET_BALANCE: u8 = 198;
+	pub const BUY_AMOUNT_NOT_POSITIVE: u8 = 199;
+	pub const INVALID_ASSET_ID: u8 = 200;
+	pub const UNKNOW_BUY_FEE_ASSET: u8 = 201;
```
hoani: `UNKNOW` -> `UNKNOWN` 😉

executor/src/lib.rs
```diff
@@ -900,7 +906,10 @@ mod tests {
 		.unwrap()
 		.into_encoded();
 		let r = ApplyExtrinsicResult::decode(&mut &r[..]).unwrap();
-		assert_eq!(r, Err(InvalidTransaction::Payment.into()));
+		assert_eq!(
+			r,
+			Err(TransactionValidityError::Invalid(InvalidTransaction::Custom(198)))
```
hoani: Is it possible to import the consts here so that:
`198` -> `INSUFFICIENT_SELLER_CORE_ASSET_BALANCE`?

hoani (approved):
Overall LGTM, a couple of small typos to consider fixing 👍

executor/src/lib.rs
mzxyz: Good point

crml/transaction-payment/src/constants.rs
```diff
@@ -0,0 +1,53 @@
...
+
+	pub fn buy_fee_asset_error_to_code(error: u8) -> u8 {
+		match error {
+			4 => INSUFFICIENT_BUYER_TRADE_ASSET_BALANCE,
```
jordy25519: Is it possible to import CENNZ-X error type here and match over that instead of raw number?

crml/transaction-payment/src/constants.rs
```diff
@@ -0,0 +1,53 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Transaction Payment Module
+//!
+//! Transaction Payment Customized Error Code Constants
+
+pub(crate) mod error_code {
```
jordy25519: Did you try using `decl_error!` for this?

crml/transaction-payment/src/constants.rs
mzxyz: Yes, using `message` will be better.

This is the struct for DispatchError:
```
pub enum DispatchError {
	/// Some error occurred.
	Other(#[codec(skip)] &'static str),
	/// Failed to lookup some data.
	CannotLookup,
	/// A bad origin.
	BadOrigin,
	/// A custom error in a module
	Module {
		/// Module index, matching the metadata module index
		index: u8,
		/// Module specific error value
		error: u8,
		/// Optional error message.
		#[codec(skip)]
		message: Option<&'static str>,
	},
}
```

So we can use `message` instead of `error` number for matching.
It will be like this:

```
pub fn buy_fee_asset_error_to_code(message: &'static str) -> u8 {
		match message {
			“InsufficientBuyerTradeAssetBalance” => INSUFFICIENT_BUYER_TRADE_ASSET_BALANCE,
                        _ => ....
}
```

crml/transaction-payment/src/constants.rs
mzxyz: The reason why I am not consider using `decl_error!` is `InvalidTransaction::Custom` only accept `u8` and also doughnut has hold some fields https://github.com/plugblockchain/plug-blockchain/blob/d2319b9c98c209be11544e5b05b4797e670c5be6/prml/doughnut/src/constants.rs#L19

so if we use `decl_error!` it will be hard to identify the `code` as for specific module the `error` code always starts from 1..n

jordy25519: Ah yes 👍

hoani (approved):
Nice changes

crml/transaction-payment/src/lib.rs
```diff
@@ -277,7 +288,9 @@ where
 				ExistenceRequirement::KeepAlive,
 			) {
 				Ok(imbalance) => imbalance,
-				Err(_) => return Err(TransactionValidityError::Invalid(InvalidTransaction::Payment)),
+				Err(_) => {
+					return Err(InvalidTransaction::Custom(error_code::INSUFFICIENT_SELLER_CORE_ASSET_BALANCE).into())
```
jordy25519: This error is abit confusing, it really means the user could not pay the transaction fee so would be good to have a simple, more meaningful error e.g.

"error_code::INSUFFICIENT_FEE_ASSET"

jordy25519 (commented):
One change requested then I think we're good to go!

hoani (approved):

runtime/tests/tests.rs
```diff
@@ -687,7 +696,7 @@ fn contract_call_fails_when_exchange_liquidity_is_low() {
 			});
 			initialize_block();
 			let r = Executive::apply_extrinsic(xt);
-			assert_eq!(r, Err(InvalidTransaction::Payment.into()));
+			assert_eq!(r, Err(InvalidTransaction::Custom(UNKNOWN_BUY_FEE_ASSET).into()));
```
jordy25519: The purpose of this PR is to make errors known so I didn't expect this 'unknown' kind to actually be used 🤔 
Should be an 'insufficient exchange liquidity for fee payment'

cowboy-bebug: Yep I'm working on this right now lol thanks for the comment

runtime/tests/tests.rs
```diff
@@ -717,6 +726,6 @@ fn contract_call_fails_when_cpay_is_used_for_fee_exchange() {
 				function: contract_call,
 			});
 			let r = Executive::apply_extrinsic(xt);
-			assert_eq!(r, Err(InvalidTransaction::Payment.into()));
+			assert_eq!(r, Err(InvalidTransaction::Custom(UNKNOWN_BUY_FEE_ASSET).into()));
```
jordy25519: This one is a rough edge:
 - 'CannotTradeFeeAssetForItSelf' seems wordy
- Could always check this case in 'buy fee asset' and return early knowing it would result in a CPAY payment anyway but more work 🤷‍♂

jordy25519 (approved):
Nice work all

crml/transaction-payment/src/constants.rs
```diff
@@ -0,0 +1,63 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! # Transaction Payment Module
+//!
+//! Transaction Payment Customized Error Code Constants
+
+pub mod error_code {
+	// Core to asset exchange errors
+	pub const CORE_TO_ASSET_PRICE_NOT_ABOVE_ZERO: u8 = 190;
```
aliXsed, line 23: I know it's approved and merged. However, I believe it's good to still share our suggestions. Here, I think enum is better suited for encompassing all the possible error cases.

crml/transaction-payment/src/constants.rs
```diff
@@ -0,0 +1,63 @@
...
+	pub const INSUFFICIENT_FEE_ASSET_BALANCE: u8 = 200;
+	pub const INVALID_ASSET_ID: u8 = 201;
+	pub const UNKNOWN_BUY_FEE_ASSET: u8 = 202;
+	pub const LIQUIDITY_TOO_LOW: u8 = 203;
+	pub const ASSET_CANNOT_SWAP_FOR_ITSELF: u8 = 204;
+	pub const INSUFFICIENT_CORE_ASSET_RESERVE: u8 = 205;
+
+	// Matches and converts crml-cennzx-spot module errors, such that
+	// they are propagated in crml-transaction-payment module
+	pub fn buy_fee_asset_error_msg_to_code(message: &'static str) -> u8 {
```
aliXsed, line 44: If you had the enum (Referring to my previous comment), then you may just need to implement ToString or Display for your enum. Then you should always pass the enum entry between functions until you want to log or print it in which case you would convert them to strings. So I believe converting strings to enum entries (or constants) which is not a light function (in terms of performance) could be avoidable.

crml/transaction-payment/src/mock.rs
```diff
@@ -142,10 +142,20 @@ impl BuyFeeAsset for Module<Runtime> {
 	) -> sp_std::result::Result<Self::Balance, DispatchError> {
 		if exchange_op.asset_id() == VALID_ASSET_TO_BUY_FEE {
 			if exchange_op.max_payment() == 0 {
-				return Err(DispatchError::Other("no money"));
+				return Err(DispatchError::Module {
+					index: 1,
+					error: 15,
+					message: Some("CoreToAssetPriceAboveMaxLimit"),
```
aliXsed, line 148: You shouldn't need to send a string like this. You should be able to return an enum entry/a defined const. Using hard-coded string literals is not a good practice.

crml/transaction-payment/src/constants.rs
cowboy-bebug: Yes I agree with you on this. But the reason why `u8` was used was because we have to propagate `InvalidTransaction::Custom(u8)` error type

crml/transaction-payment/src/constants.rs
cowboy-bebug: This function converts string message in `DispatchError` into the appropriate `InvalidTransaction::Custom(u8)` error type

crml/transaction-payment/src/mock.rs
cowboy-bebug: For this we wanted to return the error explicitly for testing purposes I guess lol

PR #144, opened 2020-02-27
Prototype/bt 154 treasury reward

roy-powerfinance:
Prototype code to award all transaction fee to all validators:
* Rewards are ```SplitTwoWays``` by 0 part to the treasury and 1 part to the validators
* Reward is then divided by the currently elected validators
* Each validator are rewarded their share with ```deposit_creating```

Note: This PR is for the team to review (optionally) and will be closed as we started to work on the actual implementation story

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -47,6 +47,7 @@ frame-executive = { git = "https://github.com/plugblockchain/plug-blockchain", d
 pallet-authority-discovery = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-authorship = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-babe = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
+pallet-balances = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-collective = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-contracts = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-contracts-rpc-runtime-api = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
--- a/runtime/src/impls.rs
+++ b/runtime/src/impls.rs
@@ -17,9 +17,9 @@
 //! Some configurable implementations as associated type for the substrate runtime.
 
 use crate::constants::fee::TARGET_BLOCK_FULLNESS;
-use crate::{BuyFeeAsset, Call, MaximumBlockWeight, Runtime};
+use crate::{Call, MaximumBlockWeight, Runtime};
 use cennznet_primitives::{
-	traits::IsGasMeteredCall,
+	traits::{BuyFeeAsset, IsGasMeteredCall},
 	types::{Balance, FeeExchange},
 };
 use crml_transaction_payment::GAS_FEE_EXCHANGE_KEY;
--- a/runtime/src/impls.rs
+++ b/runtime/src/impls.rs
@@ -268,6 +268,14 @@ where
 					T::Balance::unique_saturated_from(used_gas_cost.saturated_into()),
 					&exchange_op,
 				);
+				let imbalance = T::Currency::withdraw(
+					transactor,
+					used_gas_cost,
+					WithdrawReason::Fee.into(),
+					ExistenceRequirement::KeepAlive,
+				)
+				.expect("Used gas cost could not be withdrawn");
+				T::GasPayment::on_unbalanced(imbalance);
 			}
 		} else {
 			// Refund remaining gas by minting it as CENNZnet fee currency
--- a/runtime/src/impls.rs
+++ b/runtime/src/impls.rs
@@ -25,7 +25,8 @@ use cennznet_primitives::{
 use crml_transaction_payment::GAS_FEE_EXCHANGE_KEY;
 use frame_support::{
 	storage,
-	traits::{Currency, ExistenceRequirement, Get, OnUnbalanced, WithdrawReason},
+	traits::{Currency, ExistenceRequirement, Get, OnUnbalanced, Imbalance, WithdrawReason},
+	additional_traits::{MultiCurrencyAccounting, InherentAssetIdProvider},
 	weights::Weight,
 };
 use pallet_contracts::{Gas, GasMeter};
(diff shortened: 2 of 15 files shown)


runtime/src/lib.rs
```diff
@@ -203,11 +206,38 @@ parameter_types! {
...
+type NegativeImbalance = <<Runtime as crml_transaction_payment::Trait>::Currency as Currency<
+	<Runtime as frame_system::Trait>::AccountId,
+>>::NegativeImbalance;
+
+pub struct Validators;
+impl OnUnbalanced<NegativeImbalance> for Validators {
+	fn on_nonzero_unbalanced(amount: NegativeImbalance) {
+		let validators = pallet_staking::Module::<Runtime>::current_elected();
+		if validators.len() == 0 {
+			return;
```
hoani: [thoughts for later] Might still need to handle this case, so that we don't have a permanent imbalance between total issuance and actual issuance

runtime/src/lib.rs
```diff
@@ -203,11 +206,38 @@ parameter_types! {
...
+		}
+		let per_validator_reward = amount.peek() / (validators.len() as Balance);
```
hoani: [thoughts for later] Will need to think what we want to do with the remainder so that total issuance is correct.

eg, reward = 1000, # validators = 9
reward per validator = 111, remainder = 1

aliXsed: My understanding is that the inflation rate that we target at the end of an era would take care of the little discrepancies here.

runtime/src/lib.rs
aliXsed: Do you think this case would be likely to happen?

runtime/src/lib.rs
mzxyz: Agree with Alex, just keep it simple and the ignore the remainder will not have effect on the current mechanism

runtime/src/lib.rs
```diff
@@ -203,11 +206,38 @@ parameter_types! {
 	pub const WeightFeeCoefficient: Balance = 1_000;
 }
 
+type NegativeImbalance = <<Runtime as crml_transaction_payment::Trait>::Currency as Currency<
+	<Runtime as frame_system::Trait>::AccountId,
+>>::NegativeImbalance;
+
+pub struct Validators;
```
mzxyz: Better to move this part of code to `impl.rs`

runtime/Cargo.toml
```diff
@@ -47,6 +47,7 @@ frame-executive = { git = "https://github.com/plugblockchain/plug-blockchain", d
 pallet-authority-discovery = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-authorship = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-babe = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
+pallet-balances = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
```
jordy25519, line 50: Can remove this

runtime/src/impls.rs
```diff
@@ -35,6 +36,37 @@ use sp_runtime::{
...
+	<Runtime as frame_system::Trait>::AccountId>>::NegativeImbalance;
+type PositiveImbalance = <<Runtime as crml_transaction_payment::Trait>::Currency as Currency<
+	<Runtime as frame_system::Trait>::AccountId>>::PositiveImbalance;
+
+type CennzxSpot<T> = crml_cennzx_spot::Module<T>;
+type Contracts<T> = pallet_contracts::Module<T>;
+type GenericAsset<T> = pallet_generic_asset::Module<T>;
+type Staking<T> = pallet_staking::Module<T>;
+
+pub struct Validators;
```
jordy25519, line 49: Could use a more descriptive name + comment

runtime/src/impls.rs
```diff
@@ -35,6 +36,37 @@ use sp_runtime::{
...
+impl OnUnbalanced<NegativeImbalance> for Validators
+{
+	fn on_nonzero_unbalanced(amount: NegativeImbalance) {
+		let validators = Staking::<Runtime>::current_elected();
+		if validators.len() > 0 {
+			let per_validator_reward:Balance = amount.peek() / (validators.len() as Balance);
```
jordy25519, line 55: `per_validator_reward: Balance`

runtime/src/lib.rs
```diff
@@ -203,11 +206,22 @@ parameter_types! {
 	pub const WeightFeeCoefficient: Balance = 1_000;
 }
 
+type NegativeImbalance = <<Runtime as crml_transaction_payment::Trait>::Currency as Currency<
```
jordy25519, line 209: Re-use the version from `impls.rs` or vice versa

runtime/src/impls.rs
```diff
@@ -35,6 +36,37 @@ use sp_runtime::{
...
+
+			let mut total_imbalance = PositiveImbalance::zero();
+			for v in &validators {
+				let payout = <GenericAsset<Runtime> as MultiCurrencyAccounting>::deposit_creating(
+					&v,
+					Some(amount.asset_id().into()),
```
jordy25519, line 61: I think we should unroll this by getting the `asset_id` outside the loop once

runtime/src/impls.rs
```diff
@@ -35,6 +36,37 @@ use sp_runtime::{
...
+					per_validator_reward);
+				total_imbalance.subsume(payout);
+			}
+
+		}
+	}
```
hoani, line 67: Suggest keeping track of any unrewarded imbalance and burning it at the end of this function. 

This will also take care case where validators.len() == 0.

runtime/src/impls.rs
```diff
@@ -35,6 +36,37 @@ use sp_runtime::{
...
+		if validators.len() > 0 {
+			let per_validator_reward:Balance = amount.peek() / (validators.len() as Balance);
+
+			let mut total_imbalance = PositiveImbalance::zero();
+			for v in &validators {
+				let payout = <GenericAsset<Runtime> as MultiCurrencyAccounting>::deposit_creating(
+					&v,
+					Some(amount.asset_id().into()),
+					per_validator_reward);
+				total_imbalance.subsume(payout);
```
jordy25519, line 63: 👍

runtime/src/impls.rs
hoani: Spoke with Roy, this is handled

runtime/src/impls.rs
```diff
@@ -35,6 +36,37 @@ use sp_runtime::{
...
+type CennzxSpot<T> = crml_cennzx_spot::Module<T>;
+type Contracts<T> = pallet_contracts::Module<T>;
+type GenericAsset<T> = pallet_generic_asset::Module<T>;
+type Staking<T> = pallet_staking::Module<T>;
+
+pub struct Validators;
+impl OnUnbalanced<NegativeImbalance> for Validators
+{
+	fn on_nonzero_unbalanced(amount: NegativeImbalance) {
+		let validators = Staking::<Runtime>::current_elected();
```
jordy25519, line 53: if block fees are 0 we could also return

jordy25519 (commented):
Looks good overall! Some minor nitpicks

runtime/src/impls.rs
roy-powerfinance: i will run cargo fmt

runtime/Cargo.toml
roy-powerfinance: Removed in feature/bt_154_treasury_reward branch

runtime/src/impls.rs
roy-powerfinance: Renamed to SplitToAllValidators and added comments
 in feature/bt_154_treasury_reward branch

runtime/src/impls.rs
roy-powerfinance: Changed in the new feature branch

runtime/src/lib.rs
roy-powerfinance: libs now references negativeImbalance defined in impl
Added this to the new feature branch

runtime/src/impls.rs
roy-powerfinance: Changed this in feature/bt_154_treasury_reward branch

runtime/src/impls.rs
roy-powerfinance: Added in feature/bt_154_treasury_reward branch

roy-powerfinance:
All further work is done in feature/bt_154_treasury_reward branch.

mzxyz:
This PR can be closed

PR #146, opened 2020-02-27
Add remaining errors from cennzx-spot module

cowboy-bebug:
### Context
Transaction-payment module didn't include all the errors from the cennzx-spot module. So the remaining errors are added (so that they can be properly propagated throughout).

Changes:
- Copied errors preserving the order for easier maintenance
- Put extra errors at the top (so any new cennzx-spot errors can be easily added)
- more to come...


cowboy-bebug:
Closing this because we don't want to fully couple the errors this way

PR #150, opened 2020-03-03
Feature/bt 154 treasury reward

mzxyz:
#### Implementation:

* Rewards are `SplitTwoWays` by 0 part to the treasury and 1 part to the validators 
* Reward is then divided by the currently elected validators
* Each validator are rewarded their share with `deposit_creating`

#### Test cases:

*  Add gensis config for `staking` and  `session` modules in `mock.rs`
*  Test case: `staking_genesis_config_works`
*  Test case : `validators_received_equal_tx_fee_reward`

### DIFF
--- a/cli/src/chain_spec/mod.rs
+++ b/cli/src/chain_spec/mod.rs
@@ -46,19 +46,22 @@ pub mod rimu;
 
 type AccountPublic = <Signature as Verify>::Signer;
 
+/// A type contains authority keys
+pub type AuthorityKeys = (
+	AccountId,
+	AccountId,
+	GrandpaId,
+	BabeId,
+	ImOnlineId,
+	AuthorityDiscoveryId,
+);
+
 /// A type to hold keys used in CENNZnet node in SS58 format.
 pub struct NetworkKeys {
 	/// Endowed account address (SS58 format).
 	pub endowed_accounts: Vec<AccountId>,
 	/// List of authority keys
-	pub initial_authorities: Vec<(
-		AccountId,
-		AccountId,
-		GrandpaId,
-		BabeId,
-		ImOnlineId,
-		AuthorityDiscoveryId,
-	)>,
+	pub initial_authorities: Vec<AuthorityKeys>,
 	/// Sudo account address (SS58 format).
 	pub root_key: AccountId,
 }
@@ -112,13 +115,9 @@ pub fn get_authority_keys_from_seed(
 	)
 }
 
-/// Helper function to generate session keys
-fn session_keys(
-	grandpa: GrandpaId,
-	babe: BabeId,
-	im_online: ImOnlineId,
-	authority_discovery: AuthorityDiscoveryId,
-) -> SessionKeys {
+/// Helper function to generate session keys with authority keys
+pub fn session_keys(authority_keys: AuthorityKeys) -> SessionKeys {
+	let (_, _, grandpa, babe, im_online, authority_discovery) = authority_keys;
 	SessionKeys {
 		grandpa,
 		babe,
@@ -143,12 +142,7 @@ pub fn config_genesis(network_keys: NetworkKeys, enable_println: bool) -> Genesi
 		pallet_session: Some(SessionConfig {
 			keys: initial_authorities
 				.iter()
-				.map(|x| {
-					(
-						x.0.clone(),
-						session_keys(x.2.clone(), x.3.clone(), x.4.clone(), x.5.clone()),
-					)
-				})
+				.map(|x| (x.0.clone(), session_keys(x.clone())))
 				.collect::<Vec<_>>(),
 		}),
 		pallet_staking: Some(StakingConfig {
--- a/cli/src/chain_spec/mod.rs
+++ b/cli/src/chain_spec/mod.rs
@@ -48,17 +48,17 @@ type AccountPublic = <Signature as Verify>::Signer;
 
 /// A type contains authority keys
 pub type AuthorityKeys = (
-	/// stash account ID
+	// stash account ID
 	AccountId,
-	/// controller account ID
+	// controller account ID
 	AccountId,
-	/// Grandpa ID
+	// Grandpa ID
 	GrandpaId,
-	/// Babe ID
+	// Babe ID
 	BabeId,
-	/// ImOnline ID
+	// ImOnline ID
 	ImOnlineId,
-	/// Authority Discovery ID
+	// Authority Discovery ID
 	AuthorityDiscoveryId,
 );
 
--- a/cli/src/chain_spec/mod.rs
+++ b/cli/src/chain_spec/mod.rs
@@ -48,11 +48,17 @@ type AccountPublic = <Signature as Verify>::Signer;
 
 /// A type contains authority keys
 pub type AuthorityKeys = (
+	/// stash account ID
 	AccountId,
+	/// controller account ID
 	AccountId,
+	/// Grandpa ID
 	GrandpaId,
+	/// Babe ID
 	BabeId,
+	/// ImOnline ID
 	ImOnlineId,
+	/// Authority Discovery ID
 	AuthorityDiscoveryId,
 );
 
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -75,7 +75,6 @@ wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.
 cennznet-testing = { path = "../testing"}
 cennznet-cli = { path = "../cli" }
 sp-io = { git = "https://github.com/plugblockchain/plug-blockchain", branch = "1.0.0-rc2" }
-crml-sylo = { path = "../crml/sylo", default-features = false }
 hex = "0.4.0"
 hex-literal = "0.2.1"
 wabt = "0.9.2"
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -47,7 +47,6 @@ frame-executive = { git = "https://github.com/plugblockchain/plug-blockchain", d
 pallet-authority-discovery = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-authorship = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-babe = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
-pallet-balances = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-collective = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-contracts = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
 pallet-contracts-rpc-runtime-api = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "1.0.0-rc2" }
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -74,7 +74,9 @@ wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.
 
 [dev-dependencies]
 cennznet-testing = { path = "../testing"}
+cennznet-cli = { path = "../cli" }
 sp-io = { git = "https://github.com/plugblockchain/plug-blockchain", branch = "1.0.0-rc2" }
+crml-sylo = { path = "../crml/sylo", default-features = false }
 hex = "0.4.0"
 hex-literal = "0.2.1"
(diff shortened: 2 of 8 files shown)


runtime/tests/tests.rs
```diff
@@ -120,6 +121,74 @@ fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64, runtime_call:
...
+					total: staked_amount,
+					active: staked_amount,
+					unlocking: vec![],
+				}));
+			}
+		});
+}
+
+#[test]
+fn staking_validators_should_received_equal_transaction_fee_reward() {
```
royyang-centrality: @ @

runtime/Cargo.toml
```diff
@@ -73,7 +73,9 @@ wasm-builder-runner = { package = "substrate-wasm-builder-runner", version = "1.
 
 [dev-dependencies]
 cennznet-testing = { path = "../testing"}
+cennznet-cli = { path = "../cli" }
 sp-io = { git = "https://github.com/plugblockchain/plug-blockchain", branch = "1.0.0-rc2" }
+crml-sylo = { path = "../crml/sylo", default-features = false }
```
cowboy-bebug: I think this should be removed

runtime/src/impls.rs
```diff
@@ -35,6 +36,45 @@ use sp_runtime::{
...
+
+pub struct SplitToAllValidators;
+
+/// This handles the ```NegativeImbalance``` created for transaction fee.
+/// The reward is split evenly and distributed to all of the current elected validators.
+/// The remainder from the division are burned.
+impl OnUnbalanced<NegativeImbalance> for SplitToAllValidators {
+	fn on_nonzero_unbalanced(amount: NegativeImbalance) {
+		let validators = Staking::<Runtime>::current_elected();
+		if validators.len() > 0 || amount.peek().is_zero(){
```
jordy25519: should this be
```
if validators.len().is_zero() || amount.peek().is_zero() {
    return;
}
```

runtime/tests/mock.rs
```diff
@@ -15,30 +15,46 @@
...
 
+pub fn generate_initial_authorities() -> Vec<AuthorityKeys> {
+	vec![
+		get_authority_keys_from_seed("Alice"),
+		get_authority_keys_from_seed("Bob"),
+		get_authority_keys_from_seed("Charlie"),
+	]
+}
+
+// get all validators (slash account , controller account)
```
jordy25519: 'stash account'

runtime/tests/tests.rs
```diff
@@ -120,6 +121,85 @@ fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64, runtime_call:
...
+			let remainder = fee % validator_count;
+
+			let previous_total_issuance = GenericAsset::total_issuance(&CENTRAPAY_ASSET_ID);
+
+			initialize_block();
+			let r = Executive::apply_extrinsic(xt);
+			assert!(r.is_ok());
+
+			// Check total_issurance is adjusted
+			// FIXME: total_issurance is not adjusted
```
jordy25519: Since we mint exactly what was burned (except remainder) total issuance won't change until we introduce an inflation parameter in the reward step

runtime/tests/tests.rs
```diff
@@ -120,6 +121,74 @@ fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64, runtime_call:
...
+#[test]
+fn staking_genesis_config_works() {
+	let balance_amount = 10_000 * TransactionBaseFee::get();
+	ExtBuilder::default()
+		.initial_balance(balance_amount)
+		.build()
+		.execute_with(|| {
+			for validator in validators() {
+				let (stash, controller) = validator;
+				let staked_amount = balance_amount / 5;
```
hoani: Please comment that this value comes from Mock (ln 119)

runtime/src/impls.rs
cowboy-bebug: I thought so as well

runtime/src/impls.rs
```diff
@@ -35,6 +36,45 @@ use sp_runtime::{
...
+			// Get a list of elected validators
+			let per_validator_reward: Balance = amount.peek() / (validators.len() as Balance);
+
+			// This tracks the total amount of reward actually handed out. Used to adjust total issurance
+			let mut total_imbalance = PositiveImbalance::zero();
+			let asset_id = amount.asset_id();
+			for v in &validators {
+				let payout = <GenericAsset<Runtime> as MultiCurrencyAccounting>::deposit_creating(
+					&v,
```
jordy25519: This should pay to the validator reward destination account: https://github.com/plugblockchain/plug-blockchain/blob/d2319b9c98c209be11544e5b05b4797e670c5be6/frame/staking/src/lib.rs#L696

runtime/tests/mock.rs
```diff
@@ -82,13 +100,48 @@ impl ExtBuilder {
...
+				validators()[2].0.clone(),
+			],
 			next_asset_id: NEXT_ASSET_ID,
 			staking_asset_id: STAKING_ASSET_ID,
 			spending_asset_id: SPENDING_ASSET_ID,
 		}
 		.assimilate_storage(&mut t)
 		.unwrap();
+
+		let stash = self.initial_balance / 5;
```
cowboy-bebug: Instead of this pattern, we could add `stash` field in `struct ExtBuilder` and a setter method (up to u)

runtime/tests/tests.rs
```diff
@@ -120,6 +121,85 @@ fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64, runtime_call:
 	base_fee + fee_multiplier.saturated_multiply_accumulate(length_fee + weight_fee)
 }
 
+#[test]
+fn staking_genesis_config_works() {
+	let balance_amount = 10_000 * TransactionBaseFee::get();
+	ExtBuilder::default()
+		.initial_balance(balance_amount)
```
cowboy-bebug, line 130: We could add `.stash(balance_amount / 5)` here if we use the ExtBuilder setter pattern

jordy25519 (commented):
Looks good so far!
Could add some unit tests in impls.rs for the 0 reward case
Also would want to respect the validators preferred payment account

cli/src/chain_spec/mod.rs
```diff
@@ -46,19 +46,22 @@ pub mod rimu;
 
 type AccountPublic = <Signature as Verify>::Signer;
 
+/// A type contains authority keys
+pub type AuthorityKeys = (
```
aliXsed, line 50: Can you add a short line doc for every field of this tuple?

runtime/src/impls.rs
```diff
@@ -35,6 +36,45 @@ use sp_runtime::{
 	DispatchError, Fixed64,
 };
 
+pub type NegativeImbalance = <<Runtime as crml_transaction_payment::Trait>::Currency as Currency<
+	<Runtime as frame_system::Trait>::AccountId,
+>>::NegativeImbalance;
+pub type PositiveImbalance = <<Runtime as crml_transaction_payment::Trait>::Currency as Currency<
+	<Runtime as frame_system::Trait>::AccountId,
+>>::PositiveImbalance;
```
cowboy-bebug: This can be removed if we define them in `runtime/src/lib.rs` like this:
```rust
type PositiveImbalance = <GenericAsset as MultiCurrencyAccounting>::PositiveImbalance;
type NegativeImbalance = <GenericAsset as MultiCurrencyAccounting>::NegativeImbalance;
```
and import the above using `use crate::{PositiveImbalance, NegativeImbalance};`

roy-powerfinance: done

runtime/src/impls.rs
roy-powerfinance: yep fixed

runtime/tests/tests.rs
roy-powerfinance: Reward paid out now creates a positive imbalance that offsets the negative imbalance, so if there are remainders left, the overall issuance is adjusted.

runtime/src/impls.rs
roy-powerfinance: Implemented the required logic

runtime/src/impls.rs
```diff
@@ -47,26 +48,32 @@ pub struct SplitToAllValidators;
...
 
 		// This tracks the total amount of reward actually handed out. Used to adjust total issurance
 		let mut total_imbalance = PositiveImbalance::zero();
-		let asset_id = amount.asset_id();
-		for v in &validators {
+		let asset_id = imbalance.asset_id();
+		for validator in &validators {
+			let dest = Staking::<Runtime>::payee(validator);
+			let reward_destination_account_id = match dest {
+				RewardDestination::Controller => Staking::<Runtime>::bonded(validator).unwrap(),
```
cowboy-bebug: Maybe we want to use
```rust
Staking::<Runtime>::bonded(validator).unwrap_or_else(|| validator);
```
to ensure that if `bonded() -> None` then we just want to use validator

roy-powerfinance: Yep done.

runtime/src/impls.rs
```diff
@@ -17,24 +17,66 @@
...
+type Staking<T> = pallet_staking::Module<T>;
+
+pub struct SplitToAllValidators;
+
+/// This handles the ```NegativeImbalance``` created for transaction fee.
+/// The reward is split evenly and distributed to all of the current elected validators.
+/// The remainder from the division are burned.
+impl OnUnbalanced<NegativeImbalance> for SplitToAllValidators {
+	fn on_nonzero_unbalanced(imbalance: NegativeImbalance) {
+		let validators = Staking::<Runtime>::current_elected();
```
aliXsed, line 52: Can we be certain that on_nonzero_unbalanced would be called before a new set of validators are elected for a new era? In other words, are we rewarding the correct set of validators? [It's just a question]

runtime/src/impls.rs
```diff
@@ -17,24 +17,66 @@
...
+		if validators.len().is_zero() || imbalance.peek().is_zero() {
+			return;
+		}
+		// Get a list of elected validators
+		let per_validator_reward: Balance = imbalance.peek() / (validators.len() as Balance);
```
aliXsed, line 57: ```len() as Balance``` is a bad hack IMO. "Balance" should be divisible by a number. If it's not true we should address the root cause.

runtime/src/impls.rs
roy-powerfinance: this function is only called during validate(...) function, which I think is used to verify transactions. Therefore it will only be called when we are processing transactions - i.e. during an session when validators are all set and present.
Worse case scenario we have no validators and we just burn the transaction fee

roy-powerfinance: Also since the chain is running single threaded, I think it would not be possible for the validators to change in the middle of a transaction.
In addition, we make a copy of the list of validators, therefore even if the validators change, we would be OK

runtime/src/impls.rs
roy-powerfinance: Discussed with Alex, and we cannot find a better way of doing this. Casting it into Balance means our datatype remains flexible. We will keep as is

aliXsed (approved):

runtime/tests/mock.rs
```diff
@@ -15,30 +15,65 @@
...
 pub const GENESIS_HASH: [u8; 32] = [69u8; 32];
 pub const SPEC_VERSION: u32 = VERSION.spec_version;
 
-#[derive(Default)]
+pub fn generate_initial_authorities(n: usize) -> Vec<AuthorityKeys> {
+	let (accounts, mut authority_keys) = (vec!["Alice", "Bob", "Charlie", "Dave", "Eve", "Ferdie"], vec![]);
+	for i in 0..n as usize {
+		authority_keys.push(get_authority_keys_from_seed(accounts[i]))
+	}
+	authority_keys
```
jordy25519:
```suggestion
let accounts = vec!["Alice", "Bob", "Charlie", "Dave", "Eve", "Ferdie"];
accounts[..n].map(|s| get_authority_keys_from_seed(s)).collect()

```

runtime/tests/mock.rs
```diff
@@ -27,30 +27,45 @@ use sp_runtime::Perbill;
 pub const GENESIS_HASH: [u8; 32] = [69u8; 32];
 pub const SPEC_VERSION: u32 = VERSION.spec_version;
 
-pub fn generate_initial_authorities() -> Vec<AuthorityKeys> {
-	vec![
-		get_authority_keys_from_seed("Alice"),
-		get_authority_keys_from_seed("Bob"),
-		get_authority_keys_from_seed("Charlie"),
-	]
+pub fn generate_initial_authorities(n: usize) -> Vec<AuthorityKeys> {
```
roy-powerfinance: Since this is a public function, we cannot ensure that n <= 6. I think we should do some kind of boundary checks

cowboy-bebug: Thanks for spotting this!

jordy25519 (approved):
LGTM

PR #153, opened 2020-03-10
Amortised validator reward payouts

mzxyz:
Instead of triggering reward payout / balance transfer at every extrinsic, we want to store the rewards and make a payout once per era (per validator) for performance reasons.

## Implementation

Add:
- `CurrentEraTransactionRewards` storage in staking module
- add, set and get (behind test) methods for the storage above
- `fn split_rewards_evenly_to_all(recipients, total_amount)` to split total_amount (rewards) evenly across the recipients
- call the above method inside `fn new_era` responsible for payout at era change

Change:
- refactor `fn on_nonzero_unbalanced` from runtime/src/impls.rs so that transaction fee rewards are accumulated in the `CurrentEraTransactionRewards` storage

## Test
Config updates:
- Add `start_session` and `start_era` functions to trigger session and era manually.

Test cases:
- new session and new era work
-  transaction rewards storage update works
- validators received staking reward after new era start
- validators should received transaction reward after new era start

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -290,9 +290,6 @@ const MAX_NOMINATIONS: usize = 16;
 const MAX_UNLOCKING_CHUNKS: usize = 32;
 const STAKING_ID: LockIdentifier = *b"staking ";
 
-/// Hack reward currency asset id: Centrapay
-const REWARD_CURRENCY_ASSET_ID: u32 = 160001;
-
 /// Counter for the number of eras that have passed.
 pub type EraIndex = u32;
 
@@ -1362,22 +1359,29 @@ impl<T: Trait> Module<T> {
 		imbalance
 	}
 
-	/// Make transaction rewards to all the validators
+	/// Payout transaction rewards to all the validators. Called at the beginning of an era
 	fn transaction_reward_validators() {
 		let validators = Self::current_elected();
-		let mut total_imbalance = <RewardPositiveImbalanceOf<T>>::zero();
 		let total_reward = Self::current_era_transaction_rewards();
 		let validators_len: BalanceOf<T> = (validators.len() as u32).into();
 
 		if validators_len.is_zero() || total_reward.is_zero() {
-			return
+			return;
 		}
 
 		let reward = total_reward / validators_len;
 		for v in validators.iter() {
-			total_imbalance.maybe_subsume(Self::make_payout(&v, reward));
+			Self::make_payout(&v, reward);
 		}
 
+		// Hack: Issue/mint the amount of reward given out, so the total issuance stays consistent
+		let imbalance = T::RewardCurrency::issue(T::CurrencyToReward::from(total_reward).into());
+
+		// ```issue()``` returns a negative imbalance, which when dropped will revert the issue().
+		// To prevent this, ```mem::forget``` is used
+		sp_std::mem::forget(imbalance);
+
+		// reset chain storage for the total reward for the new era
 		<CurrentEraTransactionRewards<T>>::kill();
 	}
 
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1359,6 +1359,11 @@ impl<T: Trait> Module<T> {
 		imbalance
 	}
 
+	#[cfg(any(feature = "std", test))]
+	pub fn get_current_era_transaction_fee_reward() -> BalanceOf<T> {
+		CurrentEraTransactionRewards::<T>::get()
+	}
+
 	pub fn set_current_era_transaction_fee_reward(amount: BalanceOf<T>) {
 		CurrentEraTransactionRewards::<T>::mutate(|reward| {
 			*reward = amount;
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -728,7 +728,7 @@ decl_storage! {
 		CurrentEraPointsEarned get(fn current_era_reward): EraPoints;
 
 		/// Total transaction payment rewards for elected validators
-		CurrentEraTransactionRewards : BalanceOf<T>;
+		CurrentEraFeeRewards : BalanceOf<T>;
 
 		/// The amount of balance actively at stake for each validator slot, currently.
 		///
@@ -1361,23 +1361,23 @@ impl<T: Trait> Module<T> {
 
 	#[cfg(any(feature = "std", test))]
 	pub fn get_current_era_transaction_fee_reward() -> BalanceOf<T> {
-		CurrentEraTransactionRewards::<T>::get()
+		CurrentEraFeeRewards::<T>::get()
 	}
 
 	pub fn set_current_era_transaction_fee_reward(amount: BalanceOf<T>) {
-		CurrentEraTransactionRewards::<T>::mutate(|reward| {
+		CurrentEraFeeRewards::<T>::mutate(|reward| {
 			*reward = amount;
 		});
 	}
 
 	pub fn add_to_current_era_transaction_fee_reward(amount: BalanceOf<T>) {
-		CurrentEraTransactionRewards::<T>::mutate(|reward| {
+		CurrentEraFeeRewards::<T>::mutate(|reward| {
 			*reward = reward.checked_add(&amount).unwrap_or_else(|| *reward)
 		});
 	}
 
 	/// Payout transaction rewards to all the validators. Called at the beginning of an era
-	fn split_rewards_evenly_to_all(recipients: &Vec<T::AccountId>, total_amount: BalanceOf<T>) {
+	fn split_fee_rewards_evenly_to_all(recipients: &Vec<T::AccountId>, total_amount: BalanceOf<T>) {
 		let recipients_len: BalanceOf<T> = (recipients.len() as u32).into();
 
 		if recipients_len.is_zero() || total_amount.is_zero() {
@@ -1429,8 +1429,8 @@ impl<T: Trait> Module<T> {
 			let total_rewarded_stake = Self::slot_stake() * validator_len;
 
 			// Pay the accumulated tx fee as rewards to all validators
-			let total_tx_fee_reward = CurrentEraTransactionRewards::<T>::take();
-			Self::split_rewards_evenly_to_all(&validators, total_tx_fee_reward);
+			let total_tx_fee_reward = CurrentEraFeeRewards::<T>::take();
+			Self::split_fee_rewards_evenly_to_all(&validators, total_tx_fee_reward);
 
 			let (total_payout, max_payout) = inflation::compute_total_payout(
 				&T::RewardCurve::get(),
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -728,7 +728,7 @@ decl_storage! {
 		CurrentEraPointsEarned get(fn current_era_reward): EraPoints;
 
 		/// Total transaction payment rewards for elected validators
-		pub CurrentEraTransactionRewards get(fn current_era_transaction_rewards): BalanceOf<T>;
+		CurrentEraTransactionRewards : BalanceOf<T>;
 
 		/// The amount of balance actively at stake for each validator slot, currently.
 		///
@@ -1359,30 +1359,30 @@ impl<T: Trait> Module<T> {
 		imbalance
 	}
 
-	/// Payout transaction rewards to all the validators. Called at the beginning of an era
-	fn transaction_reward_validators() {
-		let validators = Self::current_elected();
-		let total_reward = Self::current_era_transaction_rewards();
-		let validators_len: BalanceOf<T> = (validators.len() as u32).into();
+	pub fn set_current_era_transaction_fee_reward( amount: BalanceOf<T> ){
+		CurrentEraTransactionRewards::<T>::mutate(|reward| {
+			*reward = amount;
+		});
+	}
 
-		if validators_len.is_zero() || total_reward.is_zero() {
+	pub fn add_to_current_era_transaction_fee_reward( amount: BalanceOf<T> ){
+		CurrentEraTransactionRewards::<T>::mutate(|reward| {
+			*reward = reward.checked_add(&amount).unwrap_or_else(|| *reward)
+		});
+	}
+
+	/// Payout transaction rewards to all the validators. Called at the beginning of an era
+	fn split_rewards_evenly_to_all(recipients: &Vec<T::AccountId>, total_amount: BalanceOf<T>) {
+		let recipients_len: BalanceOf<T> = (recipients.len() as u32).into();
+
+		if recipients_len.is_zero() || total_amount.is_zero() {
 			return;
 		}
 
-		let reward = total_reward / validators_len;
-		for v in validators.iter() {
-			Self::make_payout(&v, reward);
+		let reward = total_amount / recipients_len;
+		for r in recipients.iter() {
+			let _ = Self::make_payout(&r, reward);
 		}
-
-		// Hack: Issue/mint the amount of reward given out, so the total issuance stays consistent
-		let imbalance = T::RewardCurrency::issue(T::CurrencyToReward::from(total_reward).into());
-
-		// ```issue()``` returns a negative imbalance, which when dropped will revert the issue().
-		// To prevent this, ```mem::forget``` is used
-		sp_std::mem::forget(imbalance);
-
-		// reset chain storage for the total reward for the new era
-		<CurrentEraTransactionRewards<T>>::kill();
 	}
 
 	/// Session has just ended. Provide the validator set for the next session if it's an era-end.
@@ -1425,8 +1425,10 @@ impl<T: Trait> Module<T> {
 			let validator_len: BalanceOf<T> = (validators.len() as u32).into();
 			let total_rewarded_stake = Self::slot_stake() * validator_len;
 
-			// should pay transaction rewards before allocating staking reward
-			Self::transaction_reward_validators();
+			// Pay the accumulated tx fee as rewards to all validators
+			let validators = Self::current_elected();
+			let total_tx_fee_reward = CurrentEraTransactionRewards::<T>::take();
+			Self::split_rewards_evenly_to_all(&validators, total_tx_fee_reward);
 
 			let (total_payout, max_payout) = inflation::compute_total_payout(
 				&T::RewardCurve::get(),
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -245,6 +245,7 @@
 
 #![recursion_limit = "128"]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(array_into_iter)]
 
 #[cfg(test)]
 mod mock;
@@ -821,6 +822,8 @@ decl_event!(
 		/// All validators have been rewarded by the first balance; the second is the remainder
 		/// from the maximum amount of reward.
 		Reward(RewardBalance, RewardBalance),
+		/// Transaction fee rewards are split evenly across all validators.
+		RewardFees(RewardBalance, u32),
 		/// One validator (and its nominators) has been slashed by the given amount.
 		Slash(AccountId, Balance),
 		/// An old slashing report from a prior era was discarded because it could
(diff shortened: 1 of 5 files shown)


crml/staking/src/lib.rs
```diff
@@ -726,6 +726,9 @@ decl_storage! {
 		/// Rewards for the current era. Using indices of current elected set.
 		CurrentEraPointsEarned get(fn current_era_reward): EraPoints;
 
+		/// Total transaction payment rewards for elected validators
+		pub CurrentEraTransactionRewards get(fn current_era_transaction_rewards): BalanceOf<T>;
```
jordy25519: Can we create a public setter method for this instead? so that other modules cannot access it directly ❓

crml/staking/src/lib.rs
```diff
@@ -1370,6 +1373,32 @@ impl<T: Trait> Module<T> {
 		imbalance
 	}
 
+	/// Payout transaction rewards to all the validators. Called at the beginning of an era
+	fn transaction_reward_validators() {
```
jordy25519: Passing in the `validators` and `total_reward` as parameters would make this function nicer

crml/staking/src/lib.rs
```diff
@@ -1370,6 +1373,32 @@ impl<T: Trait> Module<T> {
...
+
+		// Hack: Issue/mint the amount of reward given out, so the total issuance stays consistent
+		let imbalance = T::RewardCurrency::issue(T::CurrencyToReward::from(total_reward).into());
+
+		// ```issue()``` returns a negative imbalance, which when dropped will revert the issue().
+		// To prevent this, ```mem::forget``` is used
+		sp_std::mem::forget(imbalance);
+
+		// reset chain storage for the total reward for the new era
+		<CurrentEraTransactionRewards<T>>::kill();
```
jordy25519: Is it possible to use a `<CurrentEraTransactionRewards<T>>::take()` instead?

jordy25519 (commented):
Looks really good.
Some general tidy-ups suggested

crml/staking/src/lib.rs
```diff
@@ -1370,6 +1373,32 @@ impl<T: Trait> Module<T> {
 		imbalance
 	}
 
+	/// Payout transaction rewards to all the validators. Called at the beginning of an era
+	fn transaction_reward_validators() {
+		let validators = Self::current_elected();
+		let total_reward = Self::current_era_transaction_rewards();
```
hoani: If we want to avoid making the getter for this public (which I think is what Jordy is saying above) we can use `< CurrentEraTransactionRewards <T>>::get()` instead of `Self::current_era_transaction_rewards()`

crml/staking/src/lib.rs
hoani: ^ This is probably a better way to go than my above comment on using `::get()`

runtime/tests/mock.rs
```diff
@@ -93,14 +322,14 @@ impl ExtBuilder {
...
 		let mut endowed_accounts = vec![alice(), bob(), charlie(), dave(), eve(), ferdie()];
-		let initial_authorities = generate_initial_authorities(self.validator_count);
-		let stash_accounts: Vec<_> = initial_authorities.iter().map(|x| x.0.clone()).collect();
+		let initial_validators = validators(self.validator_count);
+		let stash_accounts: Vec<_> = initial_validators.iter().map(|x| x.0.clone()).collect();
 		endowed_accounts.extend(stash_accounts);
 
 		let mut t = frame_system::GenesisConfig::default()
-			.build_storage::<Runtime>()
+			.build_storage::<Test>()
```
cowboy-bebug: Something to note is: by passing in Test instead of Runtime, we are no longer testing the actual Runtime struct <- which was the original purpose I think

crml/staking/src/lib.rs
```diff
@@ -1370,6 +1373,32 @@ impl<T: Trait> Module<T> {
...
+		let validators_len: BalanceOf<T> = (validators.len() as u32).into();
+
+		if validators_len.is_zero() || total_reward.is_zero() {
+			return;
+		}
+
+		let reward = total_reward / validators_len;
+		for v in validators.iter() {
+			Self::make_payout(&v, reward);
```
cowboy-bebug: Do we need to handle PositiveImbalance returned from make_payout?

runtime/tests/mock.rs
cowboy-bebug: Had a discussion offline. We will keep using `Test` until we can correctly trigger session end based on Babe

crml/staking/src/lib.rs
roy-powerfinance: Added a Getter and an adder, made the storage private

crml/staking/src/lib.rs
roy-powerfinance: yep did that. Repurposed the function to just paying out some reward evenly to a bunch of people

crml/staking/src/lib.rs
roy-powerfinance: Removed the getter function.

crml/staking/src/lib.rs
roy-powerfinance: Used Take instead of get and kill

crml/staking/src/lib.rs
```diff
@@ -1373,6 +1373,11 @@ impl<T: Trait> Module<T> {
 		imbalance
 	}
 
+	#[cfg(any(feature = "std", test))]
```
cowboy-bebug: Not sure if this is frowned upon to have a function guarded behind test. But there are some examples from plug here:: https://github.com/plugblockchain/plug-blockchain/blob/develop/frame/system/src/lib.rs#L853

crml/staking/src/lib.rs
```diff
@@ -1373,6 +1373,11 @@ impl<T: Trait> Module<T> {
...
+	pub fn get_current_era_transaction_fee_reward() -> BalanceOf<T> {
+		CurrentEraTransactionRewards::<T>::get()
+	}
```
cowboy-bebug: Not sure if this is frowned upon to have a function guarded behind test. But there are some examples from plug here:: https://github.com/plugblockchain/plug-blockchain/blob/develop/frame/system/src/lib.rs#L853

runtime/tests/tests.rs
```diff
@@ -136,6 +136,7 @@ fn start_session(session_index: SessionIndex) {
 	for i in Session::current_index()..up_to_session_index {
 		System::set_block_number((i + 1).into());
 		pallet_babe::CurrentSlot::put(Babe::current_slot() + EpochDuration::get());
+		Timestamp::set_timestamp((System::block_number() * 1000).into());
```
cowboy-bebug, line 139: 👍 what do u think about using `ExpectedBlockTime::get()` (which is configured as 3 seconds) instead of 1000?

runtime/tests/tests.rs
```diff
@@ -151,16 +152,34 @@ fn start_era(era_index: EraIndex) {
...
-pub fn current_total_payout_for_duration(duration: u64) -> Balance {
+fn current_total_payout(validator_count: Balance) -> Balance {
+	let now = Timestamp::now();
+	let previous_era_start = <crml_staking::CurrentEraStart<Runtime>>::get();
+	let era_duration = now - previous_era_start;
+
 	crml_staking::inflation::compute_total_payout(
 		<Runtime as crml_staking::Trait>::RewardCurve::get(),
-		<Staking>::slot_stake() * 2,
+		<crml_staking::SlotStake<Runtime>>::get() * validator_count,
```
cowboy-bebug: `SlotStake` has a public get method so this also works:
```rust
Staking::slot_stake() * validator_count,
```

runtime/tests/tests.rs
```diff
@@ -151,16 +152,34 @@ fn start_era(era_index: EraIndex) {
 	assert_eq!(Staking::current_era(), era_index);
 }
 
-pub fn current_total_payout_for_duration(duration: u64) -> Balance {
+fn current_total_payout(validator_count: Balance) -> Balance {
```
cowboy-bebug: 👍

runtime/tests/tests.rs
```diff
@@ -151,16 +152,34 @@ fn start_era(era_index: EraIndex) {
...
+fn reward_validators(validators: &[(AccountId, AccountId)]) -> Balance {
+	let validator_len = validators.len() as Balance;
+	let total_payout = current_total_payout(validator_len);
+	assert!(total_payout > 1);
+	Staking::reward_by_ids(
+		validators
+			.iter()
+			.map(|v| (v.0.clone(), 1))
+			.collect::<Vec<(AccountId, u32)>>(),
+	);
```
cowboy-bebug: Can we leave a comment to explain what this `reward_by_ids` is doing?
[Optional] can refactor into something like:
```rust
// Mapping of stash account and points to be assigned to them
let validators_points: Vec<(AccountId, u32)> = validators
    .iter()
    .map(|v| (v.0.clone(), 1))
    .collect();
Staking::reward_by_ids(validators_points); // CurrentEraPointsEarned storage is updated with new points
```

crml/staking/src/lib.rs
```diff
@@ -1445,7 +1443,6 @@ impl<T: Trait> Module<T> {
 			let total_rewarded_stake = Self::slot_stake() * validator_len;
 
 			// Pay the accumulated tx fee as rewards to all validators
-			let validators = Self::current_elected();
```
cowboy-bebug: Removed because this was a duplicated line with Line 1441

runtime/tests/tests.rs
mzxyz: This change should be good, but when do the replacement,  all the test for `rewards` failed. Not sure what's the problem

cowboy-bebug: Yeah they become off by just 1 and I'm not sure why that happens yet.

crml/staking/src/lib.rs
```diff
@@ -726,6 +726,9 @@ decl_storage! {
 		/// Rewards for the current era. Using indices of current elected set.
 		CurrentEraPointsEarned get(fn current_era_reward): EraPoints;
 
+		/// Total transaction payment rewards for elected validators
+		CurrentEraTransactionRewards : BalanceOf<T>;
```
hoani: [Suggestion] To clearly distinguish between fee rewards and staking rewards, I suggest naming:
`CurrentEraTransactionRewards -> CurrentEraFeeRewards`
and
`split_rewards_evenly_to_all -> split_fee_rewards_evenly_to_all`

crml/staking/src/lib.rs
hoani: I agree with  @cowboy-bebug, but cannot find a nice way of exposing this without a public getter.
This will have to do unless we find a better solution

runtime/src/impls.rs
```diff
@@ -49,33 +46,11 @@ pub struct SplitToAllValidators;
...
-
-			let payout = <GenericAsset<Runtime> as MultiCurrencyAccounting>::deposit_creating(
-				&reward_destination_account_id,
-				Some(asset_id.into()),
-				per_validator_reward,
-			);
-			let _ = total_imbalance.subsume(payout);
+		let amount = imbalance.peek();
+
+		if !amount.is_zero() {
```
hoani, line 50: Is this check neccessary?
I don't think we need to worry if the amount is zero

crml/staking/src/lib.rs
mzxyz: Good suggestion, will update soon

runtime/src/impls.rs
mzxyz: @royyang-centrality actually for me, I'm not sure about this, if we remove the check, there is no problem, it is just prevent some unnecessary state calculation

runtime/tests/tests.rs
```diff
@@ -146,11 +147,40 @@ fn advance_session() {
...
 	start_session((era_index * SessionsPerEra::get()).into());
 	assert_eq!(Staking::current_era(), era_index);
 }
 
+fn current_total_payout(validator_count: Balance) -> Balance {
+	let now = Timestamp::now();
+	let previous_era_start = <crml_staking::CurrentEraStart<Runtime>>::get();
+	let era_duration = now - previous_era_start;
+
+	crml_staking::inflation::compute_total_payout(
```
hoani: to make this more readable, can we do:
`let (payout, _) = crml_staking::inflation::compute_total_...`
and then return `payout` instead of `.0`?

runtime/tests/tests.rs
```diff
@@ -183,11 +213,64 @@ fn start_era_works() {
...
+				function: runtime_call_1.clone(),
+			});
+			let xt_2 = sign(CheckedExtrinsic {
+				signed: Some((bob(), signed_extra(0, 0, None, None))),
+				function: runtime_call_2.clone(),
+			});
+
+			initialize_block();
+			start_era(1);
+			advance_session(); // advance a session to trigger the beginning of era 2
```
hoani, line 249: Isn't this starting era 1, not era 2? (referring to comments)

runtime/tests/tests.rs
```diff
@@ -183,11 +213,64 @@ fn start_era_works() {
...
+			// Advancing sessions shouldn't change transaction rewards storage
+			advance_session();
+			assert_eq!(Staking::get_current_era_transaction_fee_reward(), total_transfer_fee);
+			advance_session();
+			assert_eq!(Staking::get_current_era_transaction_fee_reward(), total_transfer_fee);
+
+			// At the start of the next era (13th session), transaction rewards should be cleared (and paid out)
+			start_era(2);
+			advance_session();
+			assert_eq!(Staking::get_current_era_transaction_fee_reward(), 0);
```
hoani, line 275: Would be good to check validator balances here, if possible?
From https://github.com/cennznet/cennznet/blob/ada8ea4eb35cc861c8e0a1185b7d3f66837b6bb2/runtime/tests/mock.rs#L32
it looks like the default validators are:
"Alice", "Bob", "Charlie" (it takes 3 by default)

hoani: If we do check these accounts, we might need to make Dave, Eve and Ferdie do the transactions to prevent Alice, Bob and Charlie from having their accounts changed by fees.

runtime/tests/tests.rs
```diff
@@ -261,17 +387,21 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
...
+					<GenericAsset as MultiCurrencyAccounting>::free_balance(&stash, Some(CENTRAPAY_ASSET_ID)),
+					balance_amount
+				);
+			}
 
+			start_era(2);
 			for validator in validators {
 				// Check tx fee reward went to the stash account of validator
 				assert_eq!(
 					<GenericAsset as MultiCurrencyAccounting>::free_balance(&validator.0, Some(CENTRAPAY_ASSET_ID)),
```
hoani: Can we use the same pattern:
`let (stash, _) = validator;`
as above to keep the checks consistent?

runtime/tests/tests.rs
```diff
@@ -261,17 +387,21 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
...
-					balance_amount + fee_reward
+					balance_amount + per_fee_reward + per_staking_reward
```
hoani: Can we use the same pattern: `let (stash, _) = validator;` as above to keep the checks consistent?

runtime/tests/tests.rs
mzxyz: In my opinion, there has `balance check` before and after 2 era in `staking_genesis_config_works`, for this test is just check whether `transaction_rewards_storage` working properly.

hoani (changes requested):
Overall looks good.
I would like a couple of checks to stash account balances added to the `current_era_transaction_rewards_storage_update_works` test.
Also a few nitpicks and suggestions - up to you if you want to follow them

runtime/tests/tests.rs
hoani: Discussed with @HaoZhaiX .
This is addressed in `staking_validators_should_receive_equal_transaction_fee_reward `

hoani (approved):
Discussed changes with @HaoZhaiX, changes look good to me, the remainder of my comments are suggestions or nitpicks.

runtime/tests/tests.rs
mzxyz: The comment should be changed to era 1

hoani (approved):

runtime/tests/tests.rs
```diff
@@ -242,37 +366,31 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
 
 			let fm = TransactionPayment::next_fee_multiplier();
 			let fee = transfer_fee(&xt, fm, &runtime_call);
-			let fee_reward = fee / validators.len() as u128;
-			let remainder = fee % validators.len() as u128;
+			let per_fee_reward = fee / validators.len() as Balance;
 
-			let previous_total_issuance = GenericAsset::total_issuance(&CENTRAPAY_ASSET_ID);
```
roy-powerfinance, line 248: Optional:
I think we should check if the total issuance is adjusted correctly in this test
I.e. total issuance in the new era = previous_issuance + total_staking_reward  - transaction_fee_remainder - staking_reward_remainder

roy-powerfinance (changes requested):
Discussed with Eric - I suggested that we add a total issuance check to the test just to be safe.
Otherwise this PR LGTM

runtime/tests/tests.rs
cowboy-bebug: Thanks for pointing this out. This will be addressed

runtime/tests/tests.rs
cowboy-bebug: The comment is correct: `assert_eq!(Staking::current_era(), 2);` passes after `advance_session()`. I'll add this assert and revert the comment back to the original

runtime/tests/tests.rs
```diff
@@ -419,23 +325,27 @@ fn staking_reward_should_work() {
...
 			}
 
-			start_era(2);
-			// TODO: assert total issuance
+			// Advance a session to begin era 2
+			advance_session();
+			assert_eq!(GenericAsset::total_issuance(CENNZ_ASSET_ID), total_issuance);
+			assert_eq!(
+				GenericAsset::total_issuance(CENTRAPAY_ASSET_ID), // FIXME: 33_000_000 is coming from a weird timing issue between sessions
+				total_issuance + inflation + 33_000_000, // FIXME: Changing era_duration in fn current_total_payout outputs different inflation amount (but not sure what it should be)
```
cowboy-bebug, line 346: Tried to figure out why we end up issuing `33_000_000` much more cennz (which is more than the max_payout), but still not sure. It's definitely something to do with the `era_duration` in `fn current_total_payout` (changing this results in different inflation amount).

roy-powerfinance: If you refer to the fn new_era(), the total_reward (first parameter returned from compute_total_payout()) is what is actually used to reward the validators. The max is only used to ensure the total_payout doesn't go over the max. 
therefore I think the test here should use total_reward to calculate total issuance 
new_issuance = total_reward / validator_len * validator_len; //this also takes care of remainders

roy-powerfinance: what's leftover from given out is actually burned
```let rest = _max_payout.saturating_sub(total_payout);
Self::deposit_event(RawEvent::Reward(total_payout, rest));```

Comment (cowboy-bebug) [reply runtime/tests/tests.rs:346]: What's leftover is *not* burned. It's issued using T::RewardCurrency (cpay) back to the T::RewardRemainder (which is currently configured as treasury):
https://github.com/cennznet/cennznet/blob/19ca3f16c6ceb9f56df388d320bfc002a32c50cc/crml/staking/src/lib.rs#L1439
(also see this commit that shows the related system events: https://github.com/cennznet/cennznet/pull/153/commits/21791f30e81573fb46af086e7d6bd8e272ac6f37)

------

https://github.com/cennznet/cennznet/blob/19ca3f16c6ceb9f56df388d320bfc002a32c50cc/crml/staking/src/lib.rs#L1413
^ The variable names are confusing but `max_payout` not only ensures that you don't pay over that amount, but also it's the inflation target that we are trying to achieve. Hence `total_payout` is paid out to the validators and nominators and then the `rest` is calculated (max_payout - total_payout) and issued to the `T::RewardRemainder`:
https://github.com/cennznet/cennznet/blob/19ca3f16c6ceb9f56df388d320bfc002a32c50cc/crml/staking/src/lib.rs#L1435-L1439

Comment (cowboy-bebug) [reply runtime/tests/tests.rs:346]: And if it helps the numbers are:
```rust
let (total_reward, inflation) = reward_validators(&validators);
assert_eq!(total_reward, 711_000_000);
assert_eq!(inflation,  1_896_000_000);
```

hoani: I'm gonna look into this a bit deeper.
@cowboy-bebug makes some interesting points - it might be that the currency overall should have a fixed inflation, but we will reward less than that

roy-powerfinance: Uhh yes you are right! the unbalance is implemented for the treasury:
impl<T: Trait> OnUnbalanced<NegativeImbalanceOf<T>> for Module<T>

and it puts the money into the treasury account:
let _ = T::Currency::resolve_creating(&Self::account_id(), amount);

I'm surprised at the difference between the two numbers though - like you said, it may have something to do with the duration of the era.

cowboy-bebug: I got a nice tip from @hoani to try out and resolve this issue. I'll mark it as resolved for now 😄

roy-powerfinance: As discussed in team planning, fixing this is split into another task, and will be done in another PR.

roy-powerfinance (approved):

PR #154, opened 2020-03-12
Add session related tests for integration

cowboy-bebug:
We want to test whether reward payout is working correctly at the end of each era. This PR adds 3 function to support that.

Change:
- One refactor in `runtime/tests/mock.rs`

Add:
- `fn start_session(session_index)`: starts session(s) up to `session_index` arg
- `fn advance_session()`: advance 1 session
- `fn start_era(era_index)`: starts era(s) up to `era_index` arg
- 3 unit tests for the above functions

### DIFF
--- a/runtime/tests/mock.rs
+++ b/runtime/tests/mock.rs
@@ -30,7 +30,11 @@ pub const SPEC_VERSION: u32 = VERSION.spec_version;
 fn generate_initial_authorities(n: usize) -> Vec<AuthorityKeys> {
 	assert!(n > 0 && n < 7); // because there are 6 pre-defined accounts
 	let accounts = vec!["Alice", "Bob", "Charlie", "Dave", "Eve", "Ferdie"];
-	accounts[..n].iter().map(|s| get_authority_keys_from_seed(s)).collect()
+	accounts
+		.iter()
+		.take(n)
+		.map(|s| get_authority_keys_from_seed(s))
+		.collect()
 }
 
 // get all validators (stash account , controller account)
--- a/runtime/tests/tests.rs
+++ b/runtime/tests/tests.rs
@@ -17,15 +17,17 @@
 use cennznet_primitives::types::{AccountId, Balance, FeeExchange, FeeExchangeV1};
 use cennznet_runtime::{
 	constants::{asset::*, currency::*},
-	Call, CennzxSpot, CheckedExtrinsic, ContractTransactionBaseFee, Event, Executive, GenericAsset, Origin, Runtime,
-	Staking, TransactionBaseFee, TransactionByteFee, TransactionPayment, UncheckedExtrinsic,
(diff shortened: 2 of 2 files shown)


runtime/tests/mock.rs
```diff
@@ -30,7 +30,11 @@ pub const SPEC_VERSION: u32 = VERSION.spec_version;
 fn generate_initial_authorities(n: usize) -> Vec<AuthorityKeys> {
 	assert!(n > 0 && n < 7); // because there are 6 pre-defined accounts
 	let accounts = vec!["Alice", "Bob", "Charlie", "Dave", "Eve", "Ferdie"];
-	accounts[..n].iter().map(|s| get_authority_keys_from_seed(s)).collect()
+	accounts
+		.iter()
+		.take(n)
```
cowboy-bebug, line 35: Found a nicer looking way to loop `n` elements

runtime/tests/tests.rs
```diff
@@ -121,6 +124,65 @@ fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64, runtime_call:
...
+	start_session((era_index * SessionsPerEra::get()).into());
+	assert_eq!(Staking::current_era(), era_index);
+}
+
+#[test]
+fn start_session_works() {
+	ExtBuilder::default().build().execute_with(|| {
+		start_session(1);
+		start_session(3);
+		start_session(5);
```
jordy25519, line 159: Should this test assert something?

runtime/tests/tests.rs
```diff
@@ -121,6 +124,65 @@ fn transfer_fee<E: Encode>(extrinsic: &E, fee_multiplier: Fixed64, runtime_call:
 	base_fee + fee_multiplier.saturated_multiply_accumulate(length_fee + weight_fee)
 }
 
+fn start_session(session_index: SessionIndex) {
+	// If we run the function for the first time, block_number is 1, which won't
+	// trigger Babe::should_end_session() so we have to run one extra loop. But
+	// successive calls don't need to run one extra loop. See Babe::should_epoch_change()
+	let up_to_session_index = if Session::current_index() == 0 {
+		session_index + 1
```
jordy25519, line 132: this could be `1`

hoani (approved):
LGTM

jordy25519 (approved):
lgtm, small suggestions added

runtime/tests/tests.rs
mzxyz: Inside `fn start_session` there already has `assert_eq!(Session::current_index(), session_index)`, So we just thinking this assert is enough

cowboy-bebug: Yeah there is assert inside the function (might be an anti pattern)

runtime/tests/tests.rs
cowboy-bebug: The variable names are confusing but we should loop for one extra time if the function hasn't been run already 👍

PR #155, opened 2020-03-12
Add string conversion for serde

cowboy-bebug:
Removed weird balance type conversion (as JS doesn't support u128), following this workaround: https://github.com/plugblockchain/plug-blockchain/blob/d2319b9c98c209be11544e5b05b4797e670c5be6/frame/transaction-payment/rpc/runtime-api/src/lib.rs#L37-L42
Numbers are serialised as string.

Changes:
- mostly followed the approach from the above link
- add one more type `ResponseType` to CennzxSpotApi which wraps the result
- remove the TryInto conversion

RPC response is now (see string value of success):
```
{"jsonrpc":"2.0","result":{"success":"1004"},"id":1}
```

### DIFF
--- a/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
+++ b/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
@@ -19,22 +19,42 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use codec::{Codec, Decode, Encode};
-use sp_runtime::{traits::SimpleArithmetic, RuntimeDebug};
+#[cfg(feature = "std")]
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use sp_runtime::RuntimeDebug;
 
 /// A result of querying the exchange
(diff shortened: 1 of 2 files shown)


crml/cennzx-spot/rpc/runtime-api/src/lib.rs
```diff
@@ -50,3 +70,32 @@ sp_api::decl_runtime_apis! {
 		) -> CennzxSpotResult<Balance>;
 	}
 }
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn serde_works_with_string() {
```
hoani, line 79: Very nice!

hoani (approved):
LGTM Nice work!

jordy25519:
Want to get #158 in first and this will also break js API

cowboy-bebug:
Closing in favour of https://github.com/cennznet/cennznet/pull/191

PR #156, opened 2020-03-12
Add cargo-fmt job to exit error when not formatted

cowboy-bebug:
Adding `cargo fmt` missing from ci config.

Changes:
- add `cargo-fmt` job
- run the job in parallel
- add dependency on the job for the ci to pass

Note:
- didn't look into setting up a hook to auto-format on pushing commits, as it's not ideal for the code to change without us realising at the moment


jordy25519 (approved):

mzxyz (approved):
LGTM

PR #159, opened 2020-03-19
Update docker to the latest circleci config

cowboy-bebug:
Instead of using pegged nightly version, the following changes are made to align with the latest circleci settings:
- use stable 1.42 as default
- install nightly wasm toolchain

Update: builds locally and runs well with `docker run cennznet:latest --dev`


jordy25519 (approved):
🐳

PR #160, opened 2020-03-22
Refactor/runtime tests

cowboy-bebug:
Refactor `runtime/tests/tests.rs` to address some technical debts and prepare for upcoming tasks.

---
Changes:
- remove fee_multiplier param from  `fn transfer_fee`
    - because we don't dynamically set fee multiplier for the tests (see https://github.com/cennznet/cennznet/pull/160/commits/b1400abced24214925dbdf9745b83450092a8f0c)
- refactor `fn initialize_block` to return Header instead of wrapping Executive::initialize_block
    - because it doesn't make sense to only wrap Executive::initialize_block in a function while Executive::apply_extrinsic is not (see https://github.com/cennznet/cennznet/pull/160/commits/f2873fa0e9b7abbb19579260d6ab06ef4ba1ead4)
- bump down session_index used in `fn advance_session_works` test to reduce the number of loops
- remove `fn check_free_balance_updated` so that we know which lines of assert_eq! actually panic and check for balances in a more consistent way (see https://github.com/cennznet/cennznet/pull/160/commits/b5eaafdbf9aeef0857e1850c38b0455ef12f5ca6)
---
Something that was considered but not applied in this PR:
- some tests have boilerplate variable set up (eg, balance_amount), which could be set up as `const`
    - some of these can't be turned into a const due to `error[E0015]: calls in constants are limited to constant functions, tuple structs and tuple variants`
    - and it's not as bad as the above points

### DIFF
--- a/runtime/tests/tests.rs
+++ b/runtime/tests/tests.rs
@@ -18,7 +18,7 @@ use cennznet_primitives::types::{AccountId, Balance, FeeExchange, FeeExchangeV1}
 use cennznet_runtime::{
 	constants::{asset::*, currency::*},
 	Babe, Call, CennzxSpot, CheckedExtrinsic, ContractTransactionBaseFee, EpochDuration, Event, Executive,
-	GenericAsset, Origin, Runtime, Session, SessionsPerEra, Staking, System, Timestamp, TransactionBaseFee,
+	GenericAsset, Header, Origin, Runtime, Session, SessionsPerEra, Staking, System, Timestamp, TransactionBaseFee,
 	TransactionByteFee, TransactionPayment, UncheckedExtrinsic,
 };
 use cennznet_testing::keyring::*;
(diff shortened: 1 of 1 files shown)


runtime/tests/tests.rs
```diff
@@ -71,7 +70,7 @@ fn setup_contract(
 	let wasm = wabt::wat2wasm(contract_wabt).unwrap();
 	let code_hash = <Runtime as frame_system::Trait>::Hashing::hash(&wasm);
 
-	initialize_block();
+	Executive::initialize_block(&header());
```
mzxyz, line 73: Good change!

mzxyz (approved):

runtime/tests/tests.rs
cowboy-bebug: 👍

roy-powerfinance (approved):

PR #161, opened 2020-03-23
Fix staking reward functions to use T::RewardCurrency

jordy25519:
Some drive by fixes after looking at `fn new_era`

Notably:
- Change: reward payment functions to take `RewardBalanceOf<T>`
- Remove: `RewardDestination::Staked` as it's un-used in CENNZnet
    - Make default reward destination `Stash`
- Remove: `CurrencyToReward` type as it was un-used after previous changes

Small:
- Change: Removed `get_` prefix from getter to follow rust idiom
- Change: Move variable declarations closer to where they're used

Gets us closer to correct issuance
```rust
failures:

---- staking_reward_should_work stdout ----
thread 'staking_reward_should_work' panicked at 'assertion failed: `(left == right)`
  left: `120000001929000012`,
 right: `120000001896000000`', runtime/tests/tests.rs:344:13

---- staking_validators_should_receive_equal_transaction_fee_reward stdout ----
thread 'staking_validators_should_receive_equal_transaction_fee_reward' panicked at 'assertion failed: `(left == right)`
  left: `10000405120165066`,
 right: `10000405120166666`', runtime/tests/tests.rs:417:17
```

### DIFF
--- a/crml/staking/src/tests.rs
+++ b/crml/staking/src/tests.rs
@@ -921,8 +921,8 @@ fn reward_destination_works() {
 
 		start_era(1);
 
-		// Check that RewardDestination is Staked (default)
-		assert_eq!(Staking::payee(&11), RewardDestination::Staked);
+		// Check that RewardDestination is Stash (default)
+		assert_eq!(Staking::payee(&11), RewardDestination::Stash);
 		// Check that reward went to the stash account of validator
 		assert_eq!(Balances::free_balance(11), 1000 + total_payout_0);
 		// Check that amount at stake increased accordingly
--- a/crml/staking/src/tests.rs
+++ b/crml/staking/src/tests.rs
@@ -925,16 +925,6 @@ fn reward_destination_works() {
 		assert_eq!(Staking::payee(&11), RewardDestination::Stash);
 		// Check that reward went to the stash account of validator
 		assert_eq!(Balances::free_balance(11), 1000 + total_payout_0);
-		// Check that amount at stake increased accordingly
-		assert_eq!(
-			Staking::ledger(&10),
-			Some(StakingLedger {
-				stash: 11,
-				total: 1000 + total_payout_0,
-				active: 1000 + total_payout_0,
-				unlocking: vec![],
-			})
-		);
 
 		//Change RewardDestination to Stash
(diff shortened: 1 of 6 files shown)


cowboy-bebug (commented):
It makes sense to use T:RewardCurrency for payouts 👍  but it seems some test values have diverged slightly, eg, `﻿33_000_000 original vs 33_000_012 new`

runtime/tests/tests.rs
```diff
@@ -416,7 +416,7 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
 				// Check tx fee reward went to the stash account of validator
 				assert_eq!(
 					<GenericAsset as MultiCurrency>::free_balance(&stash, Some(CENTRAPAY_ASSET_ID)),
-					balance_amount + per_fee_reward + per_staking_reward
+					balance_amount + per_fee_reward + per_staking_reward - 1600
```
hoani, line 419: @holygits do we know why this value has changed?

crml/staking/src/tests.rs
```diff
@@ -1551,18 +1521,19 @@ fn slot_stake_is_least_staked_validator_and_exposure_defines_maximum_punishment(
...
-			assert_eq!(Staking::stakers(&21).total, 69 + total_payout_0 / 2);
+			// -- balances not change
+			assert_eq!(Staking::stakers(&11).total, 1000);
+			assert_eq!(Staking::stakers(&21).total, 69);
 
-			let _11_balance = Balances::free_balance(&11);
-			assert_eq!(_11_balance, 1000 + total_payout_0 / 2);
+			// -- stash account balance + reward
+			assert_eq!(Balances::free_balance(&11), 1000 + total_payout_0 / 2);
+			assert_eq!(Balances::free_balance(&21), 2000 + total_payout_0 / 2);
```
cowboy-bebug, line 1533: 👍👍👍

mzxyz (approved):
LGTM

cowboy-bebug (approved):
LGTM

runtime/tests/tests.rs
jordy25519: not 💯 sure on this. It will need to be resolved in the follow up PR

PR #162, opened 2020-03-24
Feature/bt 165 invulnerables

roy-powerfinance:
Added loggings to when:
* invulnerable validator is NOT slashed
* invulnerable validators are modified

Added a test that verified that  invulnerable validators can be changed via a function call
Added a test that verified that invulnerable validators can only be changed with root permission

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1197,8 +1197,8 @@ decl_module! {
 		fn set_invulnerables(origin, validators: Vec<T::AccountId>) {
 			ensure_root(origin)?;
 			<Invulnerables<T>>::put(validators.clone());
-			debug::print!("Set invulnerable:{:?}", validators.clone() );
-			Self::deposit_event(RawEvent::SetInvulnerables(validators.clone()) );
+			debug::print!("Set invulnerable:{:?}", validators );
+			Self::deposit_event(RawEvent::SetInvulnerables(validators) );
 
 		}
 
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -260,9 +260,9 @@ pub mod inflation;
 
 use codec::{Decode, Encode, HasCompact};
 use frame_support::{
-	decl_error, decl_event, decl_module, decl_storage, ensure,
+	debug, decl_error, decl_event, decl_module, decl_storage, ensure,
 	traits::{Currency, Get, Imbalance, LockIdentifier, LockableCurrency, OnUnbalanced, Time, WithdrawReasons},
-	weights::SimpleDispatchInfo, debug
+	weights::SimpleDispatchInfo,
 };
 use frame_system::{self as system, ensure_root, ensure_signed};
 use pallet_session::historical::SessionManager;
@@ -1803,7 +1803,9 @@ where
 				// Invulnerable validators do not get slashed
 				debug::print!(
 					"Invulnerable validator not slashed:{:?}, %:{:?}, session:{:?}",
-					stash, slash_fraction, slash_session
+					stash,
+					slash_fraction,
+					slash_session
 				);
 				continue;
 			}
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1197,7 +1197,9 @@ decl_module! {
 		#[weight = SimpleDispatchInfo::FixedNormal(5_000)]
 		fn set_invulnerables(origin, validators: Vec<T::AccountId>) {
 			ensure_root(origin)?;
-			<Invulnerables<T>>::put(validators);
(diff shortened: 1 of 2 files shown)


crml/staking/src/lib.rs
```diff
@@ -1191,7 +1191,9 @@ decl_module! {
 		#[weight = SimpleDispatchInfo::FixedNormal(5_000)]
 		fn set_invulnerables(origin, validators: Vec<T::AccountId>) {
 			ensure_root(origin)?;
-			<Invulnerables<T>>::put(validators);
+			<Invulnerables<T>>::put(validators.clone());
+			validators.iter().for_each( |v| debug::print!("Set invulnerable:{:?}", v) );
```
cowboy-bebug: [Option] Maybe something like this is also okay (just to avoid printing many lines): `debug::print!("Set invulnerables: {:?}", validators);`?

crml/staking/src/tests.rs
```diff
@@ -2313,6 +2313,68 @@ fn invulnerables_are_not_slashed() {
...
+	ExtBuilder::default().invulnerables(vec![11]).build().execute_with(|| {
+		assert_eq!(Staking::invulnerables(), vec![11]);
+
+		//try to change the invulnerable set without root access
+		let _ = Staking::set_invulnerables(Origin::signed(21), vec![21]);
+		assert_eq!(Staking::invulnerables(), vec![11]);
+
+		//Changing the invulnerables with root access.
+		let _ = Staking::set_invulnerables(Origin::ROOT, vec![21, 11]);
+		assert_eq!(Staking::invulnerables(), vec![21, 11]);
```
cowboy-bebug, line 2374: Can we add something like below to show that setting it empty is also possible:
```rust
let _ = Staking::set_invulnerables(Origin::ROOT, vec![]);
assert_eq!(Staking::invulnerables(), vec![]); // TODO: may want to control this logic
```

cowboy-bebug (commented):
LGTM! Could u also add an event for `set_invulerables`?

crml/staking/src/lib.rs
roy-powerfinance: yeah true

hoani: I'm assuming that each invulnerable is a 32-byte address? the print may look messy no matter how we print it. It would be awesome if we could do something like "Set invulnerables: 4ab8896...4c, 6ccb23...67", but that's definately out of scope =D

roy-powerfinance: changed this to a single call

crml/staking/src/tests.rs
roy-powerfinance: Yep added a new test for this

crml/staking/src/lib.rs
```diff
@@ -1191,7 +1196,10 @@ decl_module! {
 		#[weight = SimpleDispatchInfo::FixedNormal(5_000)]
 		fn set_invulnerables(origin, validators: Vec<T::AccountId>) {
 			ensure_root(origin)?;
-			<Invulnerables<T>>::put(validators);
+			<Invulnerables<T>>::put(validators.clone());
+			debug::print!("Set invulnerable:{:?}", validators.clone() );
+			Self::deposit_event(RawEvent::SetInvulnerables(validators.clone()) );
```
cowboy-bebug: [Optional] last clone can be removed

crml/staking/src/tests.rs
```diff
@@ -2313,6 +2313,82 @@ fn invulnerables_are_not_slashed() {
 	});
 }
 
+#[test]
+fn invulnerables_can_be_set() {
```
cowboy-bebug, line 2317: [Optional] Do we really need to add this test since it's very similar to `fn invulnerables_are_not_slashed` and `fn invulnerables_can_only_be_set_by_root` already covers it

cowboy-bebug (approved):
LGTM

crml/staking/src/lib.rs
jordy25519: `debug::print!("Set invulnerable:{:?}", validators.clone() );` does this need clone?

jordy25519 (commented):
lgtm!

jordy25519 (approved):

crml/staking/src/tests.rs
roy-powerfinance: the other 2 new tests only check if storage is correct, but do not check for invulnerability.
this is sort of a "end to end" test that tests the whole thing. 
I don't mind removing it though

crml/staking/src/lib.rs
roy-powerfinance: Yep removed both clone()

PR #163, opened 2020-03-24
Chore/staking test

mzxyz:
This PR try to fix test cases for staking reward and transaction payment reward

## Changes:
- Use `let mut total_imbalance = T::RewardCurrency::burn(Zero::zero());` to initialise `total_imbalance` which has `asset_id`.  (It is not an elegant way, we could optimise this after we fix the problem in GA)
- Add `CurrentEraDuration` in `Staking` module storage.
- Extract `current_total_payout` function from `new_era`, make it public so that in the test cases could also reuse it and make sure the `total_payout` and `max_payout` are same.

## Tests
-  Add `staking_inflation_reward_works`
- Check `total_issuance` in `multi_token_economy_tests`

---

tldr summry:
1. Extracted logic out from `fn new_era` into `pub fn current_total_payout` so it can be used in the tests
    - NOTE: a potential concern here is exposing `current_total_payout` but (1) it's only a getter and (2) doesn't expose any sensitive information
2. Added CurrentEraDuration storage with a getter to serve as one source of truth for computing payout
3. Refactored tests to meet those changes

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1476,7 +1476,7 @@ impl<T: Trait> Module<T> {
 	}
 
 	/// Compute current `total_payout` and `max_payout` for specific era duration
-	pub fn current_total_payout(total_tokens: RewardBalanceOf<T>) -> (RewardBalanceOf<T>, RewardBalanceOf<T>) {
+	pub fn current_total_payout(total_issuance: RewardBalanceOf<T>) -> (RewardBalanceOf<T>, RewardBalanceOf<T>) {
 		let validators = Self::current_elected();
 		let validator_len = validators.len() as u32;
 		let era_duration = Self::current_era_duration();
@@ -1486,7 +1486,7 @@ impl<T: Trait> Module<T> {
 		let (total_payout, max_payout) = inflation::compute_total_payout::<RewardBalanceOf<T>>(
 			&T::RewardCurve::get(),
 			total_rewarded_stake,
-			total_tokens,
+			total_issuance,
 			// Duration of era; more than u64::MAX is rewarded as u64::MAX.
 			era_duration.saturated_into::<u64>(),
 		);
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1419,7 +1419,7 @@ impl<T: Trait> Module<T> {
 			Self::split_fee_rewards_evenly_to_all(&validators, total_tx_fee_reward);
 
 			let (total_payout, max_payout) = Self::current_total_payout(T::RewardCurrency::total_issuance());
-			let mut total_imbalance = T::RewardCurrency::burn(Zero::zero()); // hack way to get new ImBalance with asset_id
+			let mut total_imbalance = T::RewardCurrency::burn(Zero::zero()); // hack to get new ImBalance with asset_id
 
 			let points = CurrentEraPointsEarned::take();
 			for (v, p) in validators.iter().zip(points.individual.into_iter()) {
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -716,6 +716,8 @@ decl_storage! {
 		/// The start of the current era.
 		pub CurrentEraStart get(fn current_era_start): MomentOf<T>;
 
+		CurrentEraDuration get(fn current_era_duration) : MomentOf<T>;
+
 		/// The session index at which the current era started.
 		pub CurrentEraStartSessionIndex get(fn current_era_start_session_index): SessionIndex;
 
@@ -1407,6 +1409,8 @@ impl<T: Trait> Module<T> {
 		let now = T::Time::now();
 		let previous_era_start = <CurrentEraStart<T>>::mutate(|v| sp_std::mem::replace(v, now));
 		let era_duration = now - previous_era_start;
+		<CurrentEraDuration<T>>::put(era_duration);
+
 		if !era_duration.is_zero() {
 			let validators = Self::current_elected();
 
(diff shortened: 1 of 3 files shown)


crml/staking/src/lib.rs
```diff
@@ -1399,25 +1401,17 @@ impl<T: Trait> Module<T> {
 		let now = T::Time::now();
 		let previous_era_start = <CurrentEraStart<T>>::mutate(|v| sp_std::mem::replace(v, now));
 		let era_duration = now - previous_era_start;
+		<CurrentEraDuration<T>>::put(era_duration);
```
jordy25519, line 1412: Does this need to be stored? can we pass it as a parameter into `current_total_payout` instead

crml/staking/src/lib.rs
```diff
@@ -1473,6 +1467,24 @@ impl<T: Trait> Module<T> {
 		maybe_new_validators
 	}
 
+	/// Compute current `total_payout` and `max_payout` for specific era duration
+	pub fn current_total_payout(total_tokens: RewardBalanceOf<T>) -> (RewardBalanceOf<T>, RewardBalanceOf<T>) {
```
jordy25519: for consistency: `total_tokens` -> `total_issuance`

Can also accept validators as a parameter here like with the `split_evenly_to_all_...` function

jordy25519 (commented):
🏅

crml/staking/src/lib.rs
mzxyz: Actually the `era_duration` is the key parameter we want to keep it same within the `new_era` function,  if we pass from outside, the test value can not be same with the one in `new_era`

cowboy-bebug: When reusing `fn current_total_payout` for tests, it's not clear what era_duration we need to pass

crml/staking/src/lib.rs
cowboy-bebug: The main reason for pulling the code out into `fn current_total_payout` was to reuse this function for our integration tests so we wanted to have minimal parameters to be passed (and rely on getter methods inside the function to have whatever we need in scope)

mzxyz: Yeah, also I think the validators should same with `current_elected`

crml/staking/src/lib.rs
```diff
@@ -1473,6 +1467,24 @@ impl<T: Trait> Module<T> {
 		maybe_new_validators
 	}
 
+	/// Compute current `total_payout` and `max_payout` for specific era duration
+	pub fn current_total_payout(total_issuance: RewardBalanceOf<T>) -> (RewardBalanceOf<T>, RewardBalanceOf<T>) {
```
cowboy-bebug, line 1479: If it's a security concern to expose this, we can make it private and make another method that wraps it and guard it behind #[test] (but I think it's okay)

jordy25519: it's ok since it's read only

crml/staking/src/lib.rs
jordy25519: It's possible to control the the duration in tests by configuring the timestamps `T::Time::now()` and `CurrentEraStart` values. This would allow tests to use some fixed duration for eras

runtime/tests/tests.rs
```diff
@@ -307,53 +286,110 @@ fn staking_genesis_config_works() {
...
+				assert_eq!(
+					GenericAsset::total_issuance(CENTRAPAY_ASSET_ID),
+					total_issuance + inflation_era_1
+				);
 
-			// Staking rewards are paid at the next era
-			for validator in validators {
+				// The balance of stash accounts remain the same within the same era
+				for validator in validators.clone() {
+					let (stash, _) = validator;
```
jordy25519:
```suggestion
				for (stash, _) in &validators {
```

runtime/tests/tests.rs
```diff
@@ -307,53 +286,110 @@ fn staking_genesis_config_works() {
...
+				assert_eq!(GenericAsset::total_issuance(CENNZ_ASSET_ID), total_issuance);
+				// Total issuance for CPAY remain the same within the same era
+				assert_eq!(
+					GenericAsset::total_issuance(CENTRAPAY_ASSET_ID),
+					total_issuance + inflation_era_1 + inflation_era_2
+				);
+
+				// The balance of stash accounts remain the same within the same era
+				for validator in validators.clone() {
+					let (stash, _) = validator;
```
jordy25519:
```suggestion
				for (stash, _) in &validators {
```

runtime/tests/tests.rs
```diff
@@ -307,53 +286,110 @@ fn staking_genesis_config_works() {
 }
 
 #[test]
-fn staking_reward_should_work() {
+fn staking_inflation_and_reward_should_work() {
```
jordy25519, line 289: As per the offline discussion: since this test doesn't do anything special outside the staking module it should be part of the staking test suite, it's not a full runtime integration test so shouldn't be here

jordy25519 (commented):
Last round of feedback from me ✌️

crml/staking/src/lib.rs
hoani: I believe the tests are configuring the timestamps of `T::Time::now()` by using `Timestamp::set_timestamp((System::block_number() * 1000).into());`, is that correct?

I suggest renaming from `CurrentEraDuration` to `PreviousEraDuration` - the value we are storing is the delta t between the start of the new era and the start of the previous era.

crml/staking/src/multi_token_economy_tests.rs
```diff
@@ -303,5 +303,12 @@ fn validator_reward_is_not_added_to_staked_amount_in_dual_currency_model() {
 				unlocking: vec![],
 			})
 		);
-	});
+		// Check total issuance
+		let total_issuance = 1_000_000_000 * 2; // one stash and controller accounts
+		assert_eq!(GenericAsset::total_issuance(STAKING_ASSET_ID), total_issuance);
+		assert_eq!(
+			GenericAsset::total_issuance(REWARD_ASSET_ID),
+			total_issuance + total_payout_0
```
hoani: [very optional suggestion] nitpick: rename `total_payout_0` to `total_payout`

runtime/tests/tests.rs
```diff
@@ -385,8 +421,7 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
 
 			start_era(1);
 			let validator_len = validators.len() as Balance;
-			let (total_staking_reward, _) = reward_validators(&validators);
-			let per_staking_reward = total_staking_reward / validator_len;
+			reward_validators(&validators);
```
hoani, line 420: [out-of-scope] Not for this PR, but something to consider for later @holygits , this test is a little soft when we are awarding all validators the same points.
i.e. it doesn't answer "are all validators awarded the same fee rewards when their points are different?"

hoani (approved):
LGTM
Some minor nitpicks to consider

crml/staking/src/lib.rs
mzxyz: > It's possible to control the the duration in tests by configuring the timestamps `T::Time::now()` and `CurrentEraStart` values. This would allow tests to use some fixed duration for eras

For the mock config in `multi_token_economy_tests`, it does config fix duration, but for our test we use `runtime` directly, so we don't find a good way to solve this

runtime/tests/tests.rs
cowboy-bebug: Yep will be addressed in a follow up PR

crml/staking/src/lib.rs
cowboy-bebug: Will be addressed in a follow up PR

crml/staking/src/lib.rs
cowboy-bebug: Will be addressed in a follow up PR (if possible)

runtime/tests/tests.rs
mzxyz: Yeah, we could move it and also the transaction payment test into `multi_token_test` in staking module in another pr

runtime/tests/tests.rs
cowboy-bebug: This will be addressed in the same followup PR

mzxyz: Good point, we could  add different points to validators and check the free balance in the future

cowboy-bebug (approved):

PR #164, opened 2020-03-26
Add minimum_bond and make it configurable

cowboy-bebug:
The purpose of this PR is to decouple the use of GA::minimum_balance(), which is currently zero. The solution is to add a configurable storage, which is (1) initialised from genesis config and (2) could be overridden by root.

## Changes:
- add private `MinimumBond` storage with a get method
- add `SetMinimumBond` event to notify when minimum_bond is set by root (via set_minimum_bond dispatch method)
- add `fn set_minimum_bond` dispatch method that requires root permission
- set default minimum_bond to be 1 in GenesisConfig (so the bond amount is at least 1)
- update condition in `fn bond` to use Self::minimum_bond
- add minimum_bond field in ExtBuilder mock and a setter for it
- update existing test to show the above works

---

## Some maintainability / refactor concerns:
The following refactors (from staking audit) could also be addressed in this PR:
- rename `BalanceOf<T>` to `StakingBalanceOf<T>` as the name is confusing (compared to `RewardBalanceOf<T>` which is really clear)
- remove migration stuff
- change all root dispatch calls with `#[weight = SimpleDispatchInfo::FixedOperational(0)]`

## Other concerns:
- The dispatch method, `fn unbond`, has a logic to not introduce any dust (if staked amount < minimum_balance then unbond the whole staked amount): https://github.com/cennznet/cennznet/blob/8bb5fa31c379418c71cdcbd0db82f300ab032f72/crml/staking/src/lib.rs#L999-L1002 this part is left alone because it makes sense to configure what qualifies as a dust inside generic asset
- Unit test(s) are not added because `bond_with_no_staked_value` test already covers it

### DIFF
--- a/cli/src/chain_spec/mod.rs
+++ b/cli/src/chain_spec/mod.rs
@@ -160,6 +160,7 @@ pub fn config_genesis(network_keys: NetworkKeys, enable_println: bool) -> Genesi
 				.map(|x| (x.0.clone(), x.1.clone(), STASH, StakerStatus::Validator))
 				.collect(),
 			invulnerables: initial_authorities.iter().map(|x| x.0.clone()).collect(),
+			minimum_bond: 1,
 			slash_reward_fraction: Perbill::from_percent(10),
 			..Default::default()
 		}),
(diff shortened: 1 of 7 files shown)


jordy25519 (approved):
gj, couple things
- Add an assert to check new event is logged
- [Optional] Can make an assert in the genesis constructor for `MinimumBond` so that it is compile error if zero (see balances existential deposit for an example)

hoani (approved):
LGTM very nice and clean changes 👍

PR #166, opened 2020-03-27
Feature/bt 183 commission cap

hoani:
This PR prevents validators from setting commissions higher than 100%.
Validator preferences are only set on the `validate` extrinsic, so this bound is added there.

## Changes:
- limit validator `commission` in `validate` extrinsic
- add test case to show larger commissions are capped at 100%.
- add protection to `reward_validator` to protect against paying out more than `reward` as commission

---

- If validator prefs are updated to include other fields, the implementation in `validate` also needs to be updated
- If we add other ways to update `commission` (eg, a `commission` extrinsic), then we would have to add the limit there too - might want to refactor that case at that time.


jordy25519 (approved):
LGTM!

cowboy-bebug (approved):
Nice! LGTM!

PR #170, opened 2020-03-27
Only update EarliestUnappliedSlash when there are unapplied slashes

hoani:
Previously, `on_offence` would update `EarliestUnappliedSlash` regardless of whether there were actually unapplied slashes being registered.
Raised in Staking module audit

## Changes:
- move `EarliestUnappliedSlash::mutate` to occur when `UnappliedSlashes::mutate` occurs.

---

- the testing around deferred slashes is pretty light, it's not even clear that the `EarliestUnappliedSlash` needs to be accurate.


cowboy-bebug (approved):
LGTM

mzxyz (approved):
LGTM

PR #173, opened 2020-03-29
Remove unnecessary passing around the fee rate in cennzx-spot

hoani:
Based on refactor opportunities identified in BT-148

## Addressed:
- `get_input/output_price` should not include the fee rate

---

## Maintainability / refactor concerns:
- None

PR #174, opened 2020-03-29
Refactor - cennzx simplify buy sell

hoani:
Based on refactor opportunities identified in BT-148

Should merge after #173 is merged.

Cut down the lines of code from ~1000 to ~700 🎉 

## Addressed:
- Consolidate duplicated code between all `make_<>_to_<>_<output/input>` into two functions
- Replace the `input` and `output` convention with `sell` and `buy`

## Major changes:
- All `make_<>_to_<>_<output/input>` functions removed
- All `get_<>_to_<>_input_price` functions renamed to `get_<>_to_<>_sell_price`
- All `get_<>_to_<>_output_price` functions renamed to `get_<>_to_<>_buy_price`
- Removed 12 unnecessary cennzxspot error definitions
- Replaced `make_asset_swap_output` with `execute_buy`
- Replaced `make_asset_swap_input` with `execute_sell`
- Added `execute_trade` to remove duplication between `buy` and `sell` trades

---

- Adding `execute_trade` means that the `core_asset` which is traded is always calculated using `get_asset_to_core_sell_price`, in the past, the `core_asset` to be traded would be calculated by either `get_core_to_asset_buy_price` when buying or `get_asset_to_core_sell_price` when selling. This discrepancy caused an off-by one error for core volumes in some `buy` tests. I prefer the new method of calculating `core_asset` because it should not matter whether we are executing on a `buy` or `sell` extrinsic into the exchange.


cowboy-bebug (approved):
LGTM !

crml/cennzx-spot/src/impls.rs
```diff
@@ -142,7 +141,11 @@ pub(crate) mod impl_tests {
 				(price * fee_rate_factor) / scale_factor // price adjusted with fee
 			};
 
-			assert_eq!(core_asset_price, 539);
+			// core asset price is now calculated the same way whether we are buying
```
aliXsed, line 144: I imagine this comment wouldn't make much sense later when there is no reference to the previous method of calculation. It's a good comment for the PR, though. So I suggest you take out from here.

crml/cennzx-spot/src/lib.rs
```diff
@@ -860,86 +517,149 @@ impl<T: Trait> Module<T> {
...
-				sell_amount,
-				min_receive,
-				fee_rate,
-			)?;
-		}
+				amount_to_buy,
+			));
+		};
+
+		Self::deposit_event(RawEvent::AssetPurchase(
```
jordy25519, line 656: We'll want to consider renaming this Event now

crml/cennzx-spot/src/lib.rs
```diff
@@ -860,86 +517,149 @@ impl<T: Trait> Module<T> {
...
+			asset_to_buy,
+			amount_to_sell,
+			amount_to_buy,
+		)?;
+
+		Ok(amount_to_buy)
+	}
+
+	fn execute_trade(
+		seller: &T::AccountId,
```
jordy25519, line 602: maybe `trader`?

crml/cennzx-spot/src/lib.rs
```diff
@@ -860,86 +517,149 @@ impl<T: Trait> Module<T> {
...
+	/// Buy `amount_to_buy` of `asset_to_buy` with `asset_to_sell`.
+	///
+	/// `seller` - Account selling `asset_to_sell`
+	/// `recipient` - Account to receive `asset_to_buy`
+	/// `asset_to_sell` - asset ID to sell
+	/// `asset_to_buy` - asset ID to buy
+	/// `amount_to_buy` - The amount of `asset_to_buy` to buy
+	/// `maximum_sell` - Maximum acceptable amount of `asset_to_sell` the seller will sell
+	pub fn execute_buy(
+		seller: &T::AccountId,
```
jordy25519, line 534: maybe `buyer`?

crml/cennzx-spot/src/lib.rs
```diff
@@ -860,86 +517,149 @@ impl<T: Trait> Module<T> {
...
+			)
+			.and(<pallet_generic_asset::Module<T>>::make_transfer(
+				&asset_to_buy,
+				&exchange_address,
+				recipient,
+				amount_to_buy,
+			));
 		} else {
-			let _ = Self::make_asset_to_asset_input(
+			let core_volume = Self::get_asset_to_core_sell_price(asset_to_sell, amount_to_sell)?;
```
jordy25519, line 632: core_liquidity

jordy25519 (approved):
TY!! some comments on naming as it does tend to be confusing in this module but not-blockers IMO

aliXsed:
Though I can see the improvement that this change is offering and its necessity, I think the change-set cannot be considered a refactor technically. This could be more efficiently reviewed if we had the refactor and the change of the functionality (for which tests needed to be corrected) in separate PRs or at least in separate commits. Having said that, this criticism is valid for at least one of my own previous PRs in a different way, where a refactor had occurred alongside adding a feature. So it's a reminder/improvement point for myself too.

hoani:
Merged into https://github.com/cennznet/cennznet/pull/177

PR #169, opened 2020-03-27
Refactor slash_out_of out of slash as apply_slash

hoani:
Pull unnecessary `slash_out_of` closure out of the `StakingLedger`'s `slash` function
Raised in Staking module audit

## Changes:
- new method `apply_slash`
- clarified the logic previously in `slash_out_of`

---

## Maintainability / refactor concerns:
- none

PR #175, opened 2020-03-29
double review and fix underflow in staking module

mzxyz:
This PR try to review all subtraction operation in the staking module and replace all subtraction operation may have risk with `saturating_sub`.

## changes
After check all the subtraction operations in `Staking module`. There only one place need to be changed. Other places have already make sure safety.

## maintainability / refactor
None

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -471,7 +471,7 @@ where
 				}
 
 				*total_remaining = total_remaining.saturating_sub(slash_from_target);
-				*value -= slash_from_target;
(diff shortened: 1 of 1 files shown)


hoani (commented):
Nice work Michael,
I suggest that we bring in this PR instead: https://github.com/cennznet/cennznet/pull/169/files
It also removes the need for this subtraction

cowboy-bebug:
As @hoani pointed out this wasn't an issue in the first place due to the line below:
```rust
let mut slash_from_target = (*value).min(*target);
// `slash_from_target` can never be greater than `value`
```

PR #176, opened 2020-03-29
Migrate to plug 1.0.0-rc3.1

jordy25519:
~Super suspicious about this since it worked with basically no errors 🤷~
Suspicions confirmed by @hoani as Cargo.lock was not pulling in new branch.

Important runtime changes from upstream:
- New storage hashers which store keys iterable by default (removes need for linked_map)
Can use faster XX hash for storage maps where trie unbalancing is not a concern (making a storage key is expensive for user, or uncontrollable by user)
Use iterable hashes for map storage and more efficient versions where security allows
- parity-scale-codec bumped to 1.3.0
- Removed `validate_trusted_extrinsic` <- was a bad idea in the first place
- `Phase::ApplyExtrinsic(0)` -> `Phase::Initialization`

as part of this migration judiciously removed un-needed things to reduce conflicts:
- Remove SlashCancelOrigin from staking
- Remove Democracy Module (too many confusing updates from upstream + don't need)
- Remove benchmarking code from runtime (don't need for launch, can make task to re-add)
- An ignored test in cli/service.rs (needs some attention to be updated and enabled on our CI, can make task for this also)


src/main.rs
```diff
@@ -27,7 +27,7 @@ fn main() -> CliResult<()> {
 		version: env!("CARGO_PKG_VERSION"),
 		executable_name: "cennznet",
 		author: "Centrality Developers <support@centrality.ai>",
-		description: "CENNZnet node",
+		description: "CENNZnet node by Centrality Investments UNlimited",
```
cowboy-bebug, line 30: 👍

testing/src/genesis.rs
```diff
@@ -101,6 +101,7 @@ pub fn config_endowed(support_changes_trie: bool, code: Option<&[u8]>, extra_end
 			],
 			validator_count: 3,
 			minimum_validator_count: 0,
+			minimum_bond: 1,
```
cowboy-bebug: This line needs to be removed since it's duplicated with line 107

runtime/tests/mock.rs
```diff
@@ -143,6 +143,7 @@ impl ExtBuilder {
 		crml_staking::GenesisConfig::<Runtime> {
 			minimum_bond: 1,
 			current_era: 0,
+			minimum_bond: 1,
```
cowboy-bebug: This line is repeated

jordy25519: Ah I fixed, then rebased so that must be where this came from

cowboy-bebug (approved):
LGTM

hoani (dismissed):
LGTM

aliXsed:
Please remember, if Plug rc3.1 is different from Plug rc3 in one of the following modules, then the local cennznet modules should be updated by bringing in the same/similar changes:
cli, staking, transaction-payments, executor, primitives, rpc, rpc-client, runtime and testing. 
https://centralitydev.atlassian.net/wiki/spaces/HOW/pages/887947277/How+To+Migrate+Plug+into+Cennznet

cli/src/service.rs
```diff
@@ -450,171 +450,6 @@ mod tests {
 		);
 	}
 
-	#[test]
```
jordy25519, line 453: Ain't no time to fix this ignored test

crml/cennzx-spot/src/impls.rs
```diff
@@ -40,17 +40,13 @@ where
 	fn exchange_address_for(core_asset_id: T::AssetId, asset_id: T::AssetId) -> T::AccountId {
 		let mut buf = Vec::new();
 		buf.extend_from_slice(b"cennz-x-spot:");
-		buf.extend_from_slice(&u64_to_bytes(core_asset_id.into()));
```
jordy25519, line 43: removed un-needed extra function here

cli/src/cli.rs
```diff
@@ -48,10 +48,6 @@ pub enum Subcommand {
 		about = "Decode given block or extrinsic using current native runtime."
 	)]
 	Inspect(node_inspect::cli::InspectCmd),
-
-	/// The custom benchmark subcommmand benchmarking runtime pallets.
```
jordy25519, line 52: https://github.com/cennznet/cennznet/issues/178

runtime/src/lib.rs
```diff
@@ -913,34 +864,4 @@ mod tests {
 		is_submit_signed_transaction::<SubmitTransaction>();
 		is_sign_and_submit_transaction::<SubmitTransaction>();
 	}
-
-	#[test]
-	fn block_hooks_weight_should_not_exceed_limits() {
```
jordy25519, line 918: Test got removed upstream after some changes to block weight involving `on_initialize` and `on_finalize`

cli/src/cli.rs
jordy25519: Have added comments with tracking issue to these parts. Will push if after build passes

cowboy-bebug (approved):
LGTM!

crml/cennzx-spot/src/impls.rs
```diff
@@ -40,17 +40,13 @@ where
...
-		buf.extend_from_slice(&u64_to_bytes(asset_id.into()));
+		buf.extend_from_slice(&core_asset_id.into().to_le_bytes());
```
hoani, line 43: I don't think this is a good idea - AssetID is a u32.
Originally, this would extend buf by 8 bytes, as far as I can tell, this change extends it by 4.

jordy25519: `.into()` makes it into a `u64` so the behaviour should be the same, just one less indirection

PR #180, opened 2020-04-01
Sylo module audit refactor

cowboy-bebug:
This PR is to refactor Sylo module based on the corresponding module audit.

## Changes:

*For reviewing the PR, it will be easier to look at individual commit and its comments rather than viewing all the file changes.*

The below should follow the commit order:
- Add the familiar ExtBuilder pattern in mock and associated changes to all tests
- `devices.rs`
    - Remove unused Event and related code (reason is the purpose of Hash type in the defined event wasn’t clear)
    - Add Error to replace string-based errors
    - Add DeviceId as type alias for u32
    - Add correct length check inside `fn append_device`
    - Add delete_device method to de-register devices
    - Add two unit tests, one for each method
- `inbox.rs`
    - Add arbitrary usize consts to limit messages, associated ensures and Error
    - Add `MessageId` and `Message` type aliases for better readability
    - Remove redundant Values mutation on next_index match
- `vault.rs`
    - Rename KEYS_MAX -> MAX_KEYS for consistency
    - Add MAX_DELETE_KEYS const to limit number of deletes per call
    - Add Error and associated ensures
    - Add a test for adding more than allowed
- `groups`
    - Add MAX_MEMBERS const to limit the number of members in a group and associated ensure
    - Rename INVITES_MAX -> MAX_INVITES for consistency
- `e2ee.rs`
    - Removed unused Event and related code
    - Use already existing DeviceId type alias for u32
    - Add Error to replace string
    - Add `fn check_total_pkbs` method and use it to check whether more pkbs can be added to the storage

## Concerns:

- The const limits are arbitrary at the moment and we need to discuss what they should be

### DIFF
--- a/crml/sylo/src/device.rs
+++ b/crml/sylo/src/device.rs
@@ -13,49 +13,122 @@
 *     https://centrality.ai/licenses/lgplv3.txt
 */
 
-use frame_support::{decl_event, decl_module, decl_storage, dispatch::DispatchResult, dispatch::Vec, ensure};
-use frame_system;
+use frame_support::{decl_error, decl_module, decl_storage, dispatch::DispatchResult, dispatch::Vec, ensure};
 
 const MAX_DEVICES: usize = 1000;
 
-pub trait Trait: frame_system::Trait {
-	/// The overarching event type.
-	type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
-}
+type DeviceId = u32;
+
+pub trait Trait: frame_system::Trait {}
 
 decl_module! {
 	pub struct Module<T: Trait> for enum Call where origin: T::Origin, system = frame_system {
-		fn deposit_event() = default;
+		type Error = Error<T>;
 	}
 }
 
 // The data that is stored
 decl_storage! {
 	trait Store for Module<T: Trait> as SyloDevice {
-		pub Devices get(devices): map hasher(blake2_128_concat) T::AccountId => Vec<u32>;
+		pub Devices get(devices): map hasher(blake2_128_concat) T::AccountId => Vec<DeviceId>;
 	}
 }
 
-decl_event!(
-	pub enum Event<T> where <T as frame_system::Trait>::Hash, <T as frame_system::Trait>::AccountId {
-		DeviceAdded(AccountId, Hash, u32),
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// There are no devices registered for user (missing user_id in Devices)
+		UserId,
+		/// Device is already registered to user (device_id is already in use)
+		DeviceId,
+		/// A user can't have more than MAX_DEVICES registered devices
+		MaxDevice,
 	}
-);
+}
 
 impl<T: Trait> Module<T> {
-	pub fn append_device(user_id: &T::AccountId, device_id: u32) -> DispatchResult {
+	pub fn append_device(user_id: &T::AccountId, device_id: DeviceId) -> DispatchResult {
 		let mut devices = <Devices<T>>::get(user_id);
 
-		ensure!(!devices.contains(&device_id), "Device Id already in use");
-		ensure!(
-			devices.len() <= MAX_DEVICES,
-			"User has registered up to the maximum number of devices"
-		);
+		ensure!(!devices.contains(&device_id), Error::<T>::DeviceId);
+		ensure!(devices.len() < MAX_DEVICES, Error::<T>::MaxDevice);
 
 		devices.push(device_id);
 
 		<Devices<T>>::insert(user_id, devices);
 
 		Ok(())
+	}
+
+	pub fn delete_device(user_id: &T::AccountId, device_id: DeviceId) -> DispatchResult {
+		ensure!(<Devices<T>>::contains_key(user_id), Error::<T>::UserId);
+		let mut devices = <Devices<T>>::take(user_id);
+		devices.retain(|device| *device != device_id);
+		<Devices<T>>::insert(user_id, devices);
+		Ok(())
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::mock::{ExtBuilder, Test};
+	use frame_support::{assert_noop, assert_ok};
+	use sp_core::H256;
+
+	type Device = Module<Test>;
+
+	#[test]
+	fn append_device_works() {
+		ExtBuilder.build().execute_with(|| {
+			let user_id = H256::from_low_u64_be(1);
+			let device_id = 7357;
+
+			// simple appending of a device works
+			assert_ok!(Device::append_device(&user_id, device_id));
+			assert_eq!(Device::devices(user_id), vec![device_id]);
+
+			// adding the same device should return error
+			assert_noop!(Device::append_device(&user_id, device_id), Error::<Test>::DeviceId);
+
+			// adding more than MAX_DEVICES is not allowed
+			let new_devices: Vec<_> = ((device_id + 1)..(device_id + MAX_DEVICES as DeviceId)).collect();
+			assert_eq!(new_devices.len(), 999); // length assert is here in case MAX_DEVICES changes
+			for new_device in new_devices.clone() {
+				assert_ok!(Device::append_device(&user_id, new_device));
+			}
+			assert_eq!(Device::devices(user_id).len(), 1000);
+			assert_noop!(Device::append_device(&user_id, 123), Error::<Test>::MaxDevice);
+		});
+	}
+
+	#[test]
+	fn delete_device_works() {
+		ExtBuilder.build().execute_with(|| {
+			let user_id = H256::from_low_u64_be(1);
+			let mut devices = vec![1, 2, 3, 4, 5];
+			for device in devices.clone() {
+				assert_ok!(Device::append_device(&user_id, device));
+			}
+			for device in devices.clone().into_iter().rev() {
+				assert_ok!(Device::delete_device(&user_id, device));
+				devices.pop().unwrap();
+				assert_eq!(Device::devices(user_id), devices);
+			}
+			assert_eq!(Device::devices(user_id), vec![]);
+		});
+	}
+
+	#[test]
+	fn delete_non_existing_device_works() {
+		ExtBuilder.build().execute_with(|| {
+			let user_id = H256::from_low_u64_be(1);
+			let devices = vec![1, 2, 3, 4];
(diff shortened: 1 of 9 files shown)


crml/sylo/src/device.rs
```diff
@@ -36,26 +32,87 @@ decl_storage! {
...
 	pub fn append_device(user_id: &T::AccountId, device_id: u32) -> DispatchResult {
 		let mut devices = <Devices<T>>::get(user_id);
 
-		ensure!(!devices.contains(&device_id), "Device Id already in use");
-		ensure!(
-			devices.len() <= MAX_DEVICES,
-			"User has registered up to the maximum number of devices"
-		);
+		ensure!(!devices.contains(&device_id), Error::<T>::DeviceId);
+		ensure!(devices.len() < MAX_DEVICES, Error::<T>::MaxDevice);
```
cowboy-bebug: Hard to see if the diff but the length check was incorrect before as it was possible to add `MAX_DEVICES + 1` devices

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,108 @@
...
+			let mut devices = vec![1, 2, 3, 4, 5];
+			for device in devices.clone() {
+				assert_ok!(Device::append_device(&user_id, device));
+			}
+			for device in devices.clone().into_iter().rev() {
+				assert_ok!(Device::delete_device(&user_id, device));
+				devices.pop().unwrap();
+				assert_eq!(Device::devices(user_id), devices);
+			}
+			assert_eq!(Device::devices(user_id), vec![]);
```
jordy25519: `assert!(Device::devices(user_id).is_empty());`

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,108 @@
...
+		ExtBuilder.build().execute_with(|| {
+			let user_id = H256::from_low_u64_be(1);
+			let device_id = 7357;
+
+			// simple appending of a device works
+			assert_ok!(Device::append_device(&user_id, device_id));
+			assert_eq!(Device::devices(user_id), vec![device_id]);
+
+			// adding the same device should return error
+			assert_noop!(Device::append_device(&user_id, device_id), Error::<Test>::DeviceId);
```
jordy25519: Should probably be two different test cases (duplicate device registration vs. max device registration)

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,108 @@
...
 	}
 }
 
-decl_event!(
-	pub enum Event<T> where <T as frame_system::Trait>::Hash, <T as frame_system::Trait>::AccountId {
-		DeviceAdded(AccountId, Hash, u32),
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// There are no devices registered for user (missing user_id in Devices)
+		UserId,
```
jordy25519: The error types should be self describing (also that way they'll give a good error string to the API/client)
```
UserIdNotRegistered,
DeviceIdExists,
MaxDeviceLimitReached,
```

crml/sylo/src/device.rs
hoani: Eagle eye! Nice catch!

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,108 @@
...
+	use frame_support::{assert_noop, assert_ok};
+	use sp_core::H256;
+
+	type Device = Module<Test>;
+
+	#[test]
+	fn append_device_works() {
+		ExtBuilder.build().execute_with(|| {
+			let user_id = H256::from_low_u64_be(1);
+			let device_id = 7357;
```
hoani, line 84: Nice 😆

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,108 @@
...
+			for device in devices.clone() {
+				assert_ok!(Device::append_device(&user_id, device));
+			}
+			for device in devices.clone().into_iter().rev() {
+				assert_ok!(Device::delete_device(&user_id, device));
+				devices.pop().unwrap();
+				assert_eq!(Device::devices(user_id), devices);
+			}
+			assert_eq!(Device::devices(user_id), vec![]);
+		});
```
hoani, line 144: [optional] Could be nice to add a check that deleting a device that isn't on the list doesn't affect the list.
eg: `Device::devices(user_id) = [1,2,3,4]; Device::delete_device(&user_id, 5); assert_eq!(Device::devices(user_id), [1,2,3,4]);`

crml/sylo/src/inbox.rs
```diff
@@ -13,57 +13,73 @@
...
 		account_values.push((peer_id.clone(), next_index));
 
 		// Store data
 		let mut values = <Values<T>>::get(&peer_id);
-		if let Some((i, _)) = values.iter().enumerate().find(|(_, item)| item.0 == next_index) {
-			values[i] = (next_index, value);
-		} else {
-			values.push((next_index, value));
-		}
+		values.push((next_index, value));
```
hoani, line 82: 🎉

hoani (approved):
Excellent improvements
Nice catch on `ext_builder` and many other great changes!
👍

crml/sylo/src/device.rs
cowboy-bebug: Added `fn delete_non_existing_device_works` to show deleting a non-existing device passes through without an error

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,122 @@
...
 	}
 }
 
-decl_event!(
-	pub enum Event<T> where <T as frame_system::Trait>::Hash, <T as frame_system::Trait>::AccountId {
-		DeviceAdded(AccountId, Hash, u32),
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// There are no devices registered for user (missing user_id in Devices)
+		UserId,
```
roy-powerfinance: I think we should have a better name for these errors - i.e. UserId should be UserIdMissing

crml/sylo/src/inbox.rs
```diff
@@ -13,57 +13,73 @@
...
+pub trait Trait: frame_system::Trait {}
+
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// An inbox cannot store more than MAX_MESSAGES where each message is smaller than MAX_MESSAGE_LENGTH
+		Message,
+		/// Cannot delete more than MAX_DELETE_MESSAGES at a time
+		Delete,
+		/// Cannot assign any more ids to message due to overflow
+		IdOverflow,
```
roy-powerfinance: same comments as before - better error names

roy-powerfinance (changes requested):

crml/sylo/src/inbox.rs
```diff
@@ -13,57 +13,73 @@
...
+		ensure!(next_index != u32::max_value(), Error::<T>::IdOverflow);
 		let mut account_values = <AccountValues<T>>::get(&peer_id);
+		ensure!(account_values.len() < MAX_MESSAGES, Error::<T>::Message);
 
 		// Add new mapping to account values
 		account_values.push((peer_id.clone(), next_index));
 
 		// Store data
 		let mut values = <Values<T>>::get(&peer_id);
-		if let Some((i, _)) = values.iter().enumerate().find(|(_, item)| item.0 == next_index) {
```
jordy25519, line 62: random note I think: item.0 here is a bug as its the first element of the vec 🐛

crml/sylo/src/device.rs
cowboy-bebug: Yep will change to `UserIdMissing`

crml/sylo/src/inbox.rs
cowboy-bebug: I wanted to keep the errors general but I'll change it to `MessageIdOverflow`

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,122 @@
...
 
-decl_event!(
-	pub enum Event<T> where <T as frame_system::Trait>::Hash, <T as frame_system::Trait>::AccountId {
-		DeviceAdded(AccountId, Hash, u32),
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// There are no devices registered for user (missing user_id in Devices)
+		UserIdMissing,
+		/// Device is already registered to user (device_id is already in use)
+		DeviceId,
```
jordy25519: Should make these self descriptive since they generate strings for the client e.g. `DeviceIdExists`

jordy25519 (commented):
The error types should be self-descriptive but otherwise looks good!

crml/sylo/src/device.rs
jordy25519: Sorry mentioned this multiple times lol

crml/sylo/src/device.rs
cowboy-bebug: Ah yeah that's more idiomatic!

crml/sylo/src/inbox.rs
cowboy-bebug: Ah yeah that's true. The tuple is already unpacked by `(_, item)` and item is actually the message (not the message_id) 🐛

crml/sylo/src/device.rs
```diff
@@ -13,49 +13,148 @@
...
-	pub enum Event<T> where <T as frame_system::Trait>::Hash, <T as frame_system::Trait>::AccountId {
-		DeviceAdded(AccountId, Hash, u32),
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// There are no devices registered for user (missing user_id in Devices)
+		UserIdNotRegistered,
+		/// Device is already registered to user (device_id is already in use)
+		DeviceIdExists,
+		/// A user can't have more than MAX_DEVICES registered devices
+		MaxDeviceLimitReached,
```
roy-powerfinance, line 44: 👍

crml/sylo/src/inbox.rs
```diff
@@ -13,57 +13,73 @@
...
+const MAX_DELETE_MESSAGES: usize = 10_000;
+
+type MessageId = u32;
+type Message = Vec<u8>;
+
+pub trait Trait: frame_system::Trait {}
+
+decl_error! {
+	pub enum Error for Module<T: Trait> {
+		/// An inbox cannot store more than MAX_MESSAGES where each message is smaller than MAX_MESSAGE_LENGTH
```
roy-powerfinance: can you change these as well?

roy-powerfinance (approved):

PR #182, opened 2020-04-01
Cennzx-spot exchange_address_for only takes trade assetId

hoani:
This is the result of an audit on CENNZxSpot

## Addresses:

- exchange key generation should only require the trading `asset_id`

## Changes:

- `exchange_address_for` only requires trading `asset_id`
- the exchange address still uses the `CoreAssetId` from storage to keep the addresses consistent

---

## Concerns:

- none


mzxyz (approved):
Nice, LGTM

jordy25519 (approved):

PR #183, opened 2020-04-01
CENNZx-Spot Test Module Tidy Up Refactor

hoani:
This is the result of an audit on CENNZxSpot

## Addresses:

- test names should use `buy` and `sell` instead of `output` and `input`
- most of test setup should be in `mock.rs` not `test.rs`
- `get_output_price_zero_cases` should try to buy less asset than what is being set in the pool

---

## Concerns:

- none


jordy25519 (approved):
Much cleaner!

cowboy-bebug (approved):
LGTM!

PR #187, opened 2020-04-04
Implement CENNZnut Contract Validation

hoani:
This PR implements the `verify_runtime_to_contract_call` and `verify_contract_to_contract_call` checks for CENNZnuts on CENNZnet.

The goal is an MVP which allows a holder to delegate a contract call to the issuer.

Before this PR is merged, cennznut-rs PR: https://github.com/cennznet/cennznut-rs/pull/47 must be merged and the reference to CENNZnut in the runtime `cargo.toml` must be updated to point to the correct location.

## Changes:

- Update cennznet to use the latest `cennznut-rs` interface (some changes were also required on `cennznut-rs` to expose the required types
- Implement `verify_runtime_to_contract_call`
- Implement `verify_contract_to_contract_call`
- Add some Runtime doughnut/cennznut helpers to reduce duplication
- Have added integration tests

## Concerns:

- None

### DIFF
--- a/runtime/src/impls.rs
+++ b/runtime/src/impls.rs
@@ -21,16 +21,20 @@ use cennznet_primitives::{
 	traits::{BuyFeeAsset, IsGasMeteredCall},
 	types::{Balance, FeeExchange},
 };
+use cennznut::{CENNZnut, RuntimeDomain, ValidationErr};
+use codec::Decode;
 use crml_transaction_payment::GAS_FEE_EXCHANGE_KEY;
 use frame_support::{
-	storage,
+	additional_traits, storage,
 	traits::{Currency, ExistenceRequirement, Get, Imbalance, OnUnbalanced, WithdrawReason},
 	weights::Weight,
 };
 use pallet_contracts::{Gas, GasMeter};
 use pallet_generic_asset::StakingAssetCurrency;
 use sp_runtime::{
-	traits::{CheckedMul, CheckedSub, Convert, SaturatedConversion, Saturating, UniqueSaturatedFrom, Zero},
+	traits::{
+		CheckedMul, CheckedSub, Convert, PlugDoughnutApi, SaturatedConversion, Saturating, UniqueSaturatedFrom, Zero,
+	},
 	DispatchError, Fixed64, Perbill,
 };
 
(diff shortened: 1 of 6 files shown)


runtime/tests/doughnut.rs
```diff
@@ -44,40 +47,66 @@ fn verify_dispatch(doughnut: &CennznetDoughnut, module: &str, method: &str) -> R
...
 		methods: vec![(method.to_string(), method_obj)],
 	};
 	CENNZnut::V0(CENNZnutV0 {
 		modules: vec![(module.to_string(), module_obj)],
+		contracts: Default::default(),
+	})
+}
+
+// A helper to make test CENNZnuts
+fn make_contract_cennznut(contract_addr: AccountId) -> CENNZnut {
```
mzxyz: [Suggestion]: make type of `contract_addr` be `&AccountId`.

runtime/tests/doughnut.rs
```diff
@@ -152,10 +181,104 @@ fn it_fails_when_module_name_is_empty() {
...
 	let doughnut = make_doughnut("cennznet", cennznut.encode());
 	assert_err!(
 		verify_dispatch(&doughnut, "-", "attest"),
 		"error during module name segmentation"
 	);
 }
+
+#[test]
+fn it_fails_when_using_contract_cennznut_for_runtime() {
+	let cennznut = make_contract_cennznut([0x11; 32].into());
```
mzxyz: [Suggestion] Make global var for`[0x11; 32]`, same with `const TEST_HOLDER: ...`, like `CONTRACT_ADDRESS`

runtime/src/lib.rs
```diff
@@ -545,6 +545,45 @@ impl frame_system::offchain::CreateTransaction<Runtime, UncheckedExtrinsic> for
...
+
+	/// Verify that the smart contract being called is permissioned in the cennznut
+	fn check_contract(
+		contract_address: &<Runtime as frame_system::Trait>::AccountId,
+		cennznut: CENNZnut,
+	) -> Result<(), &'static str> {
+		let address: [u8; 32] = contract_address.clone().into();
+		match cennznut.validate_contract_call(&address) {
+			Ok(r) => Ok(r),
+			Err(ValidationErr::NoPermission(ContractDomain::Contract)) => {
```
mzxyz: Any reason for checking two different conditions but return same err?

mzxyz (commented):
Add some suggestions, otherwise all changes LGTM

jordy25519:
This is a migration goof, I've fixed it before but seems to have come back #129 
```
pub type CennznetDoughnut = prml_doughnut::PlugDoughnut<Runtime>;
```

runtime/src/lib.rs
```diff
@@ -545,6 +545,45 @@ impl frame_system::offchain::CreateTransaction<Runtime, UncheckedExtrinsic> for
 	}
 }
 
+/// `DelegatedDispatchVerifier` helpers
+impl Runtime {
```
jordy25519: Really the helpers are focused on the doughnut itself so I'd consider making a new extension trait for `PlugDoughnut` and implementing the CENNZnut helpers within that, which would also contain the logic more cleanly.

Doesn't feel right to put this on the `Runtime` struct

Side note: we should have smth like `struct CENNZnetDispatchVerifier<Runtime>` which impls `DelegatedDispatchVerifier` to contain the logic more cleanly in general

runtime/src/lib.rs
hoani: That's a good point - could just make it:
`_ => Err("CENNZnut does not grant permission for contract")`

mzxyz (approved):

runtime/src/lib.rs
```diff
@@ -549,26 +573,60 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
...
+			Err(ValidationErr::NoPermission(RuntimeDomain::MethodArguments)) => {
 				Err("CENNZnut does not grant permission for method arguments")
 			}
 		}
 	}
+
+	/// Verify that the contract being called is permissioned on the doughnut
+	/// This is used in two cases:
+	/// * Case 1 - The `holder` of a doughnut wants to invoke a smart contract as the `issuer`
+	/// * Case 2 - The `issuer` wants to execute a contract but only with the permissions avaliable
```
jordy25519: Suggest reworking this comment to focus Case 1 and remove Case 2 for now

runtime/src/lib.rs
```diff
@@ -549,26 +573,60 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
...
+	///            in the doughnut being sent. (not implemented yet)
+	fn verify_runtime_to_contract_call(
+		caller: &Self::AccountId,
+		doughnut: &Self::Doughnut,
+		contract_addr: &Self::AccountId,
+	) -> Result<(), &'static str> {
+		if caller.clone() == doughnut.issuer() {
```
jordy25519: Can we make this a `debug_assert!` and with the correct tests should be ok 💭

runtime/tests/tests.rs
```diff
@@ -1083,3 +1083,344 @@ fn contract_call_fails_when_cpay_is_used_for_fee_exchange() {
...
+			);
+			assert_eq!(
+				<GenericAsset as MultiCurrency>::free_balance(&charlie(), Some(CENTRAPAY_ASSET_ID)),
+				balance_amount
+			);
+		});
+}
+
+#[test]
+fn generic_asset_transfer_works_with_doughnut_and_fee_exchange_combo() {
```
jordy25519, line 1206: 🏆

runtime/src/lib.rs
```diff
@@ -549,26 +573,60 @@ impl additional_traits::DelegatedDispatchVerifier for Runtime {
...
+
+	/// This is used when an issuer delegates permissions to a smart contract
+	/// It should verify that a contract being called by the smart contract is
+	/// permissioned. Not implemented yet.
+	fn verify_contract_to_contract_call(
+		_caller: &Self::AccountId,
+		doughnut: &Self::Doughnut,
+		_contract_addr: &Self::AccountId,
+	) -> Result<(), &'static str> {
+		let _: CENNZnut = Self::get_cennznut(doughnut)?;
```
jordy25519: could just be `Ok(())`
My only concern is we track that this needs some future attention via TODO, GH/jira issue etc.

jordy25519 (commented):
Some architecture things to consider
The new tests are 🔥!

runtime/src/lib.rs
hoani: Nice! Have moved `DelegatedDispatchVerfier` into `CENNZnetDispatchVerifier` under impls.rs!

runtime/src/lib.rs
hoani: Done

runtime/src/lib.rs
hoani: Yes we can (and have!)

runtime/src/lib.rs
hoani: Agreed

runtime/src/impls.rs
```diff
@@ -282,6 +286,89 @@ impl IsGasMeteredCall for GasMeteredCallResolver {
 	}
 }
 
+/// Provides a cennznet version of doughnut dispatch verification
+pub struct CENNZnetDispatchVerifier;
+
+/// Helpers which are used by the DelegatedDispatchVerifier
+impl CENNZnetDispatchVerifier {
+	/// Checks if the doughnut holds a `cennznut` and if so, it returns the decoded `cennznut`
+	fn get_cennznut(doughnut: &<Runtime as frame_system::Trait>::Doughnut) -> Result<CENNZnut, &'static str> {
```
jordy25519, line 295: These could be stand alone helper methods, not really required on `CENNZnetDispatchVerifier`, ideally would be implemented on the doughnut but seems not worth the extra effort

runtime/tests/tests.rs
```diff
@@ -1083,3 +1083,344 @@ fn contract_call_fails_when_cpay_is_used_for_fee_exchange() {
 			);
 		});
 }
+
+#[test]
+fn generic_asset_transfer_works_with_doughnut() {
+	let cennznut = doughnut::make_runtime_cennznut("generic-asset", "transfer");
```
jordy25519, line 1089: I'm starting to miss builder pattern here for readability, but for another PR definitely

jordy25519 (approved):
lgtm! 🍩 🚀

PR #191, opened 2020-04-05
CENNZxSpot Liquidity Check RPC

hoani:
This PR allows:
* a user to check the value of any account's liquidity using an RPC call `cennzx_liquidityValue` with arguments `[account_id, asset_id]`
* a user to check the price of an exchanges liquidity using an RPC call `cennzx_liquidityPrice` with arguments `[asset_id, liquidity_to_buy]`

## Changes

- Add a public function `account_liquidity_value` which gives the value of an account's liquidity for a particular `asset_id`
- Refactor calculation of liquidity value so that both `account_liquidity_value` and `remove_liquidity` use the same calculation
- Add a public function `liquidity_price` which gives the price of liquidity in terms of `core` and `asset`
- More thorough testing around liquidity value calculations
- RPC interface for `cennzx_liquidityValue` and `cennznet_liquidityPrice`

## Concerns:

- The wording around `Liquidity` storage and types in `cennzx-spot` is getting a little difficult to distinguish between what is being communicated

## Cool things:

A basic test after running a node and adding liquidity using the UI resulted in:

**liqudityValue**
```
cennznet % curl -H "Content-Type: application/json" --data '{"id": 1,"jsonrpc": "2.0","method": "cennzx_liquidityValue","params": ["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY", 16000]}' localhost:9933
{"jsonrpc":"2.0","result":[7357000000000000,7357000000000000,1337000000000000],"id":1}
```

**liqudityPrice**
```
cennznet % curl -H "Content-Type: application/json" --data '{"id": 1,"jsonrpc": "2.0","method": "cennzx_liquidityPrice","params": [16000, 1000]}' localhost:9933
{"jsonrpc":"2.0","result":[1000,182],"id":1}
```

The liquidityValue rpc call is listed under rpc.methods():

```
...
    "babe_epochAuthorship",
    "cennzx_buyPrice",
    "cennzx_liquidityPrice",
    "cennzx_liquidityValue",
    "cennzx_sellPrice",
    "chain_getBlock",
...
```

### DIFF
--- a/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
+++ b/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
@@ -55,7 +55,7 @@ sp_api::decl_runtime_apis! {
 			account: AccountId,
 			asset_id: AssetId,
 		) -> (Balance, Balance, Balance);
-		/// Query the price of liquidity for a particular asset_id
+		/// Query the price of liquidity for a particular `asset_id`
 		fn liquidity_price(
 			asset_id: AssetId,
 			liquidity_to_buy: Balance,
--- a/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
+++ b/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
@@ -55,5 +55,10 @@ sp_api::decl_runtime_apis! {
 			account: AccountId,
 			asset_id: AssetId,
 		) -> (Balance, Balance, Balance);
+		/// Query the price of liquidity for a particular asset_id
+		fn liquidity_price(
+			asset_id: AssetId,
+			liquidity_to_buy: Balance,
+		) -> (Balance, Balance);
 	}
 }
--- a/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
+++ b/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
@@ -33,9 +33,10 @@ pub enum CennzxSpotResult<Balance> {
 
 sp_api::decl_runtime_apis! {
 	/// The RPC API to interact with CENNZX Spot Exchange
-	pub trait CennzxSpotApi<AssetId, Balance> where
+	pub trait CennzxSpotApi<AssetId, Balance, AccountId> where
 		AssetId: Codec,
 		Balance: Codec + BaseArithmetic,
+		AccountId: Codec,
 	{
 		/// Query how much `asset_to_buy` will be given in exchange for `amount` of `asset_to_sell`
 		fn buy_price(
@@ -49,5 +50,10 @@ sp_api::decl_runtime_apis! {
 			amount: Balance,
 			asset_to_buy: AssetId,
 		) -> CennzxSpotResult<Balance>;
+		/// Query the value of liquidity in the exchange for `asset_id` for `account`
+		fn liquidity_value(
+			account: AccountId,
+			asset_id: AssetId,
+		) -> (Balance, Balance, Balance);
 	}
 }
--- a/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
+++ b/crml/cennzx-spot/rpc/runtime-api/src/lib.rs
@@ -51,11 +51,13 @@ sp_api::decl_runtime_apis! {
 			asset_to_buy: AssetId,
 		) -> CennzxSpotResult<Balance>;
 		/// Query the value of liquidity in the exchange for `asset_id` for `account`
+		/// Returns (liquidity_volume, core_value, asset_value)
 		fn liquidity_value(
 			account: AccountId,
 			asset_id: AssetId,
 		) -> (Balance, Balance, Balance);
 		/// Query the price of liquidity for a particular `asset_id`
+		/// Returns price as a combination of (core, asset)
 		fn liquidity_price(
 			asset_id: AssetId,
 			liquidity_to_buy: Balance,
--- a/crml/cennzx-spot/rpc/src/lib.rs
+++ b/crml/cennzx-spot/rpc/src/lib.rs
@@ -46,6 +46,10 @@ pub trait CennzxSpotApi<AssetId, Balance, AccountId> {
 	#[rpc(name = "cennzx_liquidityValue")]
 	// TODO: change to Result<Balance> once https://github.com/serde-rs/serde/pull/1679 is merged
 	fn liquidity_value(&self, account_id: AccountId, asset_id: AssetId) -> Result<(u64, u64, u64)>;
+
+	#[rpc(name = "cennzx_liquidityPrice")]
+	// TODO: change to Result<Balance> once https://github.com/serde-rs/serde/pull/1679 is merged
+	fn liquidity_price(&self, asset_id: AssetId, liquidity_to_buy: Balance) -> Result<(u64, u64)>;
 }
 
 /// An implementation of CENNZX Spot Exchange specific RPC methods.
@@ -174,5 +178,29 @@ where
 			data: Some(format!("{:?}", e).into()),
 		});
 		Ok((liquidity.unwrap(), core.unwrap(), asset.unwrap()))
+	}
+
+	fn liquidity_price(&self, asset_id: AssetId, liquidity_to_buy: Balance) -> Result<(u64, u64)> {
+		let api = self.client.runtime_api();
+		let best = self.client.info().best_hash;
+		let at = BlockId::hash(best);
+
+		let result = api.liquidity_price(&at, asset_id, liquidity_to_buy).map_err(|e| RpcError {
+			code: ErrorCode::ServerError(Error::Runtime.into()),
+			message: "Unable to query liquidity price.".into(),
+			data: Some(format!("{:?}", e).into()),
+		})?;
+
+		let core = TryInto::<u64>::try_into(result.0.saturated_into::<u128>()).map_err(|e| RpcError {
+			code: ErrorCode::ServerError(Error::PriceOverflow.into()),
+			message: "Core asset too large.".into(),
+			data: Some(format!("{:?}", e).into()),
+		});
+		let asset = TryInto::<u64>::try_into(result.1.saturated_into::<u128>()).map_err(|e| RpcError {
+			code: ErrorCode::ServerError(Error::PriceOverflow.into()),
+			message: "Trade asset too large.".into(),
+			data: Some(format!("{:?}", e).into()),
+		});
+		Ok((core.unwrap(), asset.unwrap()))
 	}
 }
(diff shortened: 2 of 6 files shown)


crml/cennzx-spot/src/lib.rs
```diff
@@ -44,6 +44,17 @@ pub type ExchangeKey<T> = (
 	<T as pallet_generic_asset::Trait>::AssetId,
 );
 
+pub struct LiquidityValue<Balance> {
```
jordy25519, line 53: Doc comments

crml/cennzx-spot/src/lib.rs
```diff
@@ -344,6 +356,96 @@ impl<T: Trait> Module<T> {
 		<TotalLiquidity<T>>::mutate(exchange_key, |balance| *balance = balance.saturating_sub(decrease));
 	}
 
+	/// The Price of Liquidity for a particular `asset` exchange
+	///
+	/// The price includes
+	///   * a required amount of core asset
+	///   * a required amount of `asset`
+	///
+	/// Note: if the exchange does not exist, the cost in `asset` is 1, because the invester
```
jordy25519: invest**o**r

crml/cennzx-spot/src/lib.rs
```diff
@@ -344,6 +356,96 @@ impl<T: Trait> Module<T> {
...
+	/// Calculate Liquidity Value
+	///
+	/// Simple helper function to calculate the value of liquidity
+	fn calculate_liquidity_value(
+		asset_reserve: T::Balance,
+		core_reserve: T::Balance,
+		liquidity_to_withdraw: T::Balance,
+		total_liquidity: T::Balance,
+	) -> LiquidityValue<T::Balance> {
+		if total_liquidity == Zero::zero() {
```
jordy25519: `total_liquidity.is_zero()`

Optional: impl Zero for liquidity value
`LiquidityValue::zero()`

crml/cennzx-spot/rpc/runtime-api/src/lib.rs
```diff
@@ -49,5 +50,15 @@ sp_api::decl_runtime_apis! {
 			amount: Balance,
 			asset_to_buy: AssetId,
 		) -> CennzxSpotResult<Balance>;
+		/// Query the value of liquidity in the exchange for `asset_id` for `account`
+		fn liquidity_value(
+			account: AccountId,
+			asset_id: AssetId,
+		) -> (Balance, Balance, Balance);
+		/// Query the price of liquidity for a particular asset_id
```
jordy25519: '`asset_id`'

crml/cennzx-spot/rpc/src/lib.rs
```diff
@@ -141,4 +150,59 @@ where
...
+		let best = self.client.info().best_hash;
+		let at = BlockId::hash(best);
+
+		let result = api.liquidity_value(&at, account, asset_id).map_err(|e| RpcError {
+			code: ErrorCode::ServerError(Error::Runtime.into()),
+			message: "Unable to query liquidity value.".into(),
+			data: Some(format!("{:?}", e).into()),
+		})?;
+
+		let liquidity = TryInto::<u64>::try_into(result.0.saturated_into::<u128>()).map_err(|e| RpcError {
```
jordy25519, line 165: wondering if `.saturated_into::<u128>` is required for these conversions? since it will turn into u64 anyway

crml/cennzx-spot/src/lib.rs
```diff
@@ -257,30 +269,26 @@ decl_module! {
 				Error::<T>::LiquidityTooLow
 			);
 
+			let user_balances = Self::get_liquidity_value(asset_id, liquidity_to_withdraw);
```
jordy25519: maybe `user_liquidity`

crml/cennzx-spot/src/lib.rs
```diff
@@ -344,6 +356,96 @@ impl<T: Trait> Module<T> {
 		<TotalLiquidity<T>>::mutate(exchange_key, |balance| *balance = balance.saturating_sub(decrease));
 	}
 
+	/// The Price of Liquidity for a particular `asset` exchange
```
jordy25519: The price of liquidity for \`asset_id\` exchange

crml/cennzx-spot/src/lib.rs
```diff
@@ -344,6 +356,96 @@ impl<T: Trait> Module<T> {
...
+	///
+	/// The price includes
+	///   * a required amount of core asset
+	///   * a required amount of `asset`
+	///
+	/// Note: if the exchange does not exist, the cost in `asset` is 1, because the invester
+	///       determines the exchange rate
+	pub fn liquidity_price(asset_id: T::AssetId, liquidity_to_buy: T::Balance) -> LiquidityPrice<T::Balance> {
```
jordy25519, line 378: Would be good to see this used in the `add_liquidity` and `remove_liquidity`, could be follow up PR

crml/cennzx-spot/rpc/src/lib.rs
```diff
@@ -141,4 +150,59 @@ where
...
+			code: ErrorCode::ServerError(Error::PriceOverflow.into()),
+			message: "Liquidity too large.".into(),
+			data: Some(format!("{:?}", e).into()),
+		});
```
jordy25519: consistency: can `?` the `map_err`s

jordy25519 (commented):
Really nice.
I think it would be worthwhile changing `add_liquidity` and `remove_liquidity` to use the new functions added here as it would make review easier and hopefully remove the need for having two similar functions `liquidity_value` and `get_liquidity_value`

crml/cennzx-spot/src/lib.rs
jordy25519: updated in review comment

crml/cennzx-spot/rpc/src/lib.rs
hoani: I think we tried that - but I don't think there was a direct conversion available from u64 to Balance - will give it a go

crml/cennzx-spot/src/lib.rs
hoani: it's a bit confusing, `liquidity_value` is actually the value of the user's liquidity (not just the liquidity balance itself).
Happy to take suggestions on clearer naming here

crml/cennzx-spot/rpc/src/lib.rs
hoani: Nice catch - will do

crml/cennzx-spot/rpc/src/lib.rs
hoani: The reason this doesn't work is that we get:
* "`<Balance as std::convert::TryInto<u64>>::Error` cannot be formatted using `{:?}`"

It is fine when we are doing `<u128 as std::convert::TryInto<u64>>::Error`
Haven't found a better way to do this yet - open to ideas

crml/cennzx-spot/src/lib.rs
hoani: Nice idea. Tried implementing `Zero` for `LiquidityValue`, but it requires that I implement `Add` for liquidity value - doesn't seem worth it yet

crml/cennzx-spot/rpc/runtime-api/src/lib.rs
```diff
@@ -49,5 +50,15 @@ sp_api::decl_runtime_apis! {
 			amount: Balance,
 			asset_to_buy: AssetId,
 		) -> CennzxSpotResult<Balance>;
+		/// Query the value of liquidity in the exchange for `asset_id` for `account`
+		fn liquidity_value(
+			account: AccountId,
+			asset_id: AssetId,
+		) -> (Balance, Balance, Balance);
```
aliXsed, line 58: If the return type was a struct, it could be clearer what this function returns. However, if you prefer a tuple, I would suggest you add a documenting comment for the return values.

crml/cennzx-spot/rpc/runtime-api/src/lib.rs
```diff
@@ -49,5 +50,15 @@ sp_api::decl_runtime_apis! {
...
+		/// Query the price of liquidity for a particular `asset_id`
+		fn liquidity_price(
+			asset_id: AssetId,
+			liquidity_to_buy: Balance,
+		) -> (Balance, Balance);
```
aliXsed, line 64: The same as my previous comment, IMO, the return values should be documented.

crml/cennzx-spot/rpc/src/lib.rs
```diff
@@ -42,6 +42,14 @@ pub trait CennzxSpotApi<AssetId, Balance> {
 	#[rpc(name = "cennzx_sellPrice")]
 	// TODO: change to Result<Balance> once https://github.com/serde-rs/serde/pull/1679 is merged
 	fn sell_price(&self, asset_to_sell: AssetId, amount_to_buy: Balance, asset_to_payout: AssetId) -> Result<u64>;
+
+	#[rpc(name = "cennzx_liquidityValue")]
+	// TODO: change to Result<Balance> once https://github.com/serde-rs/serde/pull/1679 is merged
```
aliXsed, line 47: Do we have a jira ticket for TODOs?

crml/cennzx-spot/src/lib.rs
```diff
@@ -44,6 +44,28 @@ pub type ExchangeKey<T> = (
...
+/// `liquidity` represents the volume of liquidity holdings being valued
+/// `core` represents the balance of `core_asset` that the liquidity would yield
+/// `asset` represents the balance of `trade_asset` that the liquidity
+pub struct LiquidityValue<Balance> {
+	pub liquidity: Balance,
+	pub core: Balance,
+	pub asset: Balance,
+}
+
+/// Represents the price to buy liqudiity from an exchange
```
aliXsed: liquidity is misspelled

aliXsed:
A very well documented, tested change. Thank you @hoani. It all LGTM. I have left some comments that would only mean some minor finishing touches if you agree with them.

crml/cennzx-spot/rpc/src/lib.rs
hoani: ^ Eric put something together to address this a while ago - I have captured it in ticket =)

crml/cennzx-spot/rpc/runtime-api/src/lib.rs
hoani: Nice! I have added some comments 👍

aliXsed (approved):

cowboy-bebug (approved):
LGTM

jordy25519 (approved):

PR #195, opened 2020-04-09
Chore/172 test validator set

aliXsed:
Proves and fixes the issue that was originally discovered and described by Parity in  https://github.com/paritytech/substrate/issues/4121

Staking rewards from the last era are assigned to staking::CurrentElected, however, at the point of awarding staking rewards, the staking::CurrentElectedhas been updated to the next era’s validator set.

**NOTE:** I would like to ask the reviewers to specially check the commit https://github.com/cennznet/cennznet/pull/195/commits/1e36e33533c21e63316b18e7accd3b98ebd9be38 which was necessary to make the test `contract_dispatches_runtime_call_funds_are_safu` work. This test is written in a way that I cannot understand what is being tested and why we have had those expectations. 
As all the other tests run smoothly I tend to believe the problem above was related to the fragility of the test itself.


runtime/tests/tests.rs
```diff
@@ -164,6 +219,7 @@ fn reward_validators(validators: &[(AccountId, AccountId)]) {
 #[test]
 fn start_session_works() {
 	ExtBuilder::default().build().execute_with(|| {
+		// TODO test the current session index after each set
```
mzxyz: Remove this TODO, at the end of `start_session` function already do this

runtime/tests/tests.rs
```diff
@@ -462,6 +518,125 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
...
+			rotate_to_session(last_session_of_era_0);
+
+			// The last session falls in the era 0
+			assert_eq!(Staking::current_era(), 0);
+
+			// make sure we have the correct number of validators elected
+			assert_eq!(Staking::current_elected().len(), validators_count);
+
+			// Make a block header whose author is specified as below
+			let author_index = 0; // index 0 of validators
```
mzxyz, line 602: [suggestion] Extract this part of code in a func like `specify_header()` as it both use in this and above tests and have not dependency with other vars

mzxyz (approved):
Nice work, leave some small suggestions, otherwise LGTM

runtime/tests/tests.rs
aliXsed: Yeah, I see. I would have written it differently myself. In any test, the expectations should be explicit. But not a big issue here. I will remove the comment.

runtime/tests/tests.rs
aliXsed: @HaoZhaiX I have had a second look at these few lines and realized they could be replaced with just one line like
`let header = specify_author(header_for_block_number(System::block_number() + 1), author_index.clone());`
So defining a new function for a one-liner is not really necessary. On the other side, I believe I have broken that one line into a few separate steps just for the sake of these two tests, so I can clarify how the header is built and why is it built this way.  So I prefer to keep them as they are. What do you think?

runtime/tests/tests.rs
aliXsed: Done

runtime/tests/mock.rs
```diff
@@ -140,11 +140,17 @@ impl ExtBuilder {
 		.assimilate_storage(&mut t)
 		.unwrap();
 
+		let min_validator_count = if initial_authorities.len() > 3 {
+			3u32
+		} else {
+			initial_authorities.len() as u32
+		};
+
```
jordy25519:
```suggestion
		let min_validator_count = cmp::min(initial_authorities.len(), 3);
```

runtime/tests/tests.rs
```diff
@@ -128,6 +151,37 @@ fn transfer_fee<E: Encode>(extrinsic: &E, runtime_call: &Call) -> Balance {
...
+			network_state: Default::default(),
+			session_index: Session::current_index(),
+			authority_index: i as u32,
+		};
+		let call = pallet_im_online::Call::heartbeat(heartbeat_data, Default::default());
+		ImOnline::dispatch(call, RawOrigin::None.into()).unwrap();
+	}
+}
+
+fn prepare_for_rotating_session() {
```
jordy25519: `pre_rotate_session` + some doc comment

runtime/tests/tests.rs
```diff
@@ -466,6 +521,125 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
 		});
 }
 
+#[test]
+/// This tests if authorship reward of the last block in an era is factored in.
+fn authorship_reward_of_last_block_in_an_era() {
+	let validators_count = 6;
```
jordy25519: `validator_count`

runtime/tests/tests.rs
```diff
@@ -466,6 +521,125 @@ fn staking_validators_should_receive_equal_transaction_fee_reward() {
...
+	let initial_balance = 1_000 * TransactionBaseFee::get();
+
+	ExtBuilder::default()
+		.validator_count(validators_count)
+		.initial_balance(initial_balance)
+		.stash(initial_balance)
+		.build()
+		.execute_with(|| {
+			let last_session_of_era_0 = SessionsPerEra::get() - 1;
```
jordy25519: maybe `final_session_of_era_index`

runtime/tests/tests.rs
```diff
@@ -50,16 +52,37 @@ use sp_core::crypto::Pair;
...
 	)
 }
 
+fn header() -> Header {
+	header_for_block_number(1)
+}
+
+fn specify_author(mut header: Header, author_index: AuthorityIndex) -> Header {
+	use digests::RawPreDigest;
+	use digests::SecondaryPreDigest;
```
jordy25519:
```suggestion
	use digests::{RawPreDigest, SecondaryPreDigest};
```

runtime/tests/tests.rs
```diff
@@ -50,16 +52,37 @@ use sp_core::crypto::Pair;
...
 		GENESIS_HASH.into(),      // parent_hash
 		Digest::default(),        // digest
 	)
 }
 
+fn header() -> Header {
+	header_for_block_number(1)
+}
+
+fn specify_author(mut header: Header, author_index: AuthorityIndex) -> Header {
```
jordy25519: maybe `set_author` + doc comment

jordy25519 (approved):
lgtm!
This seems to be a different, yet related issue to what is described in paritytech/substrate#4121

runtime/tests/mock.rs
aliXsed: Done

runtime/tests/tests.rs
aliXsed: Done

runtime/tests/tests.rs
aliXsed: Done

runtime/tests/tests.rs
aliXsed: Done

runtime/tests/tests.rs
aliXsed: Done

runtime/tests/tests.rs
aliXsed: Done

PR #197, opened 2020-04-11
Transaction payment module tidy up

hoani:
This PR addresses some refactor opportunities identified in a recent audit.

## Addresses:

- `fee_asset_spent` should be renamed to reflect that it is an exchange currency
- Too much nesting in calculation for imbalance
- Add test to check that imbalances are handled in validate

## Changes:

- `fee_asset_spent` renamed to `exchange_asset_spent`
- Unravelled imbalance calculation
- Add test to check that imbalances are handled in validate

## Concerns:

- The `()` implementation of `NegativeImbalance::on_imbalanceds()` calls the `NegativeImbalance::drop()`, so the test doesn't really prove that imbalances are handled in validate, since ignoring the imbalance and handling the imbalance has the same effect.


cowboy-bebug (approved):
LGTM!

mzxyz (approved):
LGTM

crml/transaction-payment/src/lib.rs
```diff
@@ -917,30 +916,26 @@ mod tests {
...
+				let len: u64 = 0;
+				let total_issuance = Balances::total_issuance();
+				let dispatch_info = info_from_weight(1);
+				let fee = ChargeTransactionPayment::<Runtime>::compute_fee(0, dispatch_info.clone(), 0);
+
+				let transaction_payment = ChargeTransactionPayment::<Runtime>::from(0, None);
+				assert!(transaction_payment
+					.pre_dispatch(&1, CALL, dispatch_info, len as usize)
 					.is_ok());
-				assert!(storage::unhashed::get::<Option<()>>(&GAS_FEE_EXCHANGE_KEY).is_none());
```
jordy25519: hmm, can't see why this test was removed?

hoani: Yeah, that doesn't seem right, nice catch

hoani: Added it back, thanks @holygits

jordy25519 (approved):

PR #199, opened 2020-04-13
Sylo data migration

cowboy-bebug:
The PR is to prepare for the upcoming Sylo data migration.

### Changes

Add a separate `migration.rs` runtime module that implements the following:
- Add `MigrationAccount` storage item, which will store only one migrator account
- Add 3 dispatch calls:
    - `fn set_migrator_account`, which requires root permission
    - `fn remove_migrator_account`, which is to kill the storage after migration is finished
    - `fn migrate_devices`, which will put new data (from arg) into Devices storage
- Add private `fn ensure_sylo_migrator` module method to only allow migration account to call migration methods
- Add `FixedOperational(0)` for the weights
- Add tests to show the above is working

### DIFF
--- a/crml/sylo/src/e2ee.rs
+++ b/crml/sylo/src/e2ee.rs
@@ -133,7 +133,7 @@ pub(super) mod tests {
 
 	#[test]
 	fn should_add_device() {
-		ExtBuilder.build().execute_with(|| {
+		ExtBuilder::default().build().execute_with(|| {
 			assert_ok!(E2EE::register_device(
 				Origin::signed(H256::from_low_u64_be(1)),
 				0,
@@ -153,7 +153,7 @@ pub(super) mod tests {
 
 	#[test]
 	fn should_replenish_pkbs() {
-		ExtBuilder.build().execute_with(|| {
+		ExtBuilder::default().build().execute_with(|| {
 			assert_ok!(E2EE::register_device(
 				Origin::signed(H256::from_low_u64_be(1)),
 				0,
@@ -174,7 +174,7 @@ pub(super) mod tests {
 
 	#[test]
 	fn should_withdraw_pkbs() {
-		ExtBuilder.build().execute_with(|| {
+		ExtBuilder::default().build().execute_with(|| {
 			assert_ok!(E2EE::register_device(
 				Origin::signed(H256::from_low_u64_be(1)),
 				0,
--- a/crml/sylo/src/e2ee.rs
+++ b/crml/sylo/src/e2ee.rs
@@ -13,7 +13,10 @@
 *     https://centrality.ai/licenses/lgplv3.txt
 */
 
-use crate::{device, groups, inbox, response};
+use crate::{
+	device::{self, DeviceId},
+	groups, inbox, response,
+};
 use frame_support::{
 	decl_error, decl_module, decl_storage,
 	dispatch::Vec,
@@ -25,8 +28,6 @@ use frame_system::ensure_signed;
 const MAX_PKBS: usize = 50;
 
 pub trait Trait: inbox::Trait + response::Trait + device::Trait + groups::Trait {}
-
-type DeviceId = device::DeviceId;
 
 // Serialized pre key bundle used to establish one to one e2ee
 pub type PreKeyBundle = Vec<u8>;
--- a/crml/sylo/src/e2ee.rs
+++ b/crml/sylo/src/e2ee.rs
@@ -26,7 +26,7 @@ const MAX_PKBS: usize = 50;
 
 pub trait Trait: inbox::Trait + response::Trait + device::Trait + groups::Trait {}
 
-type DeviceId = u32;
+type DeviceId = device::DeviceId;
 
 // Serialized pre key bundle used to establish one to one e2ee
 pub type PreKeyBundle = Vec<u8>;
--- a/crml/sylo/src/migration.rs
+++ b/crml/sylo/src/migration.rs
@@ -20,11 +20,9 @@ use sp_runtime::{DispatchError::BadOrigin, DispatchResult};
 
 pub trait Trait: device::Trait + groups::Trait + inbox::Trait + vault::Trait {}
 
-type DeviceId = device::DeviceId;
-
 decl_error! {
 	pub enum Error for Module<T: Trait> {
-		InputError,
+		MaxDeviceLimitReached,
 	}
 }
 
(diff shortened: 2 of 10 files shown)


crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,53 @@
...
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin, system = frame_system {
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		pub fn set_migrator_account(origin, account_id: T::AccountId) -> DispatchResult {
+			ensure_root(origin)?;
+			MigrationAccount::<T>::put(account_id);
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		pub fn remove_migrator_account(origin) -> DispatchResult {
```
jordy25519: I recommend making this a 'self destruct' instead of root

jordy25519 (commented):
🥇

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,53 @@
...
+*     https://centrality.ai/licenses/gplv3.txt
+*     https://centrality.ai/licenses/lgplv3.txt
+*/
+
+use crate::{device, e2ee, groups, inbox, response, vault};
+use frame_support::{decl_module, decl_storage, ensure, weights::SimpleDispatchInfo};
+use frame_system::{ensure_root, ensure_signed};
+use sp_runtime::{DispatchError::BadOrigin, DispatchResult};
+
+pub trait Trait: device::Trait + e2ee::Trait + groups::Trait + inbox::Trait + response::Trait + vault::Trait {}
```
cowboy-bebug: NOTE: `migration::Trait` is a super trait of all Sylo modules with overlaps (written this way for readability).
Eg, `e2ee::Trait` is defined as:
```rust
pub trait Trait: inbox::Trait + response::Trait + device::Trait + groups::Trait {}
```

crml/sylo/src/migration.rs
cowboy-bebug: Ah yep that's right I'll update it with `Self::ensure_sylo_migrator(origin)?;`

hoani (dismissed):

jordy25519 (approved):
lgtm, would recommend adding a test to check it works in other modules before merging

crml/sylo/src/e2ee.rs
```diff
@@ -26,7 +26,7 @@ const MAX_PKBS: usize = 50;
 
 pub trait Trait: inbox::Trait + response::Trait + device::Trait + groups::Trait {}
 
-type DeviceId = u32;
+type DeviceId = device::DeviceId;
```
jordy25519: probably don't need to re-alias this: `use device::DeviceId;`

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,226 @@
...
+*/
+
+use crate::{device, groups, inbox, vault};
+use frame_support::{decl_error, decl_module, decl_storage, dispatch::Vec, ensure, weights::SimpleDispatchInfo};
+use frame_system::{ensure_root, ensure_signed};
+use sp_runtime::{DispatchError::BadOrigin, DispatchResult};
+
+pub trait Trait: device::Trait + groups::Trait + inbox::Trait + vault::Trait {}
+
+type DeviceId = device::DeviceId;
```
jordy25519: same here

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,226 @@
...
+		pub fn remove_migrator_account(origin) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+			MigrationAccount::<T>::kill();
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_devices(origin, user_id: T::AccountId, device_ids: Vec<DeviceId>) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+			ensure!(device_ids.len() <= device::MAX_DEVICES, Error::<T>::InputError);
```
jordy25519: recommend we use custom error message or it will be difficult to debug if we hit these

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,226 @@
...
+
+			let mut devices = <device::Devices<T>>::get(user_id.clone());
+			ensure!(devices.len() + device_ids.len() <= device::MAX_DEVICES, Error::<T>::InputError);
+
+			for device_id in device_ids {
+				devices.push(device_id)
```
jordy25519: I think there's a method for this smth like: `devices.extend(&device_ids.iter())`

jordy25519 (commented):
recommend renaming `remove_migrator_account` -> `self_destruct` for style points :p

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,226 @@
...
+			}
+
+			devices.sort();
```
jordy25519: Worth checking with Sylo whether sorting here would affect the client, does seem unlikely...

crml/sylo/src/migration.rs
cowboy-bebug: Yep will be renamed to `MaxDeviceLimitReached`

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,221 @@
...
+		pub fn self_destruct(origin) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+			MigrationAccount::<T>::kill();
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_devices(origin, user_id: T::AccountId, device_ids: Vec<device::DeviceId>) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+			ensure!(device_ids.len() <= device::MAX_DEVICES, Error::<T>::MaxDeviceLimitReached);
```
jordy25519, line 59: this is being checked twice

jordy25519 (approved):

crml/sylo/src/migration.rs
cowboy-bebug: I think it's worth preserving the order. Would u accept this as an alternative:
```rust
for device_id in device_ids {
    if !devices.contains(&device_id) {
        devices.push(device_id);
    }
}
```

crml/sylo/src/migration.rs
cowboy-bebug: The first check is for the `device_ids` arg and the second is for what's in the storage (which should be 0 but just to be sure)

crml/sylo/src/migration.rs
```diff
@@ -0,0 +1,223 @@
...
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		pub fn self_destruct(origin) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+			MigrationAccount::<T>::kill();
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_devices(origin, user_id: T::AccountId, device_ids: Vec<device::DeviceId>) -> DispatchResult {
```
mzxyz: [suggestion] use `Vec<DeviceId>` directly in the parameter type. Otherwise on JS side, the type will display same with `device::DeviceId `.

mzxyz (approved):
Awesome work! Add a small suggestion, otherwise LGTM

crml/sylo/src/migration.rs
cowboy-bebug: Thanks for pointing it out. I'll update it!

PR #203, opened 2020-04-16
Support for Sylo data migration

cowboy-bebug:
### Changes
- Add `fn migrate_inbox`
- Add `fn migrate_vault`
- ~~Expose required types & storages~~
- Add unit tests

### Concerns / Notes
- method signatures could be changed later depending on the migration script and API
- `fn migrate_inbox` retrieves any existing messages and append new messages whose index is not present in the existing data
    - this is done to leave the existing messages alone, assuming that these are already (partially) migrated
    - otherwise we could end up with duplicates or two versions of data

### DIFF
--- a/crml/sylo/src/inbox.rs
+++ b/crml/sylo/src/inbox.rs
@@ -21,8 +21,8 @@ use frame_system::ensure_signed;
 const MAX_MESSAGE_LENGTH: usize = 100_000;
 const MAX_DELETE_MESSAGES: usize = 10_000;
 
-type MessageId = u32;
-type Message = Vec<u8>;
+pub type MessageId = u32;
+pub type Message = Vec<u8>;
 
 pub trait Trait: frame_system::Trait {}
 
@@ -69,8 +69,8 @@ decl_module! {
 
 decl_storage! {
 	trait Store for Module<T: Trait> as SyloInbox {
-		NextIndexes: map hasher(blake2_128_concat) T::AccountId => MessageId;
-		Values get(values): map hasher(blake2_128_concat) T::AccountId => Vec<(MessageId, Message)>;
+		pub NextIndexes: map hasher(blake2_128_concat) T::AccountId => MessageId;
+		pub Values get(values): map hasher(blake2_128_concat) T::AccountId => Vec<(MessageId, Message)>;
 	}
 }
 
--- a/crml/sylo/src/migration.rs
+++ b/crml/sylo/src/migration.rs
@@ -15,7 +15,9 @@
 
 use crate::{
 	device::{self, DeviceId},
-	groups, inbox, vault,
+	groups,
+	inbox::{self, Message, MessageId},
+	vault,
 };
 use frame_support::{decl_error, decl_module, decl_storage, dispatch::Vec, ensure, weights::SimpleDispatchInfo};
 use frame_system::{ensure_root, ensure_signed};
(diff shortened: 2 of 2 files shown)


crml/sylo/src/inbox.rs
```diff
@@ -69,8 +69,8 @@ decl_module! {
 
 decl_storage! {
 	trait Store for Module<T: Trait> as SyloInbox {
-		NextIndexes: map hasher(blake2_128_concat) T::AccountId => MessageId;
-		Values get(values): map hasher(blake2_128_concat) T::AccountId => Vec<(MessageId, Message)>;
+		pub NextIndexes: map hasher(blake2_128_concat) T::AccountId => MessageId;
```
aliXsed, line 72: @cowboy-bebug I highly recommend that we don't make the internal structure of modules public for the sake of the migration task. I'm going to make a PR that would show the alternative which is much easier to be removed later and doesn't harm the encapsulation of the modules.

cowboy-bebug: Nice! Based on the offline discussion, we decided to move dispatch methods to their own modules and keep the original scope

crml/sylo/src/migration.rs
```diff
@@ -71,6 +75,45 @@ decl_module! {
...
+			let (existing_indexes, _): (Vec<MessageId>, Vec<_>) = existing_messages.clone().into_iter().unzip();
+
+			// For repeatability, we update the existing messages that are assumed to be migrated already.
+			for (new_index, new_message) in new_messages {
+				if !existing_indexes.contains(&new_index) {
+					existing_messages.push((new_index, new_message));
+				}
+			}
+
+			ensure!(existing_messages.len() as u32 <= u32::max_value(), Error::<T>::MaxInboxLimitReached);
```
jordy25519: `existing_messages.len() as u32` will truncate to `u32`, so this `ensure!` won't ever fail afaict

crml/sylo/src/migration.rs
```diff
@@ -71,6 +75,45 @@ decl_module! {
 
 			Ok(())
 		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_inbox(origin, user_id: T::AccountId, next_index: MessageId, new_messages: Vec<(MessageId, Message)>) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+
+			let mut existing_messages = <inbox::Values<T>>::get(&user_id).clone();
```
jordy25519: is this clone needed?

crml/sylo/src/migration.rs
```diff
@@ -71,6 +75,45 @@ decl_module! {
...
+			<inbox::Values<T>>::insert(&user_id, existing_messages);
+			<inbox::NextIndexes<T>>::insert(&user_id, next_index);
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_vault(origin, user_id: T::AccountId, new_vaults: Vec<(VaultKey, VaultValue)>) -> DispatchResult {
+			Self::ensure_sylo_migrator(origin)?;
+
+			let mut existing_vaults = <vault::Vault<T>>::get(&user_id).clone();
```
jordy25519: is this clone needed?

jordy25519 (approved):
lgtm! some small tidy ups suggested

crml/sylo/src/migration.rs
cowboy-bebug: That's true lol. I'll remove this check as I'm sure we don't have this many messages anyway

crml/sylo/src/migration.rs
cowboy-bebug: Opps I'll remove it

crml/sylo/src/migration.rs
cowboy-bebug: Opps I'll remove this one as well

PR #204, opened 2020-04-17
Implement migrate groups and move device migrate functions to the dev…

aliXsed:
@cowboy-bebug @holygits I think we only need two functions for migrating the group storage: migrate_groups and migrate_group_devices. The membership storage will be created during migrating groups and thus we wouldn't need the migrate_memberships anymore.


hoani:
I suggest keeping all of the migration extrinsics in the migration runtime module. 
This way the business logic and the migration logic are completely seperate.
This was the reason we originally moved `migrate_devices` into `migrate.rs` from `devices.rs`, I don't think moving it the other way is helpful.

aliXsed:
> I suggest keeping all of the migration extrinsics in the migration runtime module.
> This way the business logic and the migration logic are completely seperate.
> This was the reason we originally moved `migrate_devices` into `migrate.rs` from `devices.rs`, I don't think moving it the other way is helpful.

Putting all the migration logic in one file had made us opening up the encapsulation of all the other modules. This means making things public that weren't public otherwise. Later on, when we would want to remove the migration logic, tracing which part should remain public, which should be made private wasn't as easy as removing the migration functions from each file. 
On the other hand, the migration logic in the same file with the business logic doesn't mean they are mixed. If they had been mixed we couldn't keep the same unit tests there. 
Finally, any other solution was more expensive and not necessary for a temporary logic.

crml/sylo/src/groups/mod.rs
```diff
@@ -398,6 +401,32 @@ decl_module! {
 
 			Ok(())
 		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_groups(origin, group_id: T::Hash, group: Group<T::AccountId, T::Hash>) -> DispatchResult {
+			ensure_signed(origin.clone())?;
```
cowboy-bebug: We don't need to `ensure_signed` as ensure_sylo_migrator already checks it

crml/sylo/src/groups/mod.rs
```diff
@@ -398,6 +401,32 @@ decl_module! {
...
+			for member in group.members {
+				Self::store_membership(&member.user_id, group_id.clone());
+			}
+
+			Ok(())
+		}
+
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		fn migrate_group_devices(origin, group_id: T::Hash, member_devices: Vec<(T::AccountId, DeviceId)>) -> DispatchResult {
+			ensure_signed(origin.clone())?;
```
cowboy-bebug: This can be removed as well

cowboy-bebug (commented):
Looks good so far. We need to fix the build issue

crml/sylo/src/groups/mod.rs
aliXsed: Done

crml/sylo/src/groups/mod.rs
aliXsed: Done

jordy25519 (commented):
Add some unit tests for the migration calls
lgtm otherwise

crml/sylo/src/groups/tests.rs
```diff
@@ -326,4 +326,144 @@ mod tests {
...
+				Groups::memberships(H256::from_low_u64_be(2)),
+				vec![H256::from_low_u64_be(0)]
+			);
+			assert_ne!(
+				Groups::memberships(H256::from_low_u64_be(2)),
+				vec![H256::from_low_u64_be(1)]
+			);
+
+			// Test user id 3 is not a member of any groups
+			assert_ne!(Groups::memberships(H256::from_low_u64_be(3)), vec![]);
```
jordy25519, line 427: `assert_ne!` is this supposed to be `assert_eq!`?

crml/sylo/src/groups/tests.rs
```diff
@@ -326,4 +326,144 @@ mod tests {
...
+			// Retrieve the group from the storage by using its id and check everything has stored properly
+			assert_eq!(Groups::group(group.group_id.clone()), *group);
+		});
+	}
+
+	#[test]
+	fn migrate_groups_updates_memberships() {
+		ExtBuilder::default().build().execute_with(|| {
+			// Set up 2 groups
+			setup_migrate_groups(2);
```
jordy25519, line 408: nit: would have been clearer to do smth like:
```rust
let group1 = make_group(1, vec![1,2]);
let group2 = make_group(2, vec![1,3]);
```
as now have to do some mental arithmetic to think about what the right IDs are

jordy25519 (approved):
lgtm

cowboy-bebug (approved):

crml/sylo/src/groups/tests.rs
aliXsed: Good idea, I will do that as soon as we need to use this function again

crml/sylo/src/groups/tests.rs
aliXsed: It's actually a mistake but not failing as surprisingly [ ] is not equal [ ]. Nice catch. I will fix it in a new PR.

PR #212, opened 2020-04-24
Add a script to install rust versions and build

cowboy-bebug:
Something I noticed while updating the wiki is that users have to make sure they have correct rust and nightly versions installed.

The shell script will pull from `./.circleci/config.yml` and use the versions there. Unfortunately, there's not a solution to inject environment variables from a file into config.yml so let's use config.yml as the centralised source of truth for Rust versions for now.

tldr; users can run `./build.sh`


init.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+nightly_pattern='^.+(nightly-[0-9]{4}-[0-9]{2}-[0-9]{2}).*$'
+nightly_version=$(grep -E $nightly_pattern ./.circleci/config.yml | head -n 1)
```
aliXsed, line 3: I hope the problem that we recently had with one of the nightly versions was just limited to that version and in practice, we can just instruct the users the install the very latest version of the nightly. We should change our own config.yml too accordingly.

jordy25519 (approved):
not sure if you `chmod` it yet but lgtm

PR #216, opened 2020-05-11
Fix/flaky membership check

jordy25519:
ZSTs like `()` are not properly supported by trie storage. This was causing non-deterministic behaviour.

This PR makes the value a boolean which solves the issue.
Also:
- made the error different to `BadOrigin` to make debugging easier.
- Bump cargo versions to 1.0.1 so running node shows the correct version

see: https://github.com/paritytech/subport/issues/7

### DIFF
--- a/crml/sylo/src/migration.rs
+++ b/crml/sylo/src/migration.rs
@@ -17,7 +17,7 @@
 
 use frame_support::{decl_module, decl_storage, ensure, weights::SimpleDispatchInfo, IterableStorageMap};
 use frame_system::{ensure_root, ensure_signed};
-use sp_runtime::{DispatchError::BadOrigin, DispatchResult};
+use sp_runtime::DispatchResult;
 use sp_std::prelude::*;
 
 pub trait Trait: frame_system::Trait {}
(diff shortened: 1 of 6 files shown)


KarishmaBothara (approved):
lgtm

aliXsed (approved):

crml/sylo/src/migration.rs
```diff
@@ -55,7 +55,7 @@ impl<T: Trait> Module<T> {
 	// Ensure `origin` is an authorized Sylo data migrator
 	pub fn ensure_sylo_migrator(origin: T::Origin) -> DispatchResult {
 		let account_id = ensure_signed(origin)?;
-		ensure!(AuthorisedMigrators::<T>::contains_key(account_id), BadOrigin);
+		ensure!(AuthorisedMigrators::<T>::get(account_id), "NotASyloMigrator");
```
aliXsed, line 58: I suppose `get(...)` returns Some/None. So I wonder how we are ensuring the boolean evaluation of an option?

jordy25519: Good q, internally `::get` will call the following:
```
from_optional_value_to_query(v: Option<bool>) -> Self::Query {
    v.unwrap_or_else(|| Default::default())
}
```
So it returns the type's default rather than `None`.

 You can check this by running `cargo expand`

PR #218, opened 2020-05-17
Fix Sylo Vault migration

### LINKED ISSUE #217: Updating previously migrated Sylo vault entries may fail
During the second phase of migrating data between two different chains, we need to update the entries which have changed since the first phase. However, the following check is preventing us from changing a Vault entry if the key for that entry already exists:
https://github.com/cennznet/cennznet/blob/43f355f7ac2866a329e16614a46349ae05e7fa93/crml/sylo/src/vault.rs#L85-L86

aliXsed:
Migration code should see the state of the source chain as the source of truth.
Closes #217


jordy25519 (approved):

jordy25519:
@alexsednz can you change target PR branch to: `1.0.1` and bump `spec_version` and `impl_version` in `runtime/src/lib.rs` by 1 each

KarishmaBothara (approved):

PR #225, opened 2020-05-25
Removed all code related to Sylo migration

KarishmaBothara:


aliXsed (approved):

benxgao:
I am not able to give much feedback in code level changes, but wondering if we intended to delete the migration code at the beginning, or did we just get the idea of removing those changes recently? @alexsednz  @holygits

jordy25519:
@benxgao yes always planned to remove it.
@KarishmaBothara can you bump the impl and spec version in runtime/src/lib.rs to `34` then it lgtm

PR #229, opened 2020-06-04
Scale CENNZ and CPAY down by 10^12 factor on a runtime upgrade

### LINKED ISSUE #228: Add a temporary feature to scale down CPAY and CENNZ balances
For the sake of better user experience, we are going to scale down all the CPAY and CENNZ free balances by 10e12 factor on a runtime upgrade.

aliXsed:
Uses the sudo key to burn the balances.

Closes #228.


runtime/src/lib.rs
```diff
@@ -815,6 +816,24 @@ impl_runtime_apis! {
...
+				let scaled_balance = balance.checked_div(ScaleDownFactor::get()).unwrap_or(balance);
+				let burn_amount = balance.checked_sub(scaled_balance).unwrap_or(Zero::zero());
+				let _ = GenericAsset::burn_free(&asset_id, &Sudo::key(), &who, &burn_amount);
+			});
+		};
+
+		scale_down(GenericAsset::spending_asset_id());
+		scale_down(GenericAsset::staking_asset_id());
+
+		0 // No weight
```
jordy25519: `Zero::zero()`

runtime/tests/tests.rs
```diff
@@ -13,21 +13,21 @@
...
 */
 
-use cennznet_primitives::types::{AccountId, Balance, FeeExchange, FeeExchangeV1};
+use cennznet_primitives::types::{AccountId, Balance, FeeExchange, FeeExchangeV1, AssetId};
 use cennznet_runtime::{
 	constants::{asset::*, currency::*},
 	Babe, Call, CennzxSpot, CheckedExtrinsic, ContractTransactionBaseFee, EpochDuration, Event, Executive,
 	GenericAsset, Header, Origin, Runtime, Session, SessionsPerEra, Staking, System, Timestamp, TransactionBaseFee,
-	TransactionByteFee, TransactionPayment, UncheckedExtrinsic,
+    TransactionByteFee, TransactionPayment, UncheckedExtrinsic, ScaleDownFactor,
```
jordy25519: Did you run `cargo fmt`?

jordy25519 (commented):
lgtm! will approve once CI passes and tested on a dev chain. Not sure how implementing `on_runtime_upgrade` on the `Runtime` struct will actually play out in practice vs. for a module

runtime/tests/tests.rs
aliXsed: Yes, and I'm confused. Is a tab 4 spaces or a tab?

jordy25519: a tab char

aliXsed: We need to add .rustfmt.toml to the project to specify our own formatting rather than the defaults. I'm afraid the defaults might change with different versions of rustfmt.

aliXsed: Okay we actually have this file but somehow not working for me

aliXsed: Will format the files again

aliXsed: Done. The problem was with CLION doing its own formatting.

runtime/src/lib.rs

aliXsed:
> module

@holygits good point. Let me investigate that.

runtime/tests/mock.rs
```diff
@@ -94,6 +98,10 @@ impl ExtBuilder {
 		self.validator_count = count;
 		self
 	}
+	pub fn sudoer(mut self, key: AccountId32) -> Self {
```
KarishmaBothara: Doubt here, do we need changes for scaling in tests/mock.rs too? As we now have separate module and its mock contains all these related function..

aliXsed: Yup, you are right. we don't need this anymore here. Thank you for seeing it.

KarishmaBothara (approved):

jordy25519 (approved):

PR #234, opened 2020-06-24
Chore/233 update cennznet

aliXsed:
The original branch contains a deep update of the staking module which is reverted but can be reviewed as a separate commit to see which direction we would need to take eventually. What is left are shallow but absolutely necessary changes. The project builds for all targets and the tests pass on my local machine.


KarishmaBothara (approved):

benxgao (approved):
Now that is is just for team internal using purpose, and not a real release candidate branch. It should be good to go.

PR #241, opened 2020-07-09
Regenerate nikau.*.json

jordy25519:
should be part of the 1.1.0-rc1 branch


aliXsed (approved):

aliXsed:
@holygits Would you add the same change to Azalea.json too?

jordy25519:
Turns out this isn't necessary, the original `nikau.raw.json` works fine

PR #245, opened 2020-07-15
Update/1.0.0 rc4.1

jordy25519:
Bring in the doughnut verification fix and additional tx pool fix
~would look to merge after #244~ done


Dockerfile
```diff
@@ -22,7 +22,10 @@ RUN apt-get update && \
     ln -s /root/.local/share/cennznet /data
 
 COPY --from=0 /cennznet/target/release/cennznet /usr/local/bin
+# copy in genesis files
```
jordy25519, line 25: Include latest wasm blob in docker images

aliXsed (approved):

PR #250, opened 2020-07-21
Chore/247 coverage report

### LINKED ISSUE #247: Add test coverage report for Sylo and cennzx-spot
Integrate grcov with our CircleCI process to produce test coverage reports for the mentioned packages.

aliXsed:
Closes #247 

Here is where the test coverage reports for this branch is kept: [artifacts](https://app.circleci.com/pipelines/github/cennznet/cennznet/1249/workflows/5db10b62-5151-41ec-913d-b0850aba208e/jobs/3128/artifacts). You can browse the test coverage report by opening [index.html](https://3128-183523539-gh.circle-artifacts.com/0/tmp/coverage/index.html) from the above link. 

Here is a screenshot of the report:
Screen Shot 2020-07-21 at 3 44 08 PM

The overhead of the test coverage report to the build process seems to be about 10%. Though we need to consider that at the moment we are only testing Sylo and cenzx-spot:
Screen Shot 2020-07-21 at 3 56 21 PM

On the good side, the more crates we add to the coverage test stage, the less we need to test separately in the build-test-and-run stage.

**Note**: As the Sylo and cennzx-spot will be tested for the sake of reporting the test coverage, testing them in the cargo-build-test-and-run step could be redundant. That's why I have excluded them there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,7 @@ commands:
           name: Report test coverage
           command: |
             cargo install grcov
+            rustup target add --toolchain=$RUST_NIGHTLY x86_64-unknown-linux-gnu
             export CARGO_INCREMENTAL=0
             export RUSTFLAGS="-Zprofile -Ccodegen-units=1 -Copt-level=0 -Clink-dead-code -Coverflow-checks=off -Zpanic_abort_tests -Cpanic=abort"
             export RUSTDOCFLAGS="-Cpanic=abort"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -82,13 +82,32 @@ commands:
...
       - run:
           name: Run tests
           command: |
-            cargo test --all
+            cargo test --workspace --exclude crml-cennzx-spot --exclude crml-sylo
+  cargo-test-coverage:
+    steps:
+      - run:
+          name: Report test coverage
+          command: |
```
benxgao, line 96: Not a big deal, but thinking if any pros/cons to put multiple commands into a shell script file like `scripts/test_coverage.sh`

aliXsed: Good idea, I will wait to see if there is any other reason to change this PR, and if so, I will do that too.

KarishmaBothara (approved):

.circleci/config.yml
aliXsed: Done

PR #251, opened 2020-07-21
Count chars instead of bytes

cowboy-bebug:
Noticed we are counting number of bytes rather than characters themselves (which is okay for ASCII values).

### DIFF
--- a/runtime/src/impls.rs
+++ b/runtime/src/impls.rs
@@ -348,7 +348,7 @@ impl additional_traits::DelegatedDispatchVerifier for CENNZnetDispatchVerifier {
 			.find('-')
 			.ok_or("CENNZnut does not grant permission for module")?
 			+ 1;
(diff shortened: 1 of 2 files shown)


runtime/src/impls.rs
```diff
@@ -348,7 +348,7 @@ impl additional_traits::DelegatedDispatchVerifier for CENNZnetDispatchVerifier {
 			.find('-')
 			.ok_or("CENNZnut does not grant permission for module")?
 			+ 1;
-		if module_offset <= 1 || module_offset >= module.len() {
+		if module_offset <= 1 || module_offset >= module.chars().count() {
```
jordy25519, line 351: Nice find! would you mind adding a quick regression test for this 😛

cowboy-bebug: Sure! 🙆‍♂️

jordy25519 (approved):
thanks!

PR #255, opened 2020-07-22
Fix to use right CENNZnut version

KarishmaBothara:
Updated cennznut version to have doughnut working fine.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -9,7 +9,7 @@ repository = "https://github.com/cennznet/cennznet"
 [dependencies]
 # third-party dependencies
 codec = { version = "1.3.0", package = "parity-scale-codec", default-features = false, features = ["derive"] }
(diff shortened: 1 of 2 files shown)


runtime/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = "https://github.com/cennznet/cennznet"
 [dependencies]
 # third-party dependencies
 codec = { version = "1.3.0", package = "parity-scale-codec", default-features = false, features = ["derive"] }
-cennznut = { git = "https://github.com/cennznet/cennznut-rs", default-features = false, branch = "0.4.0" }
+cennznut = { git = "https://github.com/cennznet/cennznut-rs", default-features = false, branch = "0.5.0" }
```
alexsednz, line 12: Why do we have a branch instead of the version here? Should we fix cennznut versioning @holygits?

aliXsed (approved):

PR #258, opened 2020-08-06
Unbond should consume active stake less than minimum bond

jordy25519:
Addresses a few bugs stemming from `unbond` leaving accounts active in the staking system.

Changes:
- Require that `minimum bond` is available to validate and/or nominate
- Unbond consumes all stake if remainder is less than minimum bond
- Unbond 'chills' the stash if remainder is less than minimum bond
- Clarify confusion between active bond usable for voting and slashable bond
- Clarify doc comments and intent in some places
- Ignore dust currency check

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1556,10 +1556,6 @@ impl<T: Trait> Module<T> {
 		let mut all_validators = Vec::new();
 		for (validator, preference) in <Validators<T>>::iter() {
 			let active_bond = Self::active_balance_of(&validator);
-			// Do not consdier zero bonds for election.
-			if active_bond.is_zero() {
-				continue;
-			}
 			let self_vote = (validator.clone(), active_bond, vec![validator.clone()]);
 			all_nominators.push(self_vote);
 			all_validators_and_prefs.insert(validator.clone(), preference);
@@ -1585,7 +1581,6 @@ impl<T: Trait> Module<T> {
 			let s = Self::active_balance_of(&n);
 			(n, s, ns)
 		}));
-		all_nominators.retain(|(_, active, _)| !active.is_zero());
 
 		let maybe_phragmen_result = sp_phragmen::elect::<_, _, T::CurrencyToVote, Perbill>(
 			Self::validator_count() as usize,
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1004,8 +1004,9 @@ decl_module! {
 			if !value.is_zero() {
 				ledger.active -= value;
 
-				// Avoid there being a dust balance left in the staking system.
-				if ledger.active < T::Currency::minimum_balance() {
+				// If active stake drops below the minimum bond threshold or currency minimum
+				// then the entirety of the stash should be unbonded here.
+				if ledger.active < T::Currency::minimum_balance() || ledger.active < Self::minimum_bond() {
 					value += ledger.active;
 					ledger.active = Zero::zero();
 				}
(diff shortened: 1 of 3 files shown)


crml/staking/src/lib.rs
```diff
@@ -999,21 +1006,28 @@ decl_module! {
...
+				return Ok(());
+			}
 
-				let era = Self::current_era() + T::BondingDuration::get();
+			// If active stake drops below the minimum bond threshold or currency minimum
+			// then the entirety of the stash should be scheduled to unlock.
+			// Care must be taken to ensure that funds are still at stake until the unlocking period is over.
+			let remaining_active = ledger.active.checked_sub(&value).unwrap_or(Zero::zero());
+			let era = Self::current_era() + T::BondingDuration::get();
+			if remaining_active < T::Currency::minimum_balance() || remaining_active < Self::minimum_bond() {
```
aliXsed: I think we should only check the active has not gone under the minimum bond. The currency's minimum balance doesn't seem to be relevant in the context of bonding.

crml/staking/src/lib.rs
```diff
@@ -1529,11 +1555,12 @@ impl<T: Trait> Module<T> {
...
 		let mut all_validators = Vec::new();
 		for (validator, preference) in <Validators<T>>::iter() {
-			let self_vote = (
-				validator.clone(),
-				Self::slashable_balance_of(&validator),
-				vec![validator.clone()],
-			);
+			let active_bond = Self::active_balance_of(&validator);
+			// Do not consdier zero bonds for election.
+			if active_bond.is_zero() {
```
jordy25519: planning to include some tests around this behaviour

crml/staking/src/lib.rs
```diff
@@ -378,20 +378,25 @@ pub struct UnlockChunk<Balance: HasCompact> {
...
-	/// Any balance that is becoming free, which may eventually be transferred out
+	/// Any balance that has been unbonded and becoming free, which may eventually be transferred out
 	/// of the stash (assuming it doesn't get slashed first).
 	pub unlocking: Vec<UnlockChunk<Balance>>,
 }
 
 impl<AccountId, Balance: HasCompact + Copy + Saturating + AtLeast32Bit> StakingLedger<AccountId, Balance> {
+	/// The amount of stash's funds slashable as of right now.
+	/// It should remain slashable until the bonding duration expires and it is withdrawn.
+	fn slashable_balance(&self) -> Balance {
```
aliXsed, line 397: As this is not a public getter, it seems adding doc comments to "self.total" and or renaming it could do the job and could be better than a function call.

crml/staking/src/tests.rs
```diff
@@ -2997,3 +3051,106 @@ fn set_minimum_bond_works() {
...
+		));
+		assert_ok!(Staking::nominate(Origin::signed(controller), vec![stash]));
+		// Unbond just enough to put the active stash below the minimum bond amount
+		assert_ok!(Staking::unbond(Origin::signed(controller), 51));
+		assert!(!<Staking as crate::Store>::Nominators::contains_key(stash));
+	});
+}
+
+#[test]
+fn nominate_should_fail_when_stash_does_not_have_minimum_bond() {}
```
aliXsed: I guess you have forgotten to remove this empty tests, right?

crml/staking/src/tests.rs
```diff
@@ -2997,3 +3051,106 @@ fn set_minimum_bond_works() {
...
+
+#[test]
+fn validate_should_fail_when_stash_does_not_have_minimum_bond() {}
```
aliXsed: Same

crml/staking/src/tests.rs
```diff
@@ -2997,3 +3051,106 @@ fn set_minimum_bond_works() {
...
+
+#[test]
+fn validators_with_insufficent_active_bond_are_not_elected() {}
```
aliXsed: Same

aliXsed:
Looks good and necessary changes to me. Thank you for that. I just left a few comments.

crml/staking/src/lib.rs
jordy25519: makes sense to me. the diff will be quite big so will address in a follow up PR

aliXsed (approved):

PR #259, opened 2020-08-21
4dp transaction fees

jordy25519:
Normalise transaction fees to the right magnitude for a 4dp currency `~(0.0001 - 10.0000)`

**Adds**:
- Add `compute_fee_parts` which returns the breakdown of a transaction fees parts + some example tests (this part could be removed and done in another PR)

**Changes**:
- Weight to fee calculation linearly maps `[min_weight, max_weight]` into `[min_weight_fee, max_weight_fee]`
- `TransactionMinWeightFee` and `TransactionMaxWeightFee` can be configured in the runtime
- Replace currency denominations in runtime, genesis config, and tests with (`DOLLARS/1.0000` and `MICROS/0.0001`)
- refactored `compute_fee` to use `compute_fee_parts` internally

**Fixes**:
- Replace 90% magic number fee constants in tests with calculated values

**Concerns**:
- Contract fees still need some investigation and are a bit opaque around gas pricing
- Future Work: Intending to expose `compute_fee_parts` via RPC to give better transparency on tx fee values

### DIFF
--- a/crml/transaction-payment/src/lib.rs
+++ b/crml/transaction-payment/src/lib.rs
@@ -161,26 +161,36 @@ impl<T: Trait> Module<T> {
 	}
 }
 
-#[derive(Debug)]
-pub struct FeeParts {
+#[derive(Debug, Encode, Decode, Clone, Eq, PartialEq)]
+/// The variable parts of a transaction's fees
+/// It does not store the `base_fee` as it is a runtime constant
+/// The `tip` is also excluded as it is known by the transactor/caller.
+pub struct FeeParts<Balance: Saturating> {
 	/// The length fee
-	length_fee: u128,
-	/// The weight fee after conversion
-	weight_fee: u128,
-	/// The len + fee after adjustment factor is applied
-	adjusted_fee: u128,
-	/// The final fee
-	final_fee: u128,
+	/// It changes based on the size of the transaction.
+	length_fee: Balance,
+	/// The weight fee
+	/// It changes based on the computational resources required by the transaction.
+	weight_fee: Balance,
+	/// The peak adjustment fee.
+	/// It changes based on recent network transaction load (or more formally "block fullness").
+	peak_adjustment_fee: Balance,
 }
 
-impl FeeParts {
-	fn new(length_fee: u128, weight_fee: u128, adjusted_fee: u128, final_fee: u128) -> Self {
+impl<Balance: Copy + Saturating> FeeParts<Balance> {
+	pub fn new(length_fee: Balance, weight_fee: Balance, peak_adjustment_fee: Balance) -> Self {
 		FeeParts {
 			length_fee,
 			weight_fee,
-			adjusted_fee,
-			final_fee,
+			peak_adjustment_fee,
 		}
+	}
+	/// Calculate the total fee from it's parts
+	pub fn total(&self) -> Balance {
+		// total = length_fee + weight_fee + peak_adjustment_fee
+		self.length_fee
+			.saturating_add(self.weight_fee)
+			.saturating_add(self.peak_adjustment_fee)
 	}
 }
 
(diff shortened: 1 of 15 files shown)


crml/transaction-payment/src/lib.rs
```diff
@@ -161,6 +161,29 @@ impl<T: Trait> Module<T> {
...
+#[derive(Debug)]
+pub struct FeeParts {
+	/// The length fee
+	length_fee: u128,
+	/// The weight fee after conversion
+	weight_fee: u128,
+	/// The len + fee after adjustment factor is applied
+	adjusted_fee: u128,
+	/// The final fee
+	final_fee: u128,
```
aliXsed: If we remove this field, we have prevented from a possible discrepancy between the other fields and this one. Instead, we can take the lazy evaluation approach and calculate final_fee only when final_fee is called.

aliXsed: In fact, I would want to suggest that we only keep the actual building blocks of a fee, which is the length of the encoded tx, its weight factor and the adjustment factor. Everything else like the fee parts should be methods for calculating the elements. Final fee can be a function (not a method) which would call the other methods to calculate the Final fee. Something like the following pseudo-code: 
```
pub struct Fee<Balance> {
   encoded_tx_len: usize,
   tx_weight: Balance,
   network_usage_adjustment: u128
}

impl<Balance> Fee<Balance> {
   fn base_fee(): { x }
   fn per_byte_fee(): { y }
   fn length_fee(&self): Balance { ... }
   fn weight_fee(&self): Balance { ... }
   fn adjusted_fee(&Self x): Balance { (x.length_fee()+x.weight_fee())*x.network_usage_adjustment }
   pub fn final_fee(&Self x): Balance { x.BASE + x.adjusted_fee() }
}
```

aliXsed: The above approach has got the following characteristics:

- It's a lazy evaluation approach. 
- Each private method does only one single cohesive job.
- Only functions are allowed to orchestrate methods for any additional service. Methods are not going to call each other and make an unnecessary intermingled structure. The benefit would be more obvious if later someone needs to add a thread lock to methods for example. With this approach, locks not going to get intertwined or reentered.
- The generic type of Balance is preferred to u128 as it integrates better with the rest of cennznet.
- For each member variable, the preferred type is used so the compiler can get possible type mistakes by type checking.

aliXsed: If we need a cache feature too, then we can use a structure like the following:
https://github.com/alexsednz/learn-rust/blob/master/src/cache.rs#L22-L54

runtime/src/lib.rs
```diff
@@ -412,15 +412,15 @@ impl pallet_treasury::Trait for Runtime {
...
-	pub const ContractTransferFee: Balance = 1 * NANOCENTS;
-	pub const ContractCreationFee: Balance = 1 * MICROCENTS;
-	pub const ContractTransactionBaseFee: Balance = 1 * NANOCENTS;
-	pub const ContractTransactionByteFee: Balance = 10 * MICROCENTS;
-	pub const ContractFee: Balance = 1 * CENTS;
-	pub const TombstoneDeposit: Balance = 1 * DOLLARS;
-	pub const RentByteFee: Balance = 1 * DOLLARS;
-	pub const RentDepositOffset: Balance = 1000 * DOLLARS;
-	pub const SurchargeReward: Balance = 150 * DOLLARS;
+	pub const ContractTransferFee: Balance = 1 * MICROS;
```
jordy25519, line 415: tried to adjust these values in relative proportion to current substrate defaults.
as mentioned in description contract fees need some deeper review

runtime/src/constants.rs
```diff
@@ -35,13 +35,12 @@ pub mod asset {
...
-
-	pub const PICOCENTS: Balance = 1;
-	pub const NANOCENTS: Balance = 1_000 * PICOCENTS;
-	pub const MICROCENTS: Balance = 1_000 * NANOCENTS;
-	pub const MILLICENTS: Balance = 1_000 * MICROCENTS;
-	pub const CENTS: Balance = 1_000 * MILLICENTS; // assume this is worth about a cent.
-	pub const DOLLARS: Balance = 100 * CENTS;
+	/// The smallest denomination of any currency
+	pub const WEI: Balance = 1;
+	/// The smallest denomination of a 4dp currency
```
aliXsed, line 40: Could be better if we hadn't coupled these definitions with 4dp which is configurable for an asset.

runtime/src/impls.rs
```diff
@@ -80,38 +83,38 @@ impl Convert<u128, Balance> for CurrencyToVoteHandler {
...
+
+		// Runtime constants
+		let min_fee = MinFee::get();
+		let max_fee = MaxFee::get();
+		debug_assert!(max_fee > min_fee);
+		debug_assert!(MAX_WEIGHT > MIN_WEIGHT);
+
+		//      (weight - MIN_WEIGHT) * [min_fee, max_fee]
+		//  y = ------------------------------------------ + min_fee
+		//              [MIN_WEIGHT, MAX_WEIGHT]
```
aliXsed, line 103: Nice comment!

runtime/tests/tests.rs
```diff
@@ -1090,67 +1071,76 @@ fn contract_call_works_without_fee_exchange() {
...
+			let gas_cost = (Schedule::default().call_base_cost + Schedule::default().transfer_cost) as u128;
+			// Check CENNZ price to buy gas (gas is 1:1 with CPAY)
+			let cennz_for_gas_fees = CennzxSpot::get_asset_to_core_buy_price(&CENNZ_ASSET_ID, gas_cost).unwrap();
+			// Check CENNZ price to buy tx fees in CPAY
+			let cennz_for_tx_fees =
+				CennzxSpot::get_asset_to_core_buy_price(&CENNZ_ASSET_ID, get_extrinsic_fee(&xt)).unwrap();
+
 			Executive::initialize_block(&header());
-			let r = Executive::apply_extrinsic(xt);
+			let r = Executive::apply_extrinsic(xt.clone());
```
aliXsed, line 1130: Very minor: do you need to clone here?

runtime/tests/tests.rs
```diff
@@ -1305,34 +1282,34 @@ fn generic_asset_transfer_works_with_doughnut() {
 
 			// Initialise block and apply the extrinsic
 			Executive::initialize_block(&header());
-			let r = Executive::apply_extrinsic(xt);
+			let r = Executive::apply_extrinsic(xt.clone());
```
alexsednz, line 1285: Same here

runtime/tests/tests.rs
```diff
@@ -1363,37 +1340,37 @@ fn generic_asset_transfer_fails_with_bad_doughnut_permissions() {
 			// Initialise block and apply the extrinsic
 			Executive::initialize_block(&header());
 			assert_eq!(
-				Executive::apply_extrinsic(xt),
+				Executive::apply_extrinsic(xt.clone()),
```
alexsednz, line 1343: Same here

aliXsed (approved):
LGTM

runtime/tests/tests.rs
jordy25519: I'll get this one in my next PR :D

PR #261, opened 2020-09-01
Feature/260 sylo fees

### LINKED ISSUE #260: Implement Sylo integrated payment
Ensure any sylo module call fees are deducted from a known sylo account.

aliXsed:
Closes #260 

Here is the list of features:
- Only Sudo will be able to set one or more payment accounts for Sylo calls. 
- Only an authorised payment account can revoke itself from being a payment account anymore.
- When there is more than one payment account, at the moment, we will simply use the first one, which is set earlier. In the future, we can make this smarter to go for the account with the highest balance.
- Having more than one account as payment accounts enables a seamless swap of the accounts. If Sylo wants to use a different account, they can simple first add the new one and then remove the old one, without being worried that it can make any pause in their operations.
- I haven't created an easy direct revoke of all the payment accounts simply because it will definitely halt all Sylo operations until a new one is set. So we'd better need to do such operations with more effort rather than less effort.
- The method used here is secure because users cannot modify which call is being paid by Sylo with runtime operations and only Sudo has the privilege to set a payment account.
- When there are no payment accounts for Sylo, we will see if the caller can pay the fee themselves before rejecting the call.

### DIFF
--- a/crml/sylo/src/payment.rs
+++ b/crml/sylo/src/payment.rs
@@ -1,0 +1,141 @@
+/* Copyright 2020 Centrality Investments Limited
+*
+* Licensed under the LGPL, Version 3.0 (the "License");
+* you may not use this file except in compliance with the License.
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+* You may obtain a copy of the License at the root of this project source code,
+* or at:
+*     https://centrality.ai/licenses/gplv3.txt
+*     https://centrality.ai/licenses/lgplv3.txt
+*/
+
+//! Manage the authorized accounts set for the Sylo fee payment
+
+use frame_support::{decl_module, decl_storage, ensure, weights::SimpleDispatchInfo, IterableStorageMap};
+use frame_system::{ensure_root, ensure_signed};
+use sp_runtime::DispatchResult;
+use sp_std::prelude::*;
+
+pub trait Trait: frame_system::Trait {}
+
+const NOT_SYLO_PAYER: &str = "You are not a Sylo payer!";
+
+decl_storage! {
+	trait Store for Module<T: Trait> as SyloModule {
+		/// Accounts which have authority to pay for Sylo fees on behalf of the users
+		AuthorisedPayers: map hasher(twox_64_concat) T::AccountId => bool;
+	}
+}
+
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin, system = frame_system {
+
+		/// Add `account_id` as an authorized Sylo fee payer. Only Sudo can set a payment account.
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		pub fn set_payment_account(origin, account_id: T::AccountId) -> DispatchResult {
+			ensure_root(origin)?;
+			AuthorisedPayers::<T>::insert(account_id, true);
+			Ok(())
+		}
+
+		/// If the origin of the call is an authorised payer, revoke its authorisation.
+		/// NOTE: This may halt all Sylo operations if there are no other payers.
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		pub fn revoke_me(origin) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			ensure!(AuthorisedPayers::<T>::get(&account_id), NOT_SYLO_PAYER);
+			AuthorisedPayers::<T>::remove(account_id);
+			Ok(())
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Return the first account that is set for payment, or None when nothing is set.
+	/// In the future, we can make this function smart so it returns the account with enough money in it.
+	pub fn get_payment_account() -> Option<T::AccountId> {
+		AuthorisedPayers::<T>::iter().map(|(x, _)| x).next()
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::mock::{ExtBuilder, Origin, Test};
+	use frame_support::assert_ok;
+	use frame_system::RawOrigin;
+	use sp_core::H256;
+	use sp_runtime::DispatchError::Other;
+
+	type SyloModule = Module<Test>;
+
+	impl Trait for Test {}
+
+	#[test]
+	fn set_payment_account() {
+		ExtBuilder::default().build().execute_with(|| {
+			let payer_a = H256::from_low_u64_be(2);
+			let payer_b = H256::from_low_u64_be(3);
+
+			assert!(!AuthorisedPayers::<Test>::get(&payer_a));
+			assert!(!AuthorisedPayers::<Test>::get(&payer_b));
+
+			assert_ok!(SyloModule::set_payment_account(Origin::ROOT, payer_a));
+			assert_ok!(SyloModule::set_payment_account(Origin::ROOT, payer_b));
+
+			assert!(AuthorisedPayers::<Test>::get(&payer_a));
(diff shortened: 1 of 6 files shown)


crml/sylo/src/payment.rs
```diff
@@ -0,0 +1,141 @@
...
+			Ok(())
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Return the first account that is set for payment, or None when nothing is set.
+	/// In the future, we can make this function smart so it returns the account with enough money in it.
+	pub fn get_payment_account() -> Option<T::AccountId> {
+		AuthorisedPayers::<T>::iter().map(|(x, _)| x).next()
```
jordy25519, line 61: I'm not certain that `::iter()` is based on insertion order. Maybe a Vec will be sufficient here

crml/transaction-payment/src/lib.rs
```diff
@@ -103,6 +109,8 @@ pub trait Trait: frame_system::Trait {
 
 	/// Something which can report whether a call is gas metered
 	type GasMeteredCallResolver: IsGasMeteredCall<Call = <Self as frame_system::Trait>::Call>;
+
+	type FeePayer: FeePayer<Call = Self::Call, AccountId = Self::AccountId>;
```
jordy25519, line 123: Doc comment

crml/sylo/src/payment.rs
```diff
@@ -0,0 +1,141 @@
...
+			AuthorisedPayers::<T>::remove(account_id);
+			Ok(())
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Return the first account that is set for payment, or None when nothing is set.
+	/// In the future, we can make this function smart so it returns the account with enough money in it.
+	pub fn get_payment_account() -> Option<T::AccountId> {
```
jordy25519: convention for getters is to not prefix with `get_`

crml/sylo/src/payment.rs
```diff
@@ -0,0 +1,141 @@
...
+		pub fn set_payment_account(origin, account_id: T::AccountId) -> DispatchResult {
+			ensure_root(origin)?;
+			AuthorisedPayers::<T>::insert(account_id, true);
+			Ok(())
+		}
+
+		/// If the origin of the call is an authorised payer, revoke its authorisation.
+		/// NOTE: This may halt all Sylo operations if there are no other payers.
+		#[weight = SimpleDispatchInfo::FixedOperational(0)]
+		pub fn revoke_me(origin) -> DispatchResult {
```
jordy25519: nit: give name some context just incase e.g.`revoke_payment_account_self`

crml/sylo/src/payment.rs
```diff
@@ -0,0 +1,141 @@
...
+use frame_system::{ensure_root, ensure_signed};
+use sp_runtime::DispatchResult;
+use sp_std::prelude::*;
+
+pub trait Trait: frame_system::Trait {}
+
+const NOT_SYLO_PAYER: &str = "You are not a Sylo payer!";
+
+decl_storage! {
+	trait Store for Module<T: Trait> as SyloModule {
```
jordy25519:
```suggestion
	trait Store for Module<T: Trait> as SyloFeePayment {
```
or something more descriptive

jordy25519 (commented):
general approach lgtm!

crml/transaction-payment/src/lib.rs
aliXsed: Done.

crml/sylo/src/payment.rs
aliXsed: Done

crml/sylo/src/payment.rs
aliXsed: Done

crml/sylo/src/payment.rs
aliXsed: Done

crml/sylo/src/payment.rs
```diff
@@ -0,0 +1,143 @@
...
+			let account_id = ensure_signed(origin)?;
+			ensure!(AuthorisedPayers::<T>::get(&account_id), NOT_SYLO_PAYER);
+			AuthorisedPayers::<T>::remove(account_id);
+			Ok(())
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Return the first account that is set for payment, or None when nothing is set.
```
jordy25519:
```suggestion
	/// Return an account that is set for payment, or `None` when nothing is set.
```

crml/transaction-payment/src/lib.rs
```diff
@@ -67,6 +67,12 @@ type NegativeImbalanceOf<T> =
 
 pub const GAS_FEE_EXCHANGE_KEY: &[u8] = b"gas-fee-exchange-key";
 
+pub trait FeePayer {
```
jordy25519, line 73: Doc comments for this trait and types

runtime/src/impls.rs
```diff
@@ -304,6 +304,28 @@ impl IsGasMeteredCall for GasMeteredCallResolver {
 	}
 }
 
+pub struct FeePayerResolver;
```
jordy25519, line 308: doc comments

crml/transaction-payment/src/lib.rs
```diff
@@ -310,9 +319,10 @@ where
 
 		// Only mess with balances if the fee is not zero.
 		if !fee.is_zero() {
+			let payer = T::FeePayer::fee_payer(call).unwrap_or(who.clone());
```
jordy25519, line 331:
```suggestion
			let payer = T::FeePayer::fee_payer(call).unwrap_or_else(|| who.clone());
```
might help with the clone

runtime/tests/tests.rs
```diff
@@ -683,6 +685,147 @@ fn runtime_mock_setup_works() {
...
+	Executive::initialize_block(&header());
+	let r = Executive::apply_extrinsic(xt);
+	assert!(r.is_ok());
+
+	fee
+}
+
+#[test]
+fn non_sylo_call_is_not_paid_by_payment_account() {
+	let call = Call::GenericAsset(pallet_generic_asset::Call::transfer(CENTRAPAY_ASSET_ID, dave(), 100));
```
jordy25519, line 705: suggestion: a single test with a for loop of all the sylo calls would be clearer with less boiler plate

jordy25519 (approved):
please add doc comments
also left some suggestions
otherwise lgtm

runtime/tests/tests.rs
aliXsed: I tried that first but then the test needs to be more complicated as there was an issue possibly with applying more than one extrinsic to the same block. On the other hand, we would usually want our test cases to do one test at a time.

crml/transaction-payment/src/lib.rs
aliXsed: Would you explain more?

jordy25519: `unwrap_or` will eagerly evaluate the `.clone()`, `unwrap_or_else` will only evaluate it when needed (if payer is None)

aliXsed: Ah you mean we do the clone only if necessary, right?

runtime/tests/tests.rs
jordy25519: I guess you'd need to make a new block each loop

crml/transaction-payment/src/lib.rs
aliXsed: But I'm not sure what would be stored for the closure then. It may defeat the purpose.

crml/transaction-payment/src/lib.rs
aliXsed: Done

runtime/src/impls.rs
aliXsed: Done

crml/sylo/src/payment.rs
aliXsed: Done

PR #262, opened 2020-09-03
Use vec instead of map for storing authorised payers

aliXsed:
Relates to #260 

As per @holygits's comment here: https://github.com/cennznet/cennznet/pull/261#discussion_r481513916, this PR substitutes map with vec for storing the list of authorised payers which seems to be the more appropriate data structure for the use here.


jordy25519 (approved):
lgtm!

PR #263, opened 2020-09-07
Modularise runtime integration tests

jordy25519:
Address some tech debt in our runtime integration tests

- All related tests grouped into independent files
- common helpers / mock moved into `common` module

smaller refactors:
- pass initial validator set explicitly to mock storage
- replace `transfer_fee` with `extrinsic_fee_for` which uses the proper runtime impl instead of copy-paste code

### DIFF
--- a/runtime/tests/common/helpers.rs
+++ b/runtime/tests/common/helpers.rs
@@ -23,7 +23,7 @@ use crml_transaction_payment::ChargeTransactionPayment;
 use frame_support::weights::GetDispatchInfo;
 use sp_runtime::{testing::Digest, traits::Header as HeaderT};
 
-/// A genesis hash to use for exstrinsic signing
+/// A genesis hash to use for extrinsic signing
 const GENESIS_HASH: [u8; 32] = [69u8; 32];
 /// The runtime version number for signing
 const VERSION: u32 = cennznet_runtime::VERSION.spec_version;
--- a/runtime/tests/common/helpers.rs
+++ b/runtime/tests/common/helpers.rs
@@ -1,0 +1,67 @@
+/* Copyright 2019-2020 Centrality Investments Limited
+*
+* Licensed under the LGPL, Version 3.0 (the "License");
+* you may not use this file except in compliance with the License.
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


runtime/tests/common/mock.rs
```diff
@@ -74,6 +57,10 @@ impl ExtBuilder {
 		self.initial_balance = initial_balance;
 		self
 	}
+	pub fn initial_authorities(mut self, initial_authorities: &[AuthorityKeys]) -> Self {
```
aliXsed, line 63: set_initial_authorities seems to be a better name

runtime/tests/common/mock.rs
```diff
@@ -90,13 +77,13 @@ impl ExtBuilder {
...
 	}
-	pub fn validator_count(mut self, count: usize) -> Self {
-		self.validator_count = count;
-		self
-	}
 	pub fn build(self) -> sp_io::TestExternalities {
 		let mut endowed_accounts = vec![alice(), bob(), charlie(), dave(), eve(), ferdie()];
-		let initial_authorities = generate_initial_authorities(self.validator_count);
+		let initial_authorities = if self.initial_authorities.is_empty() {
+			make_authority_keys(3)
```
aliXsed: Is it good to have a defined const for 3 here?

aliXsed (approved):
LGTM, I just suggested two minor changes.

runtime/tests/common/helpers.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use cennznet_cli::chain_spec::{get_authority_keys_from_seed, AuthorityKeys};
+use cennznet_primitives::types::{Balance, BlockNumber, Header};
+use cennznet_runtime::{CheckedExtrinsic, Runtime, UncheckedExtrinsic};
+use codec::Encode;
+use crml_transaction_payment::ChargeTransactionPayment;
+use frame_support::weights::GetDispatchInfo;
+use sp_runtime::{testing::Digest, traits::Header as HeaderT};
+
+/// A genesis hash to use for exstrinsic signing
```
KarishmaBothara: /// A genesis hash to use for **extrinsic** signing

KarishmaBothara (approved):

runtime/tests/common/mock.rs
jordy25519: It seems to be convention for the rust [builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html) to name setters like this so will leave it as is for consistency

PR #272, opened 2020-09-17
Create a custom image for the scanner project

aliXsed:
This will bring minimal changes to cennznet-1.1.0 which is needed for browsing historical payment info.


jordy25519:
changes lgtm but this should become `1.1.1` as an official release

PR #278, opened 2020-09-20
Fix docker to use build time context

### LINKED ISSUE #277: Fix docker to use same rust version as CircleCI
When building the docker image we should use the same version of Rust as in CircleCI's context.

aliXsed:
Bring the same change as in the release 1.1.1 to "develop"
Closes #277

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,13 +2,17 @@ FROM  rustlang/rust:nightly AS builder
 WORKDIR /cennznet
 COPY . /cennznet
 
-ENV RUST_VERSION 1.44.1
+ARG RUST_VERSION=1.44.1
+ARG RUST_NIGHTLY=nightly-2020-06-24
 RUN apt-get update && \
     apt-get -y install apt-utils cmake pkg-config libssl-dev git clang libclang-dev && \
+    rustup uninstall nightly && \
     rustup install $RUST_VERSION && \
+    rustup install $RUST_NIGHTLY && \
     rustup default $RUST_VERSION && \
-    rustup target add --toolchain nightly wasm32-unknown-unknown && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,13 +2,17 @@ FROM  rustlang/rust:nightly AS builder
 WORKDIR /cennznet
 COPY . /cennznet
 
-ENV RUST_VERSION 1.44.1
+ARG RUST_VERSION=1.44.1
```
aliXsed, line 5: By the way, 1.44.1 is the default version when no argument is passed to the docker. The actual version is coming from the Rust CircileCI context.

Dockerfile
```diff
@@ -2,13 +2,17 @@ FROM  rustlang/rust:nightly AS builder
...
 RUN apt-get update && \
     apt-get -y install apt-utils cmake pkg-config libssl-dev git clang libclang-dev && \
+    rustup uninstall nightly && \
     rustup install $RUST_VERSION && \
+    rustup install $RUST_NIGHTLY && \
     rustup default $RUST_VERSION && \
-    rustup target add --toolchain nightly wasm32-unknown-unknown && \
+    rustup target add --toolchain $RUST_NIGHTLY wasm32-unknown-unknown && \
     rustup target add --toolchain $RUST_VERSION x86_64-unknown-linux-musl && \
+    mv /usr/local/rustup/toolchains/nightly* /usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu && \
```
jordy25519, line 15: would be nice to leave a comment about what this does

jordy25519 (approved):
lgtm

Dockerfile
aliXsed: I will, in a subsequent PR

PR #285, opened 2020-09-28
Chore/cennz slash

KarishmaBothara:
This PR is to address the feature, 'The CENNZnet treasury should receive CPAY from reward payouts and CENNZ from slashes'. 
Have tested it with a test case where when slash is applied, call to pay_reporters::<T>(reward_payout, slashed_imbalance, &unapplied_slash.reporters); is made and slash goes to Treasury in CENNZ

### DIFF
--- a/runtime/tests/staking_reward.rs
+++ b/runtime/tests/staking_reward.rs
@@ -20,7 +20,7 @@ use cennznet_primitives::types::{AccountId, Balance, DigestItem};
 use cennznet_runtime::{
 	constants::{asset::*, currency::*},
 	Babe, Call, CheckedExtrinsic, EpochDuration, Executive, GenericAsset, Header, ImOnline, Runtime, Session,
-	SessionsPerEra, Staking, System, Timestamp,
+	SessionsPerEra, Staking, System, Timestamp, Treasury,
 };
 use cennznet_testing::keyring::{alice, bob, charlie, signed_extra};
 use codec::Encode;
@@ -37,6 +37,7 @@ mod common;
 
 use common::helpers::{extrinsic_fee_for, header, header_for_block_number, make_authority_keys, sign};
 use common::mock::ExtBuilder;
+use frame_support::additional_traits::MultiCurrencyAccounting;
 
 /// Get a list of stash accounts only from `authority_keys`
 fn stashes_of(authority_keys: &[AuthorityKeys]) -> Vec<AccountId> {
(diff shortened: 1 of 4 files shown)


runtime/tests/staking_reward.rs
```diff
@@ -549,3 +550,50 @@ fn authorship_reward_of_a_chilled_validator() {
...
+			);
+			let _ = <GenericAsset as MultiCurrencyAccounting>::deposit_creating(
+				&Treasury::account_id(),
+				Some(SPENDING_ASSET_ID),
+				cpay_amount,
+			);
+			assert_eq!(Staking::current_era(), 0);
+
+			// function `apply_unapplied_slashes` is private
+			// Staking::apply_unapplied_slashes(era_now);
```
aliXsed: Can you please remove the commented code?

runtime/tests/staking_reward.rs
```diff
@@ -549,3 +550,50 @@ fn authorship_reward_of_a_chilled_validator() {
...
 }
+
+#[test]
+/// This tests if slash goes to treasury as CENNZ.
+fn slashed_cennz_gets_into_treasury() {
+	let validators = make_authority_keys(6);
+	let initial_balance = 1_000 * DOLLARS;
+
+	ExtBuilder::default()
+		.slash_defer_duration(0)
```
aliXsed: Shouldn't this be 168 or even not specified to get its default value?

runtime/tests/staking_reward.rs
```diff
@@ -549,3 +550,50 @@ fn authorship_reward_of_a_chilled_validator() {
...
+			// function `apply_unapplied_slashes` is private
+			// Staking::apply_unapplied_slashes(era_now);
+
+			// slashing::apply_slash(unapplied[0]); // slashing module is private to staking and apply_slash is also private
+			// Default slash_defer_duration is 168, so have to set erra to 169 for slash to be applied.
+			start_era(169); // new_era function will internally call apply_unapplied_slashes
+			assert_eq!(Staking::current_era(), 169);
+			assert_eq!(
+				GenericAsset::free_balance(&CENNZ_ASSET_ID, &Treasury::account_id()),
+				updated_cennz_amount
```
aliXsed: What kind of misbehaviour could cause a slashing here?

runtime/tests/staking_reward.rs
```diff
@@ -549,3 +550,50 @@ fn authorship_reward_of_a_chilled_validator() {
...
+				&Treasury::account_id(),
+				Some(STAKING_ASSET_ID),
+				cennz_amount,
+			);
+			let _ = <GenericAsset as MultiCurrencyAccounting>::deposit_creating(
+				&Treasury::account_id(),
+				Some(SPENDING_ASSET_ID),
+				cpay_amount,
+			);
+			assert_eq!(Staking::current_era(), 0);
```
aliXsed: You need to check the treasury balances here so we know what it was before and what it became after the slashing.

runtime/tests/staking_reward.rs
KarishmaBothara: The defer duration is set as constant - pub const SlashDeferDuration: crml_staking::EraIndex = 24 * 7 at https://github.com/cennznet/cennznet/blob/b40b1afbd7a226823e08fb069b3b0b3ba92592aa/runtime/src/lib.rs#L313 so can't change it... I have reverted all the code which was used to set it to configurable value...

runtime/tests/staking_reward.rs
KarishmaBothara: Unresponsive offence reported by
Start_era -> calls start_session -> Session::on_initialize -> which rotates the session as it is last block of the current session
rotate_session() -> Informs the session handler that a session is going to end ->
Handler:: on_before_session_ending() -> reports - UnresponsivenessOffence (from im-online)

runtime/tests/staking_reward.rs
```diff
@@ -556,37 +557,37 @@ fn authorship_reward_of_a_chilled_validator() {
...
+			// basically, 10% can be offline with no slash, but after that, it linearly climbs up to 7%
+			// when 13/30 are offline (around 5% when 1/3 are offline).
+			if let Some(threshold) = offenders.checked_sub(validator_set_count / 10 + 1) {
+				let x = Perbill::from_rational_approximation(3 * threshold, validator_set_count);
+				slashed_amount = x.saturating_mul(Perbill::from_percent(7));
+			} else {
+				slashed_amount = Perbill::default()
+			}
+			let own_slash = slashed_amount * initial_balance;
+			let total_slashed_cennz = own_slash.saturating_mul(offenders.into());
```
jordy25519: This looks good! can you put this function in staking module or (re)use the same code as the staking module, copy-pasting code can lead to brittle tests

jordy25519 (approved):
lgtm 👏

PR #287, opened 2020-10-06
Update CENNZX

jordy25519:
Config/impl CENNZX in runtime
Impl GA/CENNZX rpcs also

- [x] fix tests

### DIFF
--- a/crml/cennzx/src/impls.rs
+++ b/crml/cennzx/src/impls.rs
@@ -15,18 +15,50 @@
 //!
 //! Extra CENNZX-Spot traits + implementations
 //!
-use crate::{Module, Trait, MODULE_ID};
+use crate::{Module, Trait};
 use cennznet_primitives::{
 	traits::{BuyFeeAsset, SimpleAssetSystem},
 	types::FeeExchange,
 };
-use frame_support::dispatch::DispatchError;
-use sp_runtime::traits::AccountIdConversion;
+use frame_support::dispatch::{DispatchError, DispatchResult};
+use prml_generic_asset::{AssetIdAuthority, MultiCurrencyAccounting};
+use sp_core::crypto::{UncheckedFrom, UncheckedInto};
+use sp_runtime::traits::Hash;
+use sp_std::{marker::PhantomData, prelude::*};
 
-/// Generate a CENNZX exchange address for a core, `asset_id` pair
-pub fn exchange_address_for<T: Trait>(asset_id: T::AssetId) -> T::AccountId {
-	let core_asset_id = Module::<T>::core_asset_id();
-	MODULE_ID.into_sub_account((core_asset_id, asset_id))
+/// A function that generates an `AccountId` for a CENNZX exchange / (core, asset) pair
+pub trait ExchangeAddressFor {
+	/// The Account Id type
+	type AccountId;
+	/// The Asset Id type
+	type AssetId;
+	/// Create and exchange address given `asset_id`
+	fn exchange_address_for(asset_id: Self::AssetId) -> Self::AccountId;
+}
+
+/// A CENNZX exchange address generator implementation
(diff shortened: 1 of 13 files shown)


crml/cennzx/src/mock.rs
```diff
@@ -48,108 +51,49 @@ impl_outer_origin! {
...
-	type BalanceToUnsignedInt = LowPrecisionUnsigned;
-	type UnsignedIntToBalance = UnsignedIntToBalance;
+	type AssetId = AssetId;
+	type Balance = Balance;
+	type AssetSystem = SimpleAssetShim<Self>;
+	type ExchangeAddressFor = ExchangeAddressGenerator<Self>;
+	type WeightInfo = ();
 }
-
-pub type Cennzx = Module<Test>;
```
jordy25519, line 105: multi-curreny impl in GA replaced all this stuff

crml/cennzx/src/impls.rs
```diff
@@ -15,26 +15,58 @@
...
+}
+
+/// A CENNZX exchange address generator implementation
+pub struct ExchangeAddressGenerator<T: Trait>(PhantomData<T>);
 
-/// Generate a CENNZX exchange address for a core, `asset_id` pair
-pub fn exchange_address_for<T: Trait>(asset_id: T::AssetId) -> T::AccountId {
-	let core_asset_id = Module::<T>::core_asset_id();
-	MODULE_ID.into_sub_account((core_asset_id, asset_id))
+impl<T: Trait> ExchangeAddressFor for ExchangeAddressGenerator<T>
```
jordy25519, line 41: - ModuleId doesn't provide enough bytes to be used here
- Couldn't find a way to refactor this as a function.
It needs the type awareness provided by associated types when configured in the runtime
to satisfy the trait bounds here

aliXsed: 8 bytes (ModuleId) could be enough for making unique exchange addresses, right? Looking at other module ids:
b"phrelect", b"py/trsry", b"py/socie", a name like b"cennz/xs" or b"cennz-xs" or similar could be a good candidate, I'm thinking.
Do we have any backwards compatibility issue here? Like previously generated exchange addresses?

crml/cennzx/src/impls.rs
```diff
@@ -54,29 +86,61 @@ impl<T: Trait> BuyFeeAsset for Module<T> {
...
+	/// Get the liquid asset balance of `account`
+	fn free_balance(asset_id: Self::AssetId, account: &Self::AccountId) -> Self::Balance {
+		prml_generic_asset::Module::<T>::free_balance(asset_id, account)
+	}
+	/// Get the default asset/currency ID in the system
+	fn default_asset_id() -> Self::AssetId {
+		<prml_generic_asset::Module<T> as MultiCurrencyAccounting>::DefaultCurrencyId::asset_id()
+	}
+}
+
```
aliXsed: I'm thinking runtime/src/impl.rs could be a better place for this implementation as we could then avoid coupling cennzx to generic asset internally.

crml/cennzx/src/impls.rs
jordy25519: ModuleId will give us one address but at the moment we need a unique address for each pool.
We could use into_sub_account but that relies on the encode implementation to guarantee uniqueness and truncates at 32 bytes 😬

so module_id.into_sub_account((core_id, asset_id)) won't take into account the full 128 bytes of entropy in the asset IDs which could lead to the same addresses being used which will throw off the liquidity accounting.

Hashing the bytes seems to be the only way to reasonably guarantee this is unique.

aliXsed (commented):
LVGTM! Nice Job!

aliXsed (approved):

KarishmaBothara (approved):

PR #290, opened 2020-10-08
Update/transaction payment

aliXsed:
Update cennznet transaction payment to be compatible with Pl^g 2.0 and enable it in cennznet runtime


KarishmaBothara (approved):
lgtm as non rust expert

PR #291, opened 2020-10-12
Make execute_trade atomic

jordy25519:
This prevents storage changes from persisting if something fails during intermediate operations (new feature in 2.0)

### DIFF
--- a/crml/cennzx/src/lib.rs
+++ b/crml/cennzx/src/lib.rs
@@ -21,11 +21,14 @@
 use cennznet_primitives::traits::SimpleAssetSystem;
 use core::convert::TryFrom;
 use frame_support::{
-	decl_error, decl_event, decl_module, decl_storage, ensure, weights::Weight, Parameter, StorageDoubleMap,
+	decl_error, decl_event, decl_module, decl_storage, ensure, transactional, weights::Weight, Parameter,
+	StorageDoubleMap,
 };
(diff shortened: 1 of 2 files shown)


crml/cennzx/src/lib.rs
```diff
@@ -781,29 +787,29 @@ impl<T: Trait> Module<T> {
 			} else {
 				T::ExchangeAddressFor::exchange_address_for(asset_to_buy)
 			};
-			let _ = T::AssetSystem::transfer(asset_to_sell, trader, &exchange_address, amount_to_sell).and(
+
+			T::AssetSystem::transfer(asset_to_sell, trader, &exchange_address, amount_to_sell).and(
```
KarishmaBothara, line 791: should **.and(** be like
**and_then**  like in 
Self::get_asset_to_core_sell_price(asset_to_sell, amount_to_sell).and_then(|core_amount| {
 				T::AssetSystem::transfer(asset_to_sell, trader, &exchange_address_a, amount_to_sell)

KarishmaBothara (approved):

crml/cennzx/src/lib.rs
jordy25519: afaik `and` will return err if either error
in the 2nd case I used `and_then` so I could pass through the `core_amount` from price calculation

aliXsed (approved):
LGTM!

PR #292, opened 2020-10-13
Update/2.0.0 staking update

aliXsed:
Staking module builds okay and passes its own tests with this update. Staking module is also enabled in the runtime. The whole cennznet build is okay. But as it was the case with the previous module of this branch, runtime tests still failing as the whole branch is still a WIP. However, I suggest we merge the changes at this stage and fix the runtime/cli and etc tests in another step.


crml/staking/src/mock.rs
```diff
@@ -111,15 +110,19 @@ impl_outer_origin! {
 }
 
 mod staking {
-	pub use crate::Event;
+	// Re-export needed for `impl_outer_event!`.
+	pub use super::super::*;
```
jordy25519, line 114: (very small nit) doesn't this still work: `pub use crate::Event;`? I think staking::Event is all that's needed

runtime/src/impls.rs
```diff
@@ -173,39 +173,55 @@ impl<T: prml_generic_asset::Trait> SimpleAssetSystem for SimpleAssetShim<T> {
...
 	fn target() -> Weight {
 		TargetBlockFullness::get() * max()
 	}
 
-	// poc reference implementation.
-	fn fee_multiplier_update(block_weight: Weight, previous: Fixed64) -> Fixed64 {
-		let block_weight = block_weight as f32;
-		let v: f32 = 0.00004;
+	// update based on runtime impl.
+	fn runtime_multiplier_update(fm: Multiplier) -> Multiplier {
```
jordy25519, line 199: these changes are abit hard to review, are they copied from plug?

jordy25519 (approved):
lgtm!

runtime/src/impls.rs
aliXsed: Yes

crml/staking/src/mock.rs
aliXsed: I've made it more like what it is in Plug.

PR #293, opened 2020-10-14
Update/fix test build

aliXsed:
This PR fixes the cennznet (update/2.0.0) build for all targets. It also passes 256 tests but fails the following 16:
Screen Shot 2020-10-14 at 7 23 06 PM
 I suggest we merge this PR as it is so we can distribute the failing tests among the team and fix them quickly.


cli/src/chain_spec/mod.rs
```diff
@@ -188,3 +203,114 @@ pub fn config_genesis(network_keys: NetworkKeys) -> GenesisConfig {
 		}),
 	}
 }
+
+#[cfg(test)]
+pub(crate) mod tests {
```
KarishmaBothara, line 208: these are new tests introduced?

aliXsed: Yes. However more than the tests, the intention for bringing them from Plug was the helper functions in this segment which are used in some other cli tests.

runtime/Cargo.toml
```diff
@@ -11,6 +11,8 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dev-dependencies]
 cennznet-cli = { path = "../cli", default-features = false }
+sp-keyring = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "spike/2.0" }
```
jordy25519, line 14: do we need keyring here? since we use it in runtime/tests/common/keyring

cli/src/chain_spec/mod.rs
```diff
@@ -71,8 +73,21 @@ pub struct NetworkKeys {
 	pub root_key: AccountId,
 }
 
+/// Node `ChainSpec` extensions.
+///
+/// Additional parameters for some Substrate core modules,
+/// customizable from the chain spec.
+#[derive(Default, Clone, Serialize, Deserialize, ChainSpecExtension)]
+#[serde(rename_all = "camelCase")]
+pub struct Extensions {
```
jordy25519, line 82: by reintroducing this you will need to update the genesis.jsons with these fields `fork_blocks` and `bad_blocks`.
Can test this by trying to run a node with `--chain=<network>.raw.json`

jordy25519 (changes requested):
lgtm overall!

I'd prefer we remove the contract related tests for now as the APIs are bound to have 
changed or will change, we don't want to pay for these tests while they aren't adding any value.

Main change needed is to fix the genesis json files

cli/src/chain_spec/mod.rs
aliXsed: I had to reintroduce these to make the tests build. The unit struct as the second template parameter for GenericChainSpec has the issue of not implementing ChainSpecExtension which is needed in some client service functions.
I will fix the genesis files and test them.

runtime/Cargo.toml
aliXsed: Yes correct.

cli/src/chain_spec/mod.rs
aliXsed: @holygits Without any change to the genesis file, I could connect and sync with Azalea.

jordy25519 (approved):

PR #296, opened 2020-10-19
Chore/config runtime fee

jordy25519:
Reconfigure fee scaling for CENNZnet using new WeightToPolynomial trait.

Since weights are now generated correctly I've removed the range translation thing we did before and gone for a simple scale down.

### DIFF
--- a/crml/transaction-payment/src/lib.rs
+++ b/crml/transaction-payment/src/lib.rs
@@ -477,9 +477,9 @@ where
 
 #[derive(Debug, Encode, Decode, Clone, Eq, PartialEq)]
 /// The variable parts of a transaction's fees
-/// It does not store the `base_fee` as it is a runtime constant
 /// The `tip` is also excluded as it is known by the transactor/caller.
 pub struct FeeParts<Balance: Saturating> {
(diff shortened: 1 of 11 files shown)


crml/transaction-payment/src/lib.rs
```diff
@@ -438,14 +438,79 @@ where
 		}
 	}
 
+	/// Introspectable version of `compute_fee_raw`.
+	/// Used to inspect fee composition in runtime integration tests, `tip` is ignored.
+	pub fn compute_fee_parts(len: u32, info: &DispatchInfoOf<T::Call>) -> FeeParts<BalanceOf<T>>
```
aliXsed, line 443: Would #[cfg(test)] be needed here to silent a compiler warning?

aliXsed (approved):
LGTM, just left a comment.

KarishmaBothara (approved):

crml/transaction-payment/src/lib.rs
jordy25519: Couldn't get it to work for some reason, will have another go in the next PR

PR #298, opened 2020-10-20
Bonding is enabled by default but can be blocked by root

aliXsed:
Keeping bonding enabled by default is necessary for the tests. However, the storage variable can be used to toggle the blocking.

Note: the PR also contains an incidental fix for the version of the dependency substrate-wasm-builder-runner

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/cennznet/cennznet"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dev-dependencies]
+crml-staking = { path = "../crml/staking", default-features = false}
 cennznet-cli = { path = "../cli", default-features = false }
 sp-keyring = { git = "https://github.com/plugblockchain/plug-blockchain", default-features = false, branch = "spike/2.0" }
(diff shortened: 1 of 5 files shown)


jordy25519 (commented):
Lgtm but can we get some tests for when BlockBonding is set?

Root only would give some more flexibility but I think we can work around it with a batch tx to toggle on/off with a bond operation in between in the same block (if we needed to bond a validator in the interim for some reason)

KarishmaBothara (approved):

jordy25519 (approved):

runtime/Cargo.toml
```diff
@@ -80,7 +80,7 @@ crml-staking = { path = "../crml/staking", default-features = false}
 crml-staking-reward-curve = { path = "../crml/staking/reward-curve", default-features = false}
 
 [build-dependencies]
-wasm-builder-runner = { version = "1.0.5", package = "substrate-wasm-builder-runner" }
+wasm-builder-runner = { package = "substrate-wasm-builder-runner", git = "https://github.com/plugblockchain/plug-blockchain", branch = "spike/2.0" }
```
jordy25519, line 83: 🏅

PR #304, opened 2020-10-27
Add rewards module

jordy25519:
Adds a new module responsible for staker reward payouts

### DIFF
--- a/crml/rewards/src/lib.rs
+++ b/crml/rewards/src/lib.rs
@@ -35,9 +35,6 @@ use sp_runtime::{
 };
 use sp_std::{collections::vec_deque::VecDeque, prelude::*};
 
-/// The number of historical eras for which tx fee payout info should be retained.
-const HISTORICAL_PAYOUT_ERAS: usize = 7;
-
 /// A balance amount in the reward currency
 type BalanceOf<T> = <<T as Trait>::CurrencyToReward as Currency<<T as system::Trait>::AccountId>>::Balance;
 /// A pending increase to total issuance of the reward currency
@@ -54,6 +51,8 @@ pub trait Trait: frame_system::Trait {
 	type CurrencyToReward: Currency<Self::AccountId>;
 	/// The treasury account for payouts
 	type TreasuryModuleId: Get<ModuleId>;
+	/// The number of historical eras for which tx fee payout info should be retained.
+	type HistoricalPayoutEras: Get<u16>;
 }
 
 decl_event!(
@@ -173,11 +172,11 @@ impl<T: Trait> Module<T> {
 		TransactionFeePot::<T>::mutate(|acc| *acc = acc.saturating_add(amount));
 	}
 
-	/// Note a fee payout for future calculations Retaining only the latest `HISTORICAL_PAYOUT_ERAS`
+	/// Note a fee payout for future calculations Retaining only the latest `T::HistoricalPayoutEras::get()`
 	fn note_fee_payout(amount: BalanceOf<T>) {
 		let mut history = TransactionFeePotHistory::<T>::get();
 		history.push_front(amount);
-		history.truncate(HISTORICAL_PAYOUT_ERAS); // truncate the oldest
+		history.truncate(T::HistoricalPayoutEras::get() as usize); // truncate the oldest
 		TransactionFeePotHistory::<T>::put(history);
 	}
 
@@ -305,14 +304,13 @@ mod tests {
 
 	parameter_types! {
 		pub const TreasuryModuleId: ModuleId = ModuleId(*b"py/trsry");
+		pub const HistoricalPayoutEras: u16 = 7;
 	}
 	impl Trait for TestRuntime {
-		/// The system event type
 		type Event = ();
-		/// The reward currency system (total issuance, account balance, etc.) for payouts.
 		type CurrencyToReward = prml_generic_asset::SpendingAssetCurrency<Self>;
-		/// The treasury account for payouts
 		type TreasuryModuleId = TreasuryModuleId;
+		type HistoricalPayoutEras = HistoricalPayoutEras;
 	}
 
 	// Provides configurable mock genesis storage data.
@@ -330,8 +328,8 @@ mod tests {
 
 	/// Alias for the mocked module under test
 	type Rewards = Module<TestRuntime>;
-	/// Alias for the currency to payout rewards with
-	type RewardCurrency = prml_generic_asset::SpendingAssetCurrency<TestRuntime>;
+	/// Alias for the reward currency in the module under test
+	type RewardCurrency = <TestRuntime as Trait>::CurrencyToReward;
 
 	/// Helper for creating the info required for validator reward payout
 	struct MockCommissionStakeInfo {
@@ -387,7 +385,7 @@ mod tests {
 	fn note_fee_payout_retains_n_latest() {
 		// note multiple fee payouts, it should keep only the latest n in state.
 		ExtBuilder::default().build().execute_with(|| {
-			let historical_payouts = [1_000_u64; HISTORICAL_PAYOUT_ERAS];
+			let historical_payouts = [1_000_u64; <TestRuntime as Trait>::HistoricalPayoutEras::get() as usize];
 			for payout in &historical_payouts {
 				Rewards::note_fee_payout(*payout);
 			}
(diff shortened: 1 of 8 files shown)


crml/rewards/src/lib.rs
```diff
@@ -0,0 +1,727 @@
...
+decl_storage! {
+	trait Store for Module<T: Trait> as Rewards {
+		/// Inflation rate % to apply on reward payouts, it may be negative
+		pub InflationRate get(fn inflation_rate): FixedI128 = FixedI128::saturating_from_integer(1);
+		/// Development fund % take for reward payouts, parts-per-billion
+		pub DevelopmentFundTake get(fn development_fund_take) config(): Perbill;
+		/// Accumulated transaction fees for reward payout
+		pub TransactionFeePot get(fn transaction_fee_pot): BalanceOf<T>;
+		/// Historic accumulated transaction fees on reward payout
+		pub TransactionFeePotHistory get(fn transaction_fee_pot_history): VecDeque<BalanceOf<T>>;
```
aliXsed, line 77: Is this going to grow infinitely?

jordy25519: no, only the most recent will be kept in state

crml/rewards/src/lib.rs
```diff
@@ -0,0 +1,727 @@
...
+		// This ensures only one storage update to total issuance will happen when dropped.
+		let mut total_payout_imbalance = <PositiveImbalanceOf<T>>::zero();
+
+		validator_commission_stake_map
+			.iter()
+			.flat_map(|(validator, validator_commission, stake_map)| {
+				Self::calculate_npos_payouts(&validator, *validator_commission, stake_map, total_payout_share)
+			})
+			.for_each(|(account, payout)| {
+				total_payout_imbalance.maybe_subsume(T::CurrencyToReward::deposit_into_existing(&account, payout).ok());
```
aliXsed, line 151: I suppose when a fee is paid for an extrinsic, we burn that and that's why we need to create an imbalance for the whole reward payout. Is that true?

crml/rewards/src/lib.rs
```diff
@@ -0,0 +1,727 @@
...
+use frame_system::{self as system, ensure_root};
+use sp_arithmetic::{FixedI128, FixedPointNumber, FixedPointOperand};
+use sp_runtime::{
+	traits::{AccountIdConversion, One, Saturating, Zero},
+	ModuleId, Perbill,
+};
+use sp_std::{collections::vec_deque::VecDeque, prelude::*};
+
+/// The number of historical eras for which tx fee payout info should be retained.
+const HISTORICAL_PAYOUT_ERAS: usize = 7;
```
aliXsed: How about having a runtime parameter for this? -> parameter_types!

crml/rewards/src/lib.rs
jordy25519: yes, it will reissue those fee tokens

crml/rewards/src/lib.rs
```diff
@@ -0,0 +1,727 @@
...
+		// Deduct development fund take %
+		let development_fund_payout = Self::development_fund_take() * total_payout;
+		let _ = T::CurrencyToReward::deposit_into_existing(
+			&T::TreasuryModuleId::get().into_account(),
+			development_fund_payout,
+		);
+		total_payout = total_payout.saturating_sub(development_fund_payout);
+
+		// Payout reward to validators and their nominators
+		let total_payout_share = total_payout / BalanceOf::<T>::from(validator_commission_stake_map.len() as u32);
```
aliXsed, line 138: This https://github.com/cennznet/cennznet/pull/304/files#diff-8b6330393ec7c1d63275edb851ba0d647bc896357706fa9377fe211ae033f0abR120-R139 
could be an individual function for calculating the payout share to each validator and its nominators.

crml/rewards/src/lib.rs
aliXsed, line 151: I guess it could be better that another individual function of StakerRewardPayment just returns "who should get how much" and then the actual deposit and then event generation is left to the Module. The more we can separate concerns, the better we can change our strategies in the future.

crml/rewards/src/lib.rs
```diff
@@ -0,0 +1,727 @@
...
+				Rewards::inflation_rate(),
+				FixedI128::saturating_from_rational(-1, 1_000)
+			)
+		});
+	}
+
+	#[test]
+	fn set_inflation_rate_bounds() {
+		ExtBuilder::default().build().execute_with(|| {
+			assert_err!(
```
aliXsed: assert_noop might be better if we don't expect any change in the storage.

crml/rewards/src/lib.rs
```diff
@@ -0,0 +1,727 @@
...
+			assert_ok!(Rewards::set_inflation_rate(
+				Origin::root(),
+				i64::max_value(),
+				i64::max_value()
+			));
+			assert_ok!(Rewards::set_inflation_rate(
+				Origin::root(),
+				i64::min_value(),
+				i64::min_value()
+			));
```
aliXsed, line 460: Is it worth testing the mix of min and max?

primitives/src/traits.rs
```diff
@@ -69,3 +76,20 @@ pub trait SimpleAssetSystem {
 	/// Get the default asset/currency ID in the system
 	fn default_asset_id() -> Self::AssetId;
 }
+
+/// Something which can perform reward payment to staked validators
+pub trait StakerRewardPayment {
```
aliXsed, line 81: There is a potential for "primitives" to become a place for holding random unrelated stuff.

primitives/src/types.rs
```diff
@@ -116,3 +117,28 @@ impl<AssetId: Copy, Balance: Copy> FeeExchange<AssetId, Balance> {
...
+	/// The stash account of the nominator in question.
+	pub who: AccountId,
+	/// Amount of funds exposed.
+	#[codec(compact)]
+	pub value: Balance,
+}
+
+/// A snapshot of the stake backing a single validator in the system.
+#[derive(PartialEq, Eq, PartialOrd, Ord, Clone, Encode, Decode, Default, RuntimeDebug)]
+pub struct Exposure<AccountId, Balance: HasCompact> {
```
aliXsed, line 135: The same here, the types that could be better located in the Staking pallet now moved to be next to other unrelated stuff. I suppose you need to put them here so you share them between rewards and staking pallets. However, I'm asking this question if rewards could be a separate unit inside the staking module, the same as reward_curve?

jordy25519: The calculations could be, the main reason for separating rewards and staking like this to isolate the reward related storage and config from the staking storage and config, it made sense to bring the calculations also.
\+ the staking module is on the big side already

primitives/src/traits.rs
jordy25519: that's what it's used for, we could rename it to something besides `primitives` to indicate that better?

primitives/src/types.rs
aliXsed: We could even separate the storage and config in the same way that we have done for Sylo within the same pallet. At the end of the day when a service like staking is complex, it's true that we should decompose it to smaller components but we don't necessarily need to then scatter those components from the only location they can have a meaningful place. If there was a different pallet or service that could run with rewards but without the staking module, it could make sense. But now I'm not sure.

aliXsed:
I have left a few comments regarding how best we can structure this change. As those comments are just suggestions, I'm going to approve this PR at this stage. I'm hoping the next PR for integrating this module with the rest of cennznet can show that there haven't been any behavioural changes as otherwise, I couldn't easily note and evaluate them through looking at this change by itslef.

aliXsed (approved):

primitives/src/types.rs
jordy25519: Is this suggestion about where the rust files/modules sit? e.g.
```
# now
crml/rewards
crml/staking
# after
crml/staking/
crml/staking/rewards/<reward crate>
```
if so then yes I should be able to make this change

jordy25519:
There are a few rough edges on the rewards API, it's not so much generic as it is decoupled. So there's definitely room to improve it, including the individual reward calculation could come in handy

primitives/src/types.rs
aliXsed: Yup correct

PR #306, opened 2020-10-29
Replace simple asset system with MultiCurrency

aliXsed:
The extended functionality of MultiCurrency in comparison with the simple asset interface will be useful in Cennzx benchmarking tests

### DIFF
--- a/crml/cennzx/src/lib.rs
+++ b/crml/cennzx/src/lib.rs
@@ -18,15 +18,18 @@
 //!
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use cennznet_primitives::traits::SimpleAssetSystem;
 use core::convert::TryFrom;
 use frame_support::{
-	decl_error, decl_event, decl_module, decl_storage, ensure, transactional, weights::Weight, Parameter,
-	StorageDoubleMap,
+	decl_error, decl_event, decl_module, decl_storage, ensure,
+	traits::{ExistenceRequirement, Imbalance},
+	transactional,
+	weights::Weight,
+	Parameter, StorageDoubleMap,
 };
 use frame_system::{ensure_root, ensure_signed};
+use prml_support::MultiCurrencyAccounting;
 use sp_runtime::{
-	traits::{AtLeast32BitUnsigned, Member, One, Saturating, Zero},
+	traits::{AtLeast32BitUnsigned, MaybeSerializeDeserialize, Member, One, Saturating, Zero},
 	DispatchError, DispatchResult, SaturatedConversion,
 };
 use sp_std::prelude::*;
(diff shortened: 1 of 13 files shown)


crml/cennzx/src/lib.rs
```diff
@@ -223,23 +238,23 @@ decl_module! {
 				Error::<T>::CannotAddLiquidityWithZero
 			);
 			ensure!(
-				T::AssetSystem::free_balance(core_asset_id, &from_account) >= core_amount,
+				T::MultiCurrency::free_balance(&from_account, Some(core_asset_id.clone())) >= core_amount,
```
jordy25519: nit: is asset_id copy? might not need all these clones

crml/cennzx/src/lib.rs
```diff
@@ -73,9 +76,21 @@ pub trait Trait: frame_system::Trait {
...
+	/// The opaque token type for an imbalance of a particular currency. This is returned by unbalanced operations
+	/// and must be dealt with. It may be dropped but cannot be cloned.
+	type PositiveImbalance: Imbalance<Self::Balance, Opposite = Self::NegativeImbalance>;
+	/// Something which can provide multi currency asset management
+	type MultiCurrency: MultiCurrencyAccounting<
+		AccountId = Self::AccountId,
+		Balance = Self::Balance,
+		CurrencyId = Self::AssetId,
+		NegativeImbalance = Self::NegativeImbalance,
+		PositiveImbalance = Self::PositiveImbalance,
```
jordy25519, line 92: is it possible to leave these imbalance types unspecified and not on the module trait?

jordy25519 (approved):
lgtm!
few small nits
hoping we don't need to specify the imbalances type on the module

crml/cennzx/src/lib.rs
aliXsed: Yes, it is. I will get rid of the clones.

crml/cennzx/src/lib.rs
aliXsed: I tried it first without the imbalance types but got an error that Imbalance was not implemented as required by the MultiCurrencyAccounting::NegativeImbalance bound. The error was suggesting that I add a "where" clause to the cennzx Trait which I can try as an alternative solution.

crml/cennzx/src/lib.rs
aliXsed: Done

crml/cennzx/src/lib.rs
aliXsed: The "where" clause approach seems to create more problems.So leaving it as it is for this PR and we can get back to it later.

PR #302, opened 2020-10-21
Configure initial staking rewards

jordy25519:
Split the rewards module addition into #304 
This will be the integration part of rewards->staking

- Removes the npos inflation curve from the staking module and moves validator reward payment and calculation to its own module
- Rewards provides a configurable inflation rate and tracks past transaction fees for dynamic inflation calculations in future.
- The rewards module also handles routing fees to the treasury
- `RewardDestination::Controller` is converted to `RewardDestination::Account(controller)` as this requires one less storage read later

In this version we still keep the notion of validator 'points' but don't factor them into reward payouts

TODO:
- [x] Fix tests, **after this PR all tests should be passing**

PR #309, opened 2020-11-03
Used substrates staking weights

KarishmaBothara:
Have used weights from [this](https://github.com/paritytech/substrate/pull/7007/files) PR 
Also there are couple of places which used parameter from module method to calculate weight, and we don't use those parameter, so I have calculated weight without using those param..
for example - **fn reap_stash(s: u32, ) -> Weight {**
uses s which comes from 
```
#[weight = T::WeightInfo::reap_stash(*num_slashing_spans)]
		fn reap_stash(_origin, stash: T::AccountId, num_slashing_spans: u32) {
			ensure!(T::Currency::total_balance(&stash).is_zero(), Error::<T>::FundedTarget);
			Self::kill_stash(&stash, num_slashing_spans)?;
			T::Currency::remove_lock(STAKING_ID, &stash);
		}
```
But we don't use this field num_slashing_spans in our code, so have ignored it

Similarly for withdraw_unbonded_kill


crml/staking/src/lib.rs
```diff
@@ -290,81 +291,118 @@ pub trait WeightInfo {
...
 	fn set_minimum_bond() -> Weight;
-	fn set_invulnerables() -> Weight;
+	fn set_invulnerables(v: u32) -> Weight;
 	fn force_unstake() -> Weight;
 	fn force_new_era_always() -> Weight;
-	fn cancel_deferred_slash() -> Weight;
+	fn cancel_deferred_slash(s: u32) -> Weight;
 }
 
 impl WeightInfo for () {
```
aliXsed: If you just remove this implementation here, then you don't need to copy paste the default implementation from Plug to here

jordy25519: Yup please use `mod default_weights` and update the implementation there. this one can be removed

jordy25519:
lgtm, would like to merge after #302 however

jordy25519 (approved):

crml/staking/src/lib.rs
```diff
@@ -1364,15 +1283,15 @@ decl_module! {
 		}
 
 		/// Set the minimum bond amount.
-		#[weight = T::WeightInfo::set_minimum_bond()]
+		#[weight = 0]
```
jordy25519: can you make this `DispatchClass::Operational` and give it some constant weight like `1_000_000`

crml/staking/src/lib.rs
```diff
@@ -1382,7 +1301,7 @@ decl_module! {
 		}
 
 		/// Force a current staker to become completely unstaked, immediately.
-		#[weight = T::WeightInfo::force_unstake()]
+		#[weight = T::WeightInfo::force_unstake(1_u32)]
```
jordy25519, line 1304: why 1?

crml/staking/src/lib.rs
```diff
@@ -1154,7 +1073,7 @@ decl_module! {
 		/// - Writes: Bonded, Slashing Spans (if S > 0), Ledger, Payee, Validators, Nominators, Stash Account, Locks
 		/// - Writes Each: SpanSlash * S
 		/// # </weight>
-		#[weight = T::WeightInfo::reap_stash()]
+		#[weight = T::WeightInfo::reap_stash(1_u32)]
```
jordy25519, line 1076: why 1?

crml/staking/src/lib.rs
KarishmaBothara: https://github.com/plugblockchain/plug-blockchain/blob/origin/spike/2.0.0/frame/staking/src/lib.rs#L1862
If you check the force_unstake
#[weight = T::WeightInfo::force_unstake(*num_slashing_spans)]
		fn force_unstake(origin, stash: T::AccountId, num_slashing_spans: u32) {

It uses num_slashing_spans which we don't have so I used 1

crml/staking/src/lib.rs
KarishmaBothara: Similarly.. here
https://github.com/plugblockchain/plug-blockchain/blob/origin/spike/2.0.0/frame/staking/src/lib.rs#L2038

crml/staking/src/lib.rs
KarishmaBothara: https://github.com/plugblockchain/plug-blockchain/blob/origin/spike/2.0.0/frame/staking/src/default_weights.rs#L102
the parameter is getting multiplied so I kept it 1

jordy25519:
squash and sign commits pls, then lgtm

PR #310, opened 2020-11-03
Add cennzx benchmark test and adjust cennzx weights

aliXsed:
Add the benchmarks, run them, use the result to set the weights. This also fixes the cennznet benchmarking facility.


crml/staking/Cargo.toml
```diff
@@ -48,3 +48,4 @@ std = [
 	"pallet-authorship/std",
 	"pallet-session/std",
 ]
+runtime-benchmarks = []
```
jordy25519: Don't need these empty features `runtime-benchmarks = []`

crml/cennzx/src/benchmarking.rs
```diff
@@ -0,0 +1,160 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+* You may obtain a copy of the License at the root of this project source code,
+* or at:
+*     https://centrality.ai/licenses/gplv3.txt
+*     https://centrality.ai/licenses/lgplv3.txt
+*/
+
+//! Attestation benchmarking.
```
jordy25519: nit: `Attestation benchmarking.`

crml/cennzx/src/benchmarking.rs
```diff
@@ -0,0 +1,160 @@
...
+
+benchmarks! {
+	_{ }
+
+	buy_asset {
+		let investor: T::AccountId = whitelisted_caller();
+		let buyer: T::AccountId = account("buyer", 0, 0);
+
+		let core_asset_id = <Cennzx<T>>::core_asset_id();
+		let asset_a: T::AssetId = 2u32.into();
```
jordy25519: nit: is it possible to use const for the asset IDs?

runtime/src/weights/crml_cennzx.rs
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 2.0.0
+
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+
+use frame_support::weights::{constants::RocksDbWeight as DbWeight, Weight};
+
+pub struct WeightInfo;
```
jordy25519, line 23: nit: I think upstream is moving away from having the weights defined in two places.
So probably don't need this one `type WeightInfo = ();` will be correct

crml/cennzx/src/benchmarking.rs
KarishmaBothara: CENNZX benchmarking

crml/cennzx/src/benchmarking.rs
```diff
@@ -0,0 +1,160 @@
...
+
+	sell_asset {
+		let investor: T::AccountId = whitelisted_caller();
+		let seller: T::AccountId = account("seller", 0, 0);
+
+		let core_asset_id = <Cennzx<T>>::core_asset_id();
+		let asset_a: T::AssetId = 2u32.into();
+		let asset_b: T::AssetId = 3u32.into();
+
+		let _ = T::MultiCurrency::deposit_creating(&investor, Some(core_asset_id), 1000u32.into());
```
jordy25519, line 62: wondering if using some larger values would sway the benchmarks in any meaningful way?
these numbers are a very small amount compared to real use
especially add/remove liquidity could be > 1,000,000.0000 easily

jordy25519 (approved):
looks great.
Few smalll nits and one concern about order of magnitude in the benchmarks

KarishmaBothara (approved):
👍

crml/staking/Cargo.toml
aliXsed: Removed them.

crml/cennzx/src/benchmarking.rs
aliXsed: Fixed

crml/cennzx/src/benchmarking.rs
aliXsed: The numbers, though small, are converted to and held in large number containers. Thus my understanding is that shouldn't make a difference.

crml/cennzx/src/benchmarking.rs
aliXsed: Used const

crml/cennzx/src/benchmarking.rs
aliXsed: Done

runtime/src/weights/crml_cennzx.rs
aliXsed: The weights in one place are supposed to be used for integration tests. In theory, you may want to specify different weights for tests. So perhaps we give this a little bit more thought in the future.

PR #315, opened 2020-11-09
Add docker-compose

jordy25519:
A little docker compose config useful for testing a local Nikau

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,103 @@
+version: '3'
+services:
+  node-0:
+    image: cennznet/cennznet:1.1.1
+    volumes:
+    - ./data/node-0:/mnt/data
+    command:
+      - --chain=/cennznet/genesis/nikau.raw.json
+      - --validator
+      - --alice
+      - --base-path=/mnt/data
+      - --name=alice-${HOSTNAME:-localhost}-0
+      - --unsafe-ws-external
+      - --unsafe-rpc-external
+      - --rpc-cors=all
+      # bootnode 1 (running 1.2.0-rc1)
+      - --bootnodes=/dns4/node-5/tcp/30333/p2p/12D3KooWEyoppNCUx8Yx66oV9fJnriXwCcXwDDUA2kj6vnc6iDEp
+      # bootnode 2 (running 1.1.1)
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,103 @@
...
+    - ./data/node-0:/mnt/data
+    command:
+      - --chain=/cennznet/genesis/nikau.raw.json
+      - --validator
+      - --alice
+      - --base-path=/mnt/data
+      - --name=alice-${HOSTNAME:-localhost}-0
+      - --unsafe-ws-external
+      - --unsafe-rpc-external
+      - --rpc-cors=all
```
KarishmaBothara, line 15: just a doubt - why we don't need **node-key** here, we have it for all other service?

jordy25519: It sets the node's network level public key. It's mostly important for hardcoding bootnodes here

KarishmaBothara (approved):

docker-compose.yml
```diff
@@ -0,0 +1,103 @@
...
+  node-0:
+    image: cennznet/cennznet:1.1.1
+    volumes:
+    - ./data/node-0:/mnt/data
+    command:
+      - --chain=/cennznet/genesis/nikau.raw.json
+      - --validator
+      - --alice
+      - --base-path=/mnt/data
+      - --name=alice-${HOSTNAME:-localhost}-0
```
aliXsed, line 12: Do we need -0, -1 and etc at the end of the names?

docker-compose.yml
```diff
@@ -0,0 +1,103 @@
+version: '3'
+services:
+  node-0:
+    image: cennznet/cennznet:1.1.1
```
aliXsed, line 4: Is it temporary that most of them will run the previous version?

docker-compose.yml
jordy25519: Yeah it's the docker container names, it could be named something else if we wanted

docker-compose.yml
jordy25519: Yeah, I'm thinking we'll update this file with each release. For now I was using it to test upgrades from 1.1.1 -> 1.2.0

aliXsed (approved):

jordy25519:
included in #320

PR #318, opened 2020-11-10
Benchmark Sylo::e2ee register_device

aliXsed:
This is 1/3 of Sylo E2ee benchmark. To create the worst-case scenario for register_device, the sender should be a member of several groups for which the new device should be declared. However, when benchmarking e2ee there is no access to the private structures and functions of the groups. The access is necessary for creating some groups when setting up the benchmark. This is why I needed to add some pub functions guarded by the feature runtime-benchmarks to that module.

### DIFF
--- a/crml/sylo/src/e2ee/benchmarking.rs
+++ b/crml/sylo/src/e2ee/benchmarking.rs
@@ -33,6 +33,7 @@ benchmarks! {
 	_{ }
 
 	register_device {
+		let p in 0 .. MAX_PKBS as u32;
 		let sender: T::AccountId = whitelisted_caller();
 		let device_id: DeviceId = 11;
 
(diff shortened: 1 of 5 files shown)


crml/sylo/src/e2ee/benchmarking.rs
```diff
@@ -0,0 +1,95 @@
...
+		let group_1 = Group::<T::AccountId, T::Hash>::new(
+			group_id_1,
+			Vec::<Member<T::AccountId>>::from([admin, member]),
+			Vec::new(),
+			meta,
+		);
+		<SyloGroups<T>>::insert(&sender, &group_id_1, group_1);
+
+		let pre_key_bundle0 = PreKeyBundle::from(*b"prekeybundle0");
+		let pre_key_bundle1 = PreKeyBundle::from(*b"prekeybundle1");
```
jordy25519: Sylo registers something around 8-12 pkbs when this is called, could be worth adding a few more or adjusting the weight function as such

pkbs from the app are shas256 hashes / 64 bytes in length

jordy25519 (approved):
lgtm, we have abit more knowledge about the real use of this function so could benchmark for that.

crml/sylo/src/e2ee/benchmarking.rs
aliXsed: I fixed it to take the number of bundled pre-keys into account.

PR #319, opened 2020-11-10
Benchmark Sylo replenish_pkbs and withdraw_pkbs

aliXsed:
Benchmark Sylo replenish_pkbs and withdraw_pkbs. Also, Make some reusable helper function for benchmarking Sylo e2ee. Place a cap for the list of wanted pkbs of withdraw_pkbs.

### DIFF
--- a/crml/sylo/src/e2ee/benchmarking.rs
+++ b/crml/sylo/src/e2ee/benchmarking.rs
@@ -33,6 +33,7 @@ benchmarks! {
 	_{ }
 
 	register_device {
+		let p in 0 .. MAX_PKBS as u32;
 		let sender: T::AccountId = whitelisted_caller();
 		let device_id: DeviceId = 11;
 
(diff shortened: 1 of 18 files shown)


crml/sylo/src/e2ee/benchmarking.rs
```diff
@@ -36,51 +82,25 @@ benchmarks! {
...
 	verify {
 		assert!(<SyloDevice<T>>::devices(sender).contains(&device_id));
 	}
+
+	replenish_pkbs {
+		let p in 0 .. MAX_PKBS as u32;
+		let sender: T::AccountId = whitelisted_caller();
+		let device_id: DeviceId = 11;
+		setup_groups::<T>(sender.clone());
+		let pre_key_bundles = create_pre_key_bundles::<T>(0);
```
jordy25519: Couples qs:
1) why declare `let pre_key_bundles = create_pre_key_bundles::<T>(0);` with 0?
2) ~also how does this allow two declarations of `let pre_key_bundles` seems like something idiosyncratic is happening? or either this doesn't compile but the CI isn't checking benchmark features atm~ nevermind didnt' realise you can declare vars with same name in same scope!
Would be good to comment what is happening here as it seems unusual

aliXsed: The function replenish_pkbs extends the current pkbs for a user-device. In my setup, I need to first register the device with some or as it's the case here 0 pkbs and then benchmark replenish_pkbs extending it. As I wanted to allow the whole range of an extension for the worst-case scenario I used 0 as the initial number.

aliXsed: I added a short comment there.

jordy25519 (approved):

PR #320, opened 2020-11-11
Prerelease/1.2.0-rc2

jordy25519:
Includes client stability fixes for rpc, grandpa, and logging targets.
Added the docker-compose here with update

Also has some preliminary Sylo benchmarked weights (this would normally require bumping runtime version but it's fine as there's no Sylo traffic on Nikau atm). Upgrading the nikau nodes with this image will work without issue

### DIFF
--- a/cli/src/rpc.rs
+++ b/cli/src/rpc.rs
@@ -94,7 +94,7 @@ pub struct FullDeps<C, P, SC, B> {
 pub type IoHandler = jsonrpc_core::IoHandler<sc_rpc::Metadata>;
 
 /// Instantiate all Full RPC extensions.
-pub fn create_full<C, P, SC, B>(deps: FullDeps<C, P, SC, B>) -> jsonrpc_core::IoHandler<sc_rpc_api::Metadata>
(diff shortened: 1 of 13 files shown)


cli/src/rpc.rs
```diff
@@ -94,7 +94,7 @@ pub struct FullDeps<C, P, SC, B> {
 pub type IoHandler = jsonrpc_core::IoHandler<sc_rpc::Metadata>;
 
 /// Instantiate all Full RPC extensions.
-pub fn create_full<C, P, SC, B>(deps: FullDeps<C, P, SC, B>) -> jsonrpc_core::IoHandler<sc_rpc_api::Metadata>
+pub fn create_full<C, P, SC, B>(deps: FullDeps<C, P, SC, B>) -> IoHandler
```
KarishmaBothara, line 97: not sure, what this is for... rest lgtm

KarishmaBothara (approved):

cli/src/rpc.rs
jordy25519: little drive by fix: I noticed the type was already aliased as `pub type IoHandler = jsonrpc_core::IoHandler<sc_rpc::Metadata>;` so used it

PR #325, opened 2020-11-13
Feature/sylo groups benchmark

aliXsed:
Benchmark create_group and update_member alongside the newly added upsert_group_meta after supposing a variable capped size for meta.

### DIFF
--- a/crml/sylo/src/groups/default_weights.rs
+++ b/crml/sylo/src/groups/default_weights.rs
@@ -21,8 +21,13 @@
 use frame_support::weights::{constants::RocksDbWeight as DbWeight, Weight};
 
 impl crate::groups::WeightInfo for () {
-	fn create_group() -> Weight {
-		0
+	fn create_group(i: usize) -> Weight {
+		(81_563_000 as Weight)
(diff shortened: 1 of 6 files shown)


crml/sylo/src/groups/default_weights.rs
```diff
@@ -21,29 +21,37 @@
 use frame_support::weights::{constants::RocksDbWeight as DbWeight, Weight};
 
 impl crate::groups::WeightInfo for () {
-	fn create_group(i: usize) -> Weight {
-		(45_728_000 as Weight)
-			.saturating_add((59_625_000 as Weight).saturating_mul(i as Weight))
+	fn create_group(i: usize, m: usize) -> Weight {
+		(0 as Weight)
```
KarishmaBothara, line 25: why it changed from 45_728_000 as weight to 0 as weight?

jordy25519 (approved):

crml/sylo/src/groups/default_weights.rs
aliXsed: Benchmark changed it. The latest results are affected by having the maximum number of meta for the group.

PR #326, opened 2020-11-16
Feature/sylo groups benchmark

aliXsed:
Benchmark revoke_invites. Fix the test name for create_invites. Improve the benchmark for leave_group.

### DIFF
--- a/crml/sylo/src/groups/default_weights.rs
+++ b/crml/sylo/src/groups/default_weights.rs
@@ -56,7 +56,9 @@ impl crate::groups::WeightInfo for () {
 	fn accept_invite() -> Weight {
 		0
 	}
-	fn revoke_invites() -> Weight {
-		0
+	fn revoke_invites(_i: usize) -> Weight {
+		(435_834_000 as Weight)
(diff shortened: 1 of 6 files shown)


crml/sylo/src/groups/default_weights.rs
```diff
@@ -56,7 +56,9 @@ impl crate::groups::WeightInfo for () {
 	fn accept_invite() -> Weight {
 		0
 	}
-	fn revoke_invites() -> Weight {
-		0
+	fn revoke_invites(_i: usize) -> Weight {
```
jordy25519, line 59: was it intentional to exclude `_i` here?

jordy25519 (approved):

crml/sylo/src/groups/default_weights.rs
aliXsed: Yes, in theory, the length of the revoked list should affect the performance. In practical and within our limits (the max invites and etc), it didn't depend on it. However, as the benchmark remains parametric, we should keep this.

PR #327, opened 2020-11-16
Benchmark accept_invite and set the corresponding weight function

aliXsed:
Because the key generation needs certain rand_core features which are not available during a wasm-run, I had to resort to predefined key and signature taken from the test mode. However invitee_id is different in the test mode and thus using the same key and signature can cause InvitationSignatureRejected in the runtime mode for benchmarking. For this reason, accept_invite is modified not to emit InvitationSignatureRejected during the benchmark.


KarishmaBothara (approved):
lgtm as a non rust expert.

PR #338, opened 2020-11-21
Split the reward payouts over several blocks of an era when needed

### LINKED ISSUE #337: Amortised rewards payouts
Schedule staking reward payouts over multiple future blocks to amortise the cost.

aliXsed:
This is using a split threshold as an indicator to decide whether it should try to split the payouts or not. The idea is we should serve as many payouts as possible in the subsequent blocks right after the end of an era. We will do our best to process most of the payouts immediately and we will never let them be drawn out beyond the duration of the era. In theory, if the number of payouts is very very large, we wouldn't be able to process just up to the split threshold. This is because we should finish them before the end of an era. So it's possible that in some scenarios we would still have to process more payouts than the threshold is suggesting in a block. However, practically we don't expect this to happen often or any time soon for the current scales of most of the blockchains. But even if it happens, it means the processing affects the block duration sensibly which is inevitable. 
Closes #337


crml/staking/src/mock.rs
```diff
@@ -287,9 +295,24 @@ impl pallet_timestamp::Trait for Test {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub const HistoricalPayoutEras: u16 = 7;
```
jordy25519, line 300: is this impl required?

crml/staking/src/rewards/mod.rs
```diff
@@ -76,6 +87,10 @@ decl_storage! {
 		pub TransactionFeePot get(fn transaction_fee_pot): BalanceOf<T>;
 		/// Historic accumulated transaction fees on reward payout
 		pub TransactionFeePotHistory get(fn transaction_fee_pot_history): VecDeque<BalanceOf<T>>;
+		/// Remained reward amount in an era
+		pub EraRemainedRewardAmount get(fn era_total_payout): BalanceOf<T>;
```
jordy25519: 'Remained' -> 'Remaining' in other places also

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
 
-		// Any unallocated reward amount can go to the development fund
-		let remainder = total_payout.saturating_sub(total_payout_imbalance.peek());
-		total_payout_imbalance.subsume(
-			T::CurrencyToReward::deposit_into_existing(&T::TreasuryModuleId::get().into_account(), remainder)
-				.unwrap_or_else(|_| PositiveImbalanceOf::<T>::zero()),
-		);
+		EraRemainedRewardAmount::<T>::put(total_payout.saturating_sub(total_payout_imbalance.peek()));
+	}
+
```
jordy25519, line 177: doc comment pls.

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
+			for _ in 1..quota {
+				if let Some((a, m)) = p.pop() {
+					total_payout_imbalance.maybe_subsume(T::CurrencyToReward::deposit_into_existing(&a, m).ok());
+				}
+			}
+
+			let remainder = EraRemainedRewardAmount::<T>::get().saturating_sub(total_payout_imbalance.peek());
 
-		Self::deposit_event(RawEvent::RewardPayout(total_payout_imbalance.peek(), remainder));
+			if p.len() > 0 {
```
```suggestion
			if !p.is_empty() {
```

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
-			T::CurrencyToReward::deposit_into_existing(&T::TreasuryModuleId::get().into_account(), remainder)
-				.unwrap_or_else(|_| PositiveImbalanceOf::<T>::zero()),
-		);
+		EraRemainedRewardAmount::<T>::put(total_payout.saturating_sub(total_payout_imbalance.peek()));
+	}
+
+	fn process_reward_payouts(remained_blocks: Self::BlockNumber) -> Weight {
+		let remained_payouts = EraRemainedPayouts::<T>::get().len();
+		let quota = Self::calculate_payout_quota(remained_payouts, remained_blocks);
+		if quota == 0 {
```
```suggestion
		if quota.is_zero() {
```

runtime/src/lib.rs
```diff
@@ -474,12 +476,15 @@ impl pallet_treasury::Trait for Runtime {
 
 parameter_types! {
 	pub const HistoricalPayoutEras: u16 = 7;
+	pub const PayoutSplitThreshold: usize = 1000;
```
jordy25519: nit: would prefer this to be an ordinary uint like`u32` (I don't trust API usize support :D)

crml/staking/src/rewards/mod.rs
```diff
@@ -230,6 +278,23 @@ impl<T: Trait> Module<T> {
 		));
 		(*payouts).to_vec()
 	}
+
+	/// Return the number of reward payouts that need to be processed in the current block.
+	/// The result is dependant on the number of the current era's remained payouts and the number
```
jordy25519: nit: dependant -> dependent

crml/staking/src/rewards/mod.rs
```diff
@@ -230,6 +278,23 @@ impl<T: Trait> Module<T> {
...
+	/// of remained blocks before a new era.
+	fn calculate_payout_quota(remained_payouts: usize, remained_blocks: T::BlockNumber) -> usize {
+		let payout_split_threshold = T::PayoutSplitThreshold::get();
+		if remained_payouts <= payout_split_threshold || remained_blocks == Zero::zero() {
```
```suggestion
		if remaining_payouts <= payout_split_threshold || remaining_blocks.is_zero() {
```
could also do remaning_blocks zero check first to avoid storage read of `payout_split_threshold`

crml/staking/src/rewards/mod.rs
```diff
@@ -230,6 +278,23 @@ impl<T: Trait> Module<T> {
...
+			return remained_payouts;
+		}
+		let remained_payouts = <T::BlockNumber as UniqueSaturatedFrom<usize>>::unique_saturated_from(remained_payouts);
+		let min_payouts = remained_payouts / (remained_blocks + One::one());
```
jordy25519: why `+ One::one()`?

crml/staking/src/rewards/mod.rs
```diff
@@ -230,6 +278,23 @@ impl<T: Trait> Module<T> {
...
+		let min_payouts = <T::BlockNumber as UniqueSaturatedInto<usize>>::unique_saturated_into(min_payouts);
+		if min_payouts < payout_split_threshold {
+			return payout_split_threshold;
+		}
+		min_payouts
```
```suggestion
		min_payouts.max(payout_split_threshold)
```

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
+			}
+
+			let remainder = EraRemainedRewardAmount::<T>::get().saturating_sub(total_payout_imbalance.peek());
 
-		Self::deposit_event(RawEvent::RewardPayout(total_payout_imbalance.peek(), remainder));
+			if p.len() > 0 {
+				EraRemainedRewardAmount::<T>::put(remainder);
+				return;
+			}
+			EraRemainedRewardAmount::<T>::put(BalanceOf::<T>::zero());
```
```suggestion
			EraRemainedRewardAmount::<T>::kill();
```

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
+
+			// Any unallocated reward amount can go to the development fund
+			total_payout_imbalance.subsume(
+				T::CurrencyToReward::deposit_into_existing(&T::TreasuryModuleId::get().into_account(), remainder)
+					.unwrap_or_else(|_| PositiveImbalanceOf::<T>::zero()),
+			);
+			Self::deposit_event(RawEvent::RewardPayout(total_payout_imbalance.peek(), remainder));
```
jordy25519: we should log the event on line 200 when we early exit also

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
+	fn process_reward_payouts(remained_blocks: Self::BlockNumber) -> Weight {
+		let remained_payouts = EraRemainedPayouts::<T>::get().len();
+		let quota = Self::calculate_payout_quota(remained_payouts, remained_blocks);
+		if quota == 0 {
+			return T::WeightInfo::process_zero_payouts();
+		}
+
+		let weight = T::WeightInfo::process_reward_payouts(quota);
+
+		EraRemainedPayouts::<T>::mutate(|p| {
```
jordy25519: this will decode the entire EraRemainedPayouts storage,
not sure how much affect it will have though...
other option I can think of is to pre-bucket the payouts at end of an era, then you only need to read 1 bucket per payout

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +167,49 @@ where
...
-		let remainder = total_payout.saturating_sub(total_payout_imbalance.peek());
-		total_payout_imbalance.subsume(
-			T::CurrencyToReward::deposit_into_existing(&T::TreasuryModuleId::get().into_account(), remainder)
-				.unwrap_or_else(|_| PositiveImbalanceOf::<T>::zero()),
-		);
+		EraRemainedRewardAmount::<T>::put(total_payout.saturating_sub(total_payout_imbalance.peek()));
+	}
+
+	fn process_reward_payouts(remained_blocks: Self::BlockNumber) -> Weight {
+		let remained_payouts = EraRemainedPayouts::<T>::get().len();
```
jordy25519: more efficient
```suggestion
		let remained_payouts = EraRemainedPayouts::<T>::decode_len();
```

jordy25519 (commented):
Direction looks good!
Some design things:
- On a forced era we'd want to schedule all remaining payouts to happen
- Would prefer to amortise reward payouts over the first few blocks of an era vs. the whole era because dripping the rewards out is unfair to the last validator/nominators and will pollute too many blocks.
- liberal early exits for process_reward_payout are important since it can run every block...

I think I can see why the scheduler approach is useful since it takes care of a lot of the edge cases

aliXsed:
> Direction looks good!
> Some design things:
> 
> * On a forced era we'd want to schedule all remaining payouts to happen
> * Would prefer to amortise reward payouts over the first few blocks of an era vs. the whole era because dripping the rewards out is unfair to the last validator/nominators and will pollute too many blocks.
> * liberal early exits for process_reward_payout are important since it can run every block...
> 
> I think I can see why the scheduler approach is useful since it takes care of a lot of the edge cases

@holygits regarding your design points:
* On a forced, all remaining payouts will be in front of all the other payments and will be processed first. This will help the cost of force new_era to be deterministic.
* We are amortising the reward payouts over the first few blocks because we go with the maximum sustainable peek of service unless for super large payments for which we need to even process more and go faster.
* There is a liberal early exit in process_reward_payouts for which I have even calculated the weight individually.

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs
aliXsed: Addressed

jordy25519:
~We should settle the reward payout in the same era when forced because stakes can change and the accounting will be off. It's also simpler to understand (from an outside perspective watching payments happen e.g. on a block explorer)~

---

offline discussion the current approach is good, need extra events to distinguish start / end of era payouts

crml/staking/src/rewards/mod.rs
aliXsed: Done

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs
aliXsed: Corrected

aliXsed:
> We should settle the reward payout in the same era when forced because stakes can change and the accounting will be off. It's also simpler to understand (from an outside perspective watching payments happen e.g. on a block explorer)

@holygits Alright then. As discussed there is a variable cost involved when an era is forced because it will force all payments to happen suddenly but if it's more important to not allow reward payments to extend beyond one era, then I guess I can do that by a simple change.

crml/staking/src/rewards/mod.rs
aliXsed: The good thing is with benchmarking, we can try different approaches and test their actual cost.

crml/staking/src/rewards/mod.rs
aliXsed: The remaining blocks are those we can postpone the payout to. One is the current one. When zero blocks left, all remaining payouts should be processed.

aliXsed:
@holygits The interesting fact that the benchmarking shows is that all the optimizations we have done so far, hasn't affected the weight of the zero exit case but has made the non zero cases ever worse! https://github.com/cennznet/cennznet/pull/338/commits/ec09d4f492e161c5f82a691dfc54dcade580f6a0
P.S.: After the following commit the benchmark result for the non zero cases yielded the initial better results:  https://github.com/cennznet/cennznet/pull/338/commits/d642854562973c734b5e73d83accbcd0c23f7acc 
So I guess our optimisations didn't help but didn't hinder either. This is possible because most of the obvious optimisations that we can catch, the compiler can catch and address too.

runtime/src/lib.rs

crml/staking/src/rewards/mod.rs
aliXsed: I would suggest we postpone this to later when the actual usage and scale of our network is better known. That's to avoid any premature optimisations.

crml/staking/src/rewards/mod.rs
aliXsed: Logging is now totally improved.

crml/staking/src/mock.rs
aliXsed: Yes, `rewards::Trait` should be implemented for `mock::Test` so test_benchmark_process_reward_payouts::<Test>() can work.

crml/staking/src/rewards/default_weights.rs
```diff
@@ -0,0 +1,35 @@
...
+	fn process_reward_payouts(p: u32) -> Weight {
+		(79_140_000 as Weight)
+			.saturating_add((50_308_000 as Weight).saturating_mul(p as Weight))
+			.saturating_add(DbWeight::get().reads(5 as Weight))
+			.saturating_add(DbWeight::get().reads((1 as Weight).saturating_mul(p as Weight)))
+			.saturating_add(DbWeight::get().writes(4 as Weight))
+			.saturating_add(DbWeight::get().writes((1 as Weight).saturating_mul(p as Weight)))
+	}
+	fn process_zero_payouts() -> Weight {
+		(7_000_000 as Weight).saturating_add(DbWeight::get().reads(1 as Weight))
```
jordy25519, line 33: at first I was worried this seemed high for every block but works out to `0.0000035%` of the maximum weight so seems good!

crml/staking/src/rewards/mod.rs
```diff
@@ -230,6 +322,26 @@ impl<T: Trait> Module<T> {
 		));
 		(*payouts).to_vec()
 	}
+
+	/// Return the number of reward payouts that need to be processed in the current block.
+	/// The result is dependent on the number of the current era's remaining payouts and the number
+	/// of remaining blocks before a new era.
+	fn calculate_payout_quota(remaining_payouts: u32, remaining_blocks: T::BlockNumber) -> u32 {
+		if remaining_blocks == Zero::zero() {
```
jordy25519: `if remaining_blocks.is_zero()`

crml/staking/src/rewards/default_weights.rs
```diff
@@ -0,0 +1,35 @@
...
+use frame_support::weights::{constants::RocksDbWeight as DbWeight, Weight};
+
+impl crate::rewards::WeightInfo for () {
+	fn process_reward_payouts(p: u32) -> Weight {
+		(79_140_000 as Weight)
+			.saturating_add((50_308_000 as Weight).saturating_mul(p as Weight))
+			.saturating_add(DbWeight::get().reads(5 as Weight))
+			.saturating_add(DbWeight::get().reads((1 as Weight).saturating_mul(p as Weight)))
+			.saturating_add(DbWeight::get().writes(4 as Weight))
+			.saturating_add(DbWeight::get().writes((1 as Weight).saturating_mul(p as Weight)))
```
jordy25519, line 30: Could you add the benchmarking command used for this in a comment somewhere so it's reproducible?

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +172,88 @@ where
...
 
-		// Any unallocated reward amount can go to the development fund
-		let remainder = total_payout.saturating_sub(total_payout_imbalance.peek());
-		total_payout_imbalance.subsume(
-			T::CurrencyToReward::deposit_into_existing(&T::TreasuryModuleId::get().into_account(), remainder)
-				.unwrap_or_else(|_| PositiveImbalanceOf::<T>::zero()),
-		);
+		RemainingRewardAmounts::<T>::mutate(|rra| {
+			rra.push_back(total_payout.saturating_sub(total_payout_imbalance.peek()))
+		});
```
jordy25519: Can use the already calculated amount
could also unit-test that `RemainingRewardAmounts` is being set correctly by this method
```suggestion
		RemainingRewardAmounts::<T>::mutate(|rra| {
			rra.push_back(validator_payout)
		});
```

crml/staking/src/rewards/mod.rs
```diff
@@ -76,6 +91,10 @@ decl_storage! {
 		pub TransactionFeePot get(fn transaction_fee_pot): BalanceOf<T>;
 		/// Historic accumulated transaction fees on reward payout
 		pub TransactionFeePotHistory get(fn transaction_fee_pot_history): VecDeque<BalanceOf<T>>;
+		/// Remaining reward amount of the eras which are not fully processed yet
+		pub RemainingRewardAmounts get(fn remaining_reward_amount): VecDeque<BalanceOf<T>>;
```
jordy25519: Maybe `QueuedEraReward`

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +172,88 @@ where
...
+	fn process_reward_payouts(remaining_blocks: Self::BlockNumber) -> Weight {
+		let remaining_payouts = Payouts::<T>::get().len() as u32;
+		let quota = Self::calculate_payout_quota(remaining_payouts, remaining_blocks);
+		if quota.is_zero() {
+			return T::WeightInfo::process_zero_payouts();
+		}
+
+		let weight = T::WeightInfo::process_reward_payouts(quota as u32);
+
+		Payouts::<T>::mutate(|p| {
```
jordy25519: nit can reduce nesting here:
- `let mut p = Payouts::<T>::get()`
- <use / modify p>
- `Payouts::<T>::insert(p);` when done

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +172,88 @@ where
...
+			// First payout in the current series. Without separating the first payout,
```
jordy25519: can do this trick to get an imbalance with asset ID: `T::CurrencyToReward::burn(Zero::zero());`

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +172,88 @@ where
...
+			// PositiveImbalanceOf::<T>::zero() to it, would not set the underlying asset id correctly.
+			let (first_payee, first_amount, first_era) = p.pop_front().unwrap_or_default();
+			let mut total_payout_imbalance = T::CurrencyToReward::deposit_into_existing(&first_payee, first_amount)
+				.ok()
+				.unwrap_or_else(PositiveImbalanceOf::<T>::zero);
+			Self::deposit_event(RawEvent::RewardPayout(first_payee, first_amount, first_era));
+			let mut era_under_process = first_era;
+			let mut era_reward_amount = first_amount;
+
+			let clear_up = |imbalance: &mut PositiveImbalanceOf<T>| -> BalanceOf<T> {
```
jordy25519: nit: something about handling remainder e.g `handle_remainder`
can't this deposit the event also?
not sure closure adds anything here over an ordinary function

crml/staking/src/rewards/mod.rs
```diff
@@ -151,17 +172,88 @@ where
...
+				remainder
+			};
+
+			for _ in 1..quota {
+				if let Some((payee, amount, era)) = p.pop_front() {
+					if era > era_under_process {
+						let remainder = clear_up(&mut total_payout_imbalance);
+						Self::deposit_event(RawEvent::AllRewardsPaidOut(
+							era_under_process,
+							era_reward_amount,
```
jordy25519: if the era rewards are paid across multiple payouts won't `era_reward_amount` be only representative of this payout not the entire era's worth, if so what usefulness is the value?

crml/staking/src/rewards/mod.rs
```diff
@@ -563,12 +689,109 @@ mod tests {
...
+			for i in 0..validators_number {
+				validators_stake_info.push(
+					MockCommissionStakeInfo::new(
+						((i + 1) * 10, 1_000),
+						vec![(2, 2_000), (3, 3_000)],
+						Perbill::from_percent(10),
+					)
+					.as_tuple(),
+				);
+			}
```
jordy25519: more idiomatic
```suggestion
			let validators_stake_info = (0..validators_number).map(|i| {
					MockCommissionStakeInfo::new(
						((i + 1) * 10, 1_000),
						vec![(2, 2_000), (3, 3_000)],
						Perbill::from_percent(10),
					).as_tuple(),
				);
			}).collect();
```

jordy25519 (commented):
AllRewardsPaidOut value should be the era total value, however as long as we signal the start and end of an era payout with events that's enough, so could put the total in the starting event?

left some suggestions to make `process_reward_payouts` easier to follow / less nested

lvgtm in general!

crml/staking/src/rewards/mod.rs
jordy25519: would be nice if we could queue the treasury payout like any other payout

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/default_weights.rs
aliXsed: Done

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs
aliXsed: Actually separating the first one, would be still needed as it gives the right context for processing the rest. Also T::CurrencyToReward::burn(Zero::zero()); has a cost, though trivial.

crml/staking/src/rewards/mod.rs

crml/staking/src/rewards/mod.rs
aliXsed: Addressed by https://github.com/cennznet/cennznet/pull/338/commits/b7844ebfb512b3525f12d67f193dd3fbb11a7742

crml/staking/src/rewards/mod.rs
aliXsed: - Will change to the name to `handle_remainder`
- If I want to include deposit event in this lambda then I need to pass era_under_process and era_reward_amount as arguments too. Simply using them inside the lambda distresses the borrow checker. So, in that case, it wouldn't be even neat or useful.
- I used closure because I didn't want to add a function to either the module or the interface, while this function is only used locally here

aliXsed: > would be nice if we could queue the treasury payout like any other payout

Treasury payout only happens once in an era and there is no reason to hold it and process it later. So perhaps it's not worth it.

crml/staking/src/rewards/mod.rs
alexsednz: era_reward_amount is representing the whole rewards accrued on a particular era (the sum of those with the same era tag).

aliXsed: Yup, I got what you mean. I removed the era_reward_amount for now.

jordy25519 (approved):
lgtm 💪
Just needs conflict resolving

PR #339, opened 2020-12-01
Prerelease/1.2.0 hotifx 1

jordy25519:
Fix JSON-RPC for `@polkadot/rpc-provider@1.2.1`
Intended as a hotfix for 1.2.0 until the upgrade to v37 runtime.

--
Ignore GH actions changes, they're to fix the release behaviour when target branch is a `release/*` branch

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
-        toolchain: nightly-2020-09-27
+        toolchain: nightly-2020-11-18
         components: rustfmt
         target: wasm32-unknown-unknown
         default: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,9 @@
 name: Release
 
 on:
+  push:
+    branches:
+      - "prerelease/*"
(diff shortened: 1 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -27,14 +28,15 @@ jobs:
...
+    if: ${{ startsWith(github.head_ref, 'prerelease/') }} && ${{ github.event.label.name == 'automerge' }}
     needs:
     - merge-release-branch
     runs-on: [self-hosted, linux]
     steps:
     - name: Check out develop
       uses: actions/checkout@v2
       with:
-        ref: develop
+        ref: release/1.2.0
```
aliXsed, line 39: This is more specific than I would expect here. Just wanted to check that it's intended.

aliXsed (approved):

.github/workflows/release.yml
jordy25519: yeah need to find a proper variable for 'target branch name'...

jordy25519: `${{ github.event.pull_request.base.ref }}` this change will only affect this branch. Will make a proper PR to develop with the real change

KarishmaBothara (approved):

PR #344, opened 2020-12-10
Prerelease/1.2.2

jordy25519:
revert epoch duration change 😅 
this will be the new canonical release for v37 support


runtime/src/lib.rs
```diff
@@ -112,7 +112,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// implementation changes and behavior does not, then leave `spec_version` as
 	// is and increment `impl_version`.
 	spec_version: 37,
-	impl_version: 37,
+	impl_version: 38,
```
KarishmaBothara: this will be release 38? runtime version 38

jordy25519: it's only `imp_version` change but yeah might be best to leave it at `37`

KarishmaBothara (approved):

genesis/azalea.raw.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "CENNZnet Azalea",
   "id": "CENNZnet Azalea V1",
+  "badBlocks": [
+    "0x058c78f6570d6be57f7e31bbb74656957e94cec3b0040b9fdb93b8e870aaf25e"
```
jordy25519, line 5: for posterity: ~10 blocks before the fork with botched v37 runtime upgrade

PR #353, opened 2021-01-12
Increase SESSIONS_PER_ERA 24 times

### LINKED ISSUE #352: Increase staking era from 1 hour to 1 day
As Plug and Substrate are not supporting a change to Babe Epoch duration, we need to change the number of sessions per era for this purpose.

aliXsed:
Closes #352

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -726,6 +726,7 @@ version = "1.2.2"
 dependencies = [
  "cennznet-cli",
  "cennznet-primitives",
+ "cennznet-runtime",
  "crml-cennzx",
  "crml-cennzx-rpc-runtime-api",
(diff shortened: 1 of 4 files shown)


jordy25519 (approved):

jordy25519 (changes requested):
I think some tests are now going to run for 24 hours 😓 which is why CI ~is~ appears stalled.
will need to resolve that

jordy25519:
@alexsednz here's a quirky workaround to create integration tests specific config
https://github.com/cennznet/cennznet/compare/chore/352-increase-session-per-era...idea/integration-config?expand=1

jordy25519 (approved):

PR #358, opened 2021-01-20
Add rata as a --chain option

jordy25519:
Regenerate rata with new runtime (so we don't need to keep updating it manually)

Reduce initial total issuance

---

Drive by fix:
~Added some asserts to staking genesis config after @KarishmaBothara noticed staking ledger balance is `0` when starting dev node. I think it is same issue you had @alexsednz where GA storage isn't initialized/available yet. Doesn't fix the issue, just means we will error instead of silent fail on setup~

- Reducing total issuance on dev+rata fixes the exposure calculations total


aliXsed (approved):
lgtm!

PR #363, opened 2021-01-28
Nft module

jordy25519:
Adds a dynamic NFT module
The goal of this module is firstly to allow dapps to create and manage simple NFTs without writing rust code
Secondly, it should be extensible so that NFT behaviour can be extended and managed by smart contacts

TODO:
- [x] Benchmarks
- [x] Direct sales
- [x] General auction (amount, optional reserve, close time)
cc cennznet/rfcs/pull/1

Optional / Nice to have:
- [ ] Require a deposit to create new NFT classes (disincentivize camping on collection names for free)
- [x] Add description tag to NFT field schemas

---

new type defs
```json
{
  "CollectionId": "String",
  "TokenId": "u32",
  "NFTAttributeValue": {
    "_enum": {
      "i32": "i32",
      "u8": "u8",
      "u16": "u16",
      "u32": "u32",
      "u64": "u64",
      "u128": "u128",
      "Bytes32": "[u8; 32]",
      "Bytes": "Bytes",
      "String": "String",
      "Hash": "[u8; 32]",
      "Timestamp": "Moment",
      "Url": "String"
    }
  },
  "NFTAttributeName": "String",
  "NFTAttributeTypeId": {
    "_enum": {
      "i32": null,
      "u8": null,
      "u16": null,
      "u32":  null,
      "u64": null,
      "u128": null,
      "Bytes32": null,
      "Bytes": null,
      "String": null,
      "Hash": null,
      "Timestamp": null,
      "Url": null
    }
  },
  "NFTSchema": {
    "_": "Vec<(String, NFTAttributeTypeId)>"
  },
  "RoyaltiesSchedule": {
    "entitlements": "Vec<(AccountId, Percent)>"
  },
  "Listing": {
    "_enum": {
      "DirectListing": "DirectListing",
      "AuctionListing": "AuctionListing"
    }
  },
  "DirectListing": {
    "payment_asset": "AssetId",
    "fixed_price": "Balance",
    "close": "BlockNumber",
    "buyer": "AccountId"
  },
  "AuctionClosureReason": {
    "_enum": {
      "ExpiredNoBids": null,
      "SettlementFailed": null,
      "VendorCancelled": null
    }
  }
}
```


cli/Cargo.toml
```diff
@@ -15,69 +15,69 @@ futures = { version = "0.3.1", features = ["compat"] }
...
-sp-consensus-babe = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sc-consensus = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sc-consensus-epochs = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sp-consensus = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sc-network = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sc-finality-grandpa = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sp-finality-grandpa = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sc-client-api = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
-sp-runtime = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
+sc-authority-discovery = { branch = "feat/multi-currency-reserve", git = "https://github.com/plugblockchain/plug-blockchain" }
```
aliXsed: Depending on the status of a branch is not a good idea. I would suggest releasing a tag on Plug like 2.0.6 for this purpose.

crml/nft/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+name = "crml-nft"
+version = "1.0.0"
+authors = ["Centrality Developers <support@centrality.ai>"]
+edition = "2018"
+repository = "https://github.com/cennznet/cennznet"
+description = "CENNZnet NFT module"
+
+[dependencies]
+cennznet-primitives = { path = "../../primitives", default-features = false }
+serde = { version = "1.0.101", optional = true }
```
aliXsed: In cli and some other modules 1.0.102 is used. In my experience, it's safer to have a single version of serder every everywhere. Would you want to uniform them at this stage?

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! # NFT Module
+//!
+//! Provides the basic creation and management of dynamic NFTs (created at runtime).
+//!
+//! Intended to be used "as is" by dapps and provide basic NFT feature set for smart contracts
+//! to extend.
+
+use cennznet_primitives::types::{AssetId, Balance};
```
aliXsed, line 24: Relying on cennznet_primitives makes the module less flexible plus type collision possibilities in the future where for example asset id is set differently in different places. Most other modules depend on generic types for all their used types. I suggest we follow the same pattern for these two as well.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		BidTooLow
+	}
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Nft {
+		/// Map from collection to owner address
+		pub CollectionOwner get(fn collection_owner): map hasher(blake2_128_concat) CollectionId => Option<T::AccountId>;
+		/// Map from collection to its schema definition
+		pub CollectionSchema get(fn collection_schema): map hasher(blake2_128_concat) CollectionId => Option<NFTSchema>;
```
aliXsed, line 158: Under which scenario we can have a collection id stored with None schema?

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		RoyaltiesOvercommitment,
+		/// Auction bid was lower than reserve or current highest bid
+		BidTooLow
+	}
+}
+
+decl_storage! {
+	trait Store for Module<T: Trait> as Nft {
+		/// Map from collection to owner address
+		pub CollectionOwner get(fn collection_owner): map hasher(blake2_128_concat) CollectionId => Option<T::AccountId>;
```
aliXsed, line 156: Do we need the Option here? Looking at the code I don't see how a collection id can map to None.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// Map from collection to its schema definition
+		pub CollectionSchema get(fn collection_schema): map hasher(blake2_128_concat) CollectionId => Option<NFTSchema>;
+		/// Map from collection to it's defacto royalty scheme
+		pub CollectionRoyalties get(fn collection_royalties): map hasher(blake2_128_concat) CollectionId => Option<RoyaltiesSchedule<T::AccountId>>;
```
aliXsed, line 160: In this case, royalty is in fact optional for a collection id, but we can save the storage by saving royalties only they if they exist. This way we can avoid storing the collectionid => None cases. And thus the option as the type can be avoided.

crml/nft/src/lib.rs
jordy25519: It won't happen, this is to present a nice API so instead of returning a default when it doesn't exist it returns None

crml/nft/src/lib.rs
jordy25519: Abit of future proofing but also same reason as above. it's nicer to return none than the default account

crml/nft/src/lib.rs
jordy25519: Yeah I realized we pay alot for the generic-ness mostly as a habit from substrate and plug but it doesn't give us any value in CENNZnet, just makes code confusing maintaining all the type conversions

crml/nft/src/lib.rs
aliXsed: A potential storage entry mapped to None is a potential for extra unneeded storage. I would suggest we leave the default case to be handled with the API.

crml/nft/src/lib.rs
aliXsed: Same here, an empty vector is clear enough that there is no schema.

crml/nft/src/lib.rs
aliXsed: I get what you mean, but the mix of concrete types and generic types especially for the types that are defined generic in some other modules, doesn't seem like a good pattern to me. We are specifying asset id and balance in the runtime for some modules and not for this one! This is a possibility for type collission and looks untidy to me.

aliXsed: I would be leaning toward agreeing with you for the types that are specific to this module.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// Map from a token to it's royalty scheme
+		pub TokenRoyalties get(fn token_royalties): double_map hasher(blake2_128_concat) CollectionId, hasher(twox_64_concat) T::TokenId => Option<RoyaltiesSchedule<T::AccountId>>;
```
aliXsed, line 162: Same here, the default RoyaltiesSchedule is distinct enough. So I'd be on the side of no need for the return value to be an option.

crml/nft/src/lib.rs
jordy25519: I don't think anything is stored because `Option::default()` is `None`

crml/nft/Cargo.toml
jordy25519: will update this

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// Map from (collection, token) to it's attributes (as defined by schema)
+		pub Tokens get(fn tokens): double_map hasher(blake2_128_concat) CollectionId, hasher(twox_64_concat) T::TokenId => Vec<NFTAttributeValue>;
+		/// The next available token Id for an NFT collection
+		pub NextTokenId get(fn next_token_id): map hasher(twox_64_concat) CollectionId => T::TokenId;
```
aliXsed, line 166: Can we use the `last` function of the iterator to the Tokens storage and avoid this storage entirely.

jordy25519: No because if a token is burnt it will be removed from the tokens map.
Should use `decode_len` rather than iterator for getting item count

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// The total amount of an NFT collection in existence
+		/// Map from (collection, token) to it's owner
+		pub TokenOwner get(fn token_owner): double_map hasher(blake2_128_concat) CollectionId, hasher(blake2_128_concat) T::TokenId => T::AccountId;
+		/// Map from (collection, account) to the account owned tokens of that collection
+		pub CollectedTokens get(fn collected_tokens): double_map hasher(blake2_128_concat) CollectionId, hasher(blake2_128_concat) T::AccountId => Vec<T::TokenId>;
```
aliXsed: This is another view to the same storage as held by TokenOwner. Design-wise, the redundancy here may lead to potential out of sync bugs in the future, plus it's icreasing the amount of storage. I understand it may increase the performance in some scenarios. But I'd be on the side that we need to resort to these hacks when we have measured the performance for an end to end use case and the need for optimisation is clear.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// The total amount of an NFT collection in existence
+		pub TokenIssuance get(fn token_issuance): map hasher(blake2_128_concat) CollectionId => T::TokenId;
```
aliXsed, line 170: I'd say for this purpose adding an extrinsic that would count `Tokens` is a better idea.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// The total amount of an NFT collection burned
+		pub TokensBurnt get(fn tokens_burnt): map hasher(blake2_128_concat) CollectionId => T::TokenId;
```
aliXsed: Why do we need this?

crml/nft/src/lib.rs
aliXsed: Under the hood, what's stored for a map's value is actually an `Option` (look at `fn place_storage(&mut self, key: StorageKey, value: Option<StorageValue>`). Tracing the map storage's insert function, down to `fn set(
		&mut self,
		value: V,
		first_write_in_tx: bool,
		at_extrinsic: Option<u32>,
	)` in `impl<V> OverlayedEntry<V>`, it's not obvious to me where the map refuses to insert a value which happens to be the same as the default value.

crml/nft/src/lib.rs
aliXsed: Even when a token is burnt, the last token should have the token id from which we can generate the next one.

crml/nft/src/lib.rs
jordy25519: Storage map get calls do this
https://github.com/plugblockchain/plug-blockchain/blob/ebfb7f707a7ccd066213176d9385cd46830cb680/frame/support/src/storage/generator/map.rs#L225-L227

inside it creates a default return value
https://github.com/plugblockchain/plug-blockchain/blob/ebfb7f707a7ccd066213176d9385cd46830cb680/frame/support/procedural/src/storage/storage_struct.rs#L27-L38

Option<T> default is None
https://doc.rust-lang.org/src/core/option.rs.html#1326-1340

jordy25519: we don't store `None` it just means it doesn't exist which is a better API than returning default value

crml/nft/src/lib.rs
jordy25519: what if the last token is burnt?

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			Self::deposit_event(RawEvent::Burn(collection_id, token_id));
+		}
+
+		/// Sell an NFT to specific account at a fixed price
+		/// `buyer` optionally, the account to receive the NFT. If unspecified, then any account may purchase
+		/// `asset_id` fungible asset Id to receive as payment for the NFT
+		/// `fixed_price` ask price
+		/// Caller must be the token owner
+		#[weight = T::WeightInfo::direct_sale()]
+		fn direct_sale(origin, collection_id: CollectionId, token_id: T::TokenId, buyer: Option<T::AccountId>, payment_asset: AssetId, fixed_price: Balance) {
```
aliXsed: I would suggest the duration to be an optional input parameter. Especially important for the fixed price sale.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			ensure!(schema.len() <= MAX_SCHEMA_FIELDS as usize, Error::<T>::SchemaMaxAttributes);
+
+			// Check the provided attribute types are valid
+			ensure!(
+				schema.iter().all(|(_name, type_id)| NFTAttributeValue::is_valid_type_id(*type_id)),
+				Error::<T>::SchemaInvalid
+			);
+
+			// Attribute names must be unique (future proofing for map lookups etc.)
+			let (attribute_names, _): (Vec<NFTAttributeName>, Vec<NFTAttributeTypeId>) = schema.iter().cloned().unzip();
```
aliXsed: I would expect `schema.iter().map().collect()` to be more efficient than 'unzip', when you are throwing away `Vec<NFTAttributeTypeId>` here.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// Caller must be the collection owner
+		#[weight = T::WeightInfo::create_token()]
+		#[transactional]
+		fn create_token(origin, collection_id: CollectionId, owner: T::AccountId, attributes: Vec<NFTAttributeValue>, royalties_schedule: Option<RoyaltiesSchedule<T::AccountId>>) -> DispatchResult {
+			let origin = ensure_signed(origin)?;
+
+			// Permission and existence check
+			let collection_owner = Self::collection_owner(&collection_id);
+			ensure!(collection_owner.is_some(), Error::<T>::NoCollection);
+			ensure!(collection_owner.unwrap() == origin, Error::<T>::NoPermission);
```
aliXsed, line 261: If you do `ensure!(collection_owner == Some(origin), Error::<T>::NoPermission);` here, the safety of this line wouldn't depend on the previous line.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+
+			// Check we can issue a new token
+			let token_id = Self::next_token_id(&collection_id);
+			let next_token_id = token_id.checked_add(&One::one()).ok_or(Error::<T>::NoAvailableIds)?;
+			Self::token_issuance(&collection_id).checked_add(&One::one()).ok_or(Error::<T>::MaxTokensIssued)?;
+
+			// Quick `attributes` sanity checks
+			ensure!(!attributes.is_empty(), Error::<T>::SchemaEmpty);
+			ensure!(attributes.len() as u32 <= MAX_SCHEMA_FIELDS, Error::<T>::SchemaMaxAttributes);
```
aliXsed: I suggest the above two checks to precede the storage read `Self::collection_owner` at line 270, so we avoid 2 reads when there is an error in the input param.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			// Check the provided attribute types are valid
+			ensure!(
+				schema.iter().all(|(_name, type_id)| NFTAttributeValue::is_valid_type_id(*type_id)),
+				Error::<T>::SchemaInvalid
+			);
+
+			// Attribute names must be unique (future proofing for map lookups etc.)
+			let (attribute_names, _): (Vec<NFTAttributeName>, Vec<NFTAttributeTypeId>) = schema.iter().cloned().unzip();
+			let deduped = BTreeSet::from_iter(attribute_names);
+			ensure!(deduped.len() == schema.len(), Error::<T>::SchenmaDuplicateAttribute);
```
aliXsed: I think we should allow the names for the attributes to be optional as a token schema creator may want to rely solely on the order of the fields and give a name to them in the application. They can save another bit of storage by that. If we do so, then we should handle the default or empty attribute name here as well.

crml/nft/src/lib.rs
aliXsed: I understand that when an entry doesn't exist, the storage map would create a default and return that. My question was, if you try to store None, would it check and return early to avoid occupying any storage?

crml/nft/src/lib.rs
aliXsed: That's a good question. Would it be a problem to reuse the token id in that case? I'm thinking we'd burn tokens to free their resources. There should be a good reason why a token id should remain reserved for a burnt token.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+
+			// Update token ownership
+			<CollectedTokens<T>>::mutate(&collection_id, current_owner, |tokens| {
+				tokens.retain(|t| t != &token_id)
+			});
+			<TokenOwner<T>>::take(&collection_id, token_id);
+			<Tokens<T>>::take(&collection_id, token_id);
+
+			// Will not overflow, cannot exceed the amount issued qed.
+			let tokens_burnt = Self::tokens_burnt(&collection_id).checked_add(&One::one()).unwrap();
```
aliXsed: Unwrap can potentially crash the network. Should avoid at all costs.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			<TokensBurnt<T>>::insert(&collection_id, tokens_burnt);
```
aliXsed: `mutate` seems to be a better choice here because it can include the previous line.

crml/nft/src/types.rs
```diff
@@ -0,0 +1,233 @@
...
+					.entitlements
+					.iter()
+					.map(|(_who, share)| share.deconstruct() as u32)
+					.sum::<u32>() <= Permill::ACCURACY
+	}
+	/// Calculate the total % entitled for royalties
+	/// It will return `0` if the `entitlements` are overcommitted
+	pub fn calculate_total_entitlement(&self) -> Permill {
+		// if royalties are in a strange state
+		if !self.validate() {
```
aliXsed, line 143: Looks like, we check the validity of royalties on creation of a collection. So this check here can be removed. Also, design wise, the public interfaces shouldn't call each other.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+				if royalty_fees.is_zero() {
+					// full proceeds to seller/`current_owner`
+					T::MultiCurrency::transfer(&origin, &current_owner, Some(listing.payment_asset), listing.fixed_price, ExistenceRequirement::AllowDeath)?;
+				} else {
+					// withdraw funds from buyer, split between royalty payments and seller
+					let for_royalties = royalty_fees * listing.fixed_price;
+					let for_seller = listing.fixed_price - for_royalties;
+
+					let mut imbalance = T::MultiCurrency::withdraw(&origin, Some(listing.payment_asset), listing.fixed_price, WithdrawReason::Transfer.into(), ExistenceRequirement::AllowDeath)?;
+					imbalance = imbalance.offset(T::MultiCurrency::deposit_into_existing(&current_owner, Some(listing.payment_asset), for_seller)?).map_err(|_| Error::<T>::InternalPayment)?;
```
aliXsed: We shouldn't return here, in both `?` cases of this line, before returning the withdrawn balance from the buyer.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+					for (who, entitlement) in royalties_schedule.entitlements.into_iter() {
+						let amount = entitlement * listing.fixed_price;
+						imbalance = imbalance.offset(T::MultiCurrency::deposit_into_existing(&who, Some(listing.payment_asset), amount)?).map_err(|_| Error::<T>::InternalPayment)?;
```
aliXsed: I would suggest paying the royalties first while skipping the payments to the accounts who don't exist and then the rest/remaining amount to the seller.

aliXsed: I am particularly concerned with the case where Round(ax+ay) != Round(ax) + Round(ay), suppose a is the price and x and y are the royalty shares. In the case that we calculate the royalties first and pay the rest to the seller, there wouldn't be any different calculation for the share of the seller and thus we wouldn't encounter this issue.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		/// The total amount of an NFT collection in existence
+		pub TokenIssuance get(fn token_issuance): map hasher(blake2_128_concat) CollectionId => T::TokenId;
+		/// The total amount of an NFT collection burned
+		pub TokensBurnt get(fn tokens_burnt): map hasher(blake2_128_concat) CollectionId => T::TokenId;
+		/// NFT sale/auction listings. keyed by collection id and token id
+		pub Listings get(fn listings): double_map hasher(blake2_128_concat) CollectionId, hasher(twox_64_concat) T::TokenId => Option<Listing<T>>;
+		/// Winning bids on open listings. keyed by collection id and token id
+		pub ListingWinningBid get(fn listing_winning_bid): double_map hasher(blake2_128_concat) CollectionId, hasher(twox_64_concat) T::TokenId => Option<(T::AccountId, Balance)>;
+		/// Map from block numbers to listings scheduled to close
+		pub ListingEndSchedule get(fn listing_end_schedule): map hasher(twox_64_concat) T::BlockNumber => Vec<(CollectionId, T::TokenId)>;
```
aliXsed: The fact that we cannot have the same pair of (collection id - token id) repeated for a specific block number suggests we better have BTreeSet instead of Vec here.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		<CollectedTokens<T>>::mutate(collection_id, current_owner, |tokens| tokens.retain(|t| t != &token_id));
+		<TokenOwner<T>>::insert(collection_id, token_id, new_owner);
+		<CollectedTokens<T>>::append(collection_id, new_owner, token_id);
+	}
+	/// Remove a single direct listing and all it's metadata
+	fn remove_direct_listing(collection_id: &CollectionId, token_id: T::TokenId) {
+		let listing_type = Listings::<T>::take(collection_id, token_id);
+		ListingWinningBid::<T>::remove(collection_id, token_id);
+		if let Some(Listing::<T>::DirectSale(listing)) = listing_type {
+			ListingEndSchedule::<T>::mutate(listing.close, |listings| {
```
aliXsed: With `BtreeSet` instead of `Vec` for tha values of this map, you can simply use the `remove` function here.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			Self::deposit_event(RawEvent::AuctionOpen(collection_id, token_id, payment_asset, reserve_price));
+		}
+
+		/// Place a bid on an open auction
+		/// - `amount` to bid (in the seller's requested payment asset)
+		#[weight = T::WeightInfo::bid()]
+		#[transactional]
+		fn bid(origin, collection_id: CollectionId, token_id: T::TokenId, amount: Balance) {
+			let origin = ensure_signed(origin)?;
+			ensure!(<Listings<T>>::contains_key(&collection_id, token_id), Error::<T>::NotForAuction);
```
aliXsed: The next line check covers this too. So you can remove this check.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+
+			if let Some(Listing::Auction(listing)) = Self::listings(&collection_id, token_id) {
+				if let Some(current_bid) = Self::listing_winning_bid(&collection_id, token_id) {
+					ensure!(amount > current_bid.1, Error::<T>::BidTooLow);
+				} else {
+					// first bid
+					ensure!(amount >= listing.reserve_price, Error::<T>::BidTooLow);
```
aliXsed, line 456: We only need to do this check when settling the auction. It's good piece of information to the seller if they know there has been a bid though below the reserve_price.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			Ok(())
+		}
+
+		/// Issue a new NFT
+		/// `owner` - the token owner
+		/// `attributes` - initial values according to the NFT collection/schema
+		/// `royalties_schedule` - optional royalty schedule for secondary sales of _this_ token, defaults to the collection config
+		/// Caller must be the collection owner
+		#[weight = T::WeightInfo::create_token()]
+		#[transactional]
```
aliXsed, line 254: I notice that some extrinsics and functions in this module marked as transactional some others don't. What have been the criteria for this?

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+					ensure!(amount > current_bid.1, Error::<T>::BidTooLow);
+				} else {
+					// first bid
+					ensure!(amount >= listing.reserve_price, Error::<T>::BidTooLow);
+				}
+
+				// check user has the requisite funds to make this bid
+				let balance = T::MultiCurrency::free_balance(&origin, Some(listing.payment_asset));
+				if let Some(balance_after_bid) = balance.checked_sub(amount) {
+					// TODO: review this during 3.0 upgrade
```
aliXsed: Can you please mention this on https://github.com/cennznet/cennznet/issues/414? Also it would be good to add the link to 414 to the TODO comment so we ensure every TODO has an issue number assigned to it.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+			ensure!(current_owner == origin, Error::<T>::NoPermission);
+
+			match Self::listings(&collection_id, token_id) {
+				Some(Listing::<T>::DirectSale(sale)) => {
+					ListingEndSchedule::<T>::mutate(sale.close, |schedule| schedule.retain(|(c, t)| (c, t) != (&collection_id, &token_id)));
+					Listings::<T>::remove(&collection_id, token_id);
+					Self::deposit_event(RawEvent::DirectSaleClosed(collection_id, token_id));
+				},
+				Some(Listing::<T>::Auction(auction)) => {
+					ensure!(Self::listing_winning_bid(&collection_id, token_id).is_none(), Error::<T>::TokenListingProtection);
```
aliXsed, line 502: If we decide to keep bids below the reserve price until the settlement, which I highly suggest, here we should check if there is a bid that meets the reserved price.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+					Listings::<T>::remove(&collection_id, token_id);
+					ListingEndSchedule::<T>::mutate(auction.close, |schedule| schedule.retain(|(c, t)| (c, t) != (&collection_id, &token_id)));
+					Self::deposit_event(RawEvent::AuctionClosed(collection_id, token_id, AuctionClosureReason::VendorCancelled));
+				},
+				None => {},
+			}
+		}
+
+	}
```
aliXsed, line 511: I would suggest we add another extrinsic and let the sellers ease their auction conditions. Something like reduce_auction_reserve. That would facilitate more successful auctions which will be positive on our economy.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+					} else {
+						// normal closure, no acceptable bids
+						Self::deposit_event(RawEvent::AuctionClosed(
+							collection_id.clone(),
+							token_id,
+							AuctionClosureReason::ExpiredNoBids,
+						));
+					}
+				}
+				_ => (),
```
aliXsed: Please remove this line, so if in the future someone added another item to the Listing enum, they automatically know they should cover the case here as well.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+	#[transactional]
+	fn settle_auction(
+		collection_id: &CollectionId,
+		token_id: T::TokenId,
+		listing: &AuctionListing<T>,
+		winner: &T::AccountId,
+		hammer_price: Balance,
+	) -> DispatchResult {
+		// if there are no custom royalties, fallback to default if it exists
+		let royalties_schedule = if let Some(royalties_schedule) = Self::token_royalties(&collection_id, token_id) {
```
aliXsed: I would highly suggest that we share and reuse a function that would handle royalties between this function and direct_purchase.

crml/nft/src/lib.rs
jordy25519: it's like this to give a more specific error to the user depending on the situation

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+		}
+
+		let current_owner = Self::token_owner(collection_id, token_id);
+		let seller_balance = T::MultiCurrency::free_balance(&current_owner, Some(listing.payment_asset));
+		let _ =
+			T::MultiCurrency::repatriate_reserved(&winner, Some(listing.payment_asset), &current_owner, for_seller)?;
+
+		// The implementation of `repatriate_reserved` may take less than the required amount and succeed
+		// this should not happen but could for reasons outside the control of this module
+		ensure!(
```
aliXsed, line 626: A late error, after royalties are handled, is detrimental to the state of the chain. Please take an approach that wouldn't fail here at all and move any checks to the beginning of this function.

crml/nft/src/lib.rs
jordy25519: this has a 'qed' on the line above, same as substrate practice

crml/nft/src/lib.rs
jordy25519: there's a `#[transactional]` annotation on this method which will rollback on failure

crml/nft/src/lib.rs
jordy25519: should be fixed now

crml/nft/src/lib.rs
jordy25519: please read about `#[transactional]`

crml/nft/src/lib.rs
jordy25519: Seems like nice to have, they can cancel and relist any time so will leave out for now

crml/nft/src/lib.rs
aliXsed: That's alright to have second more specific error. What I am suggesting is that try to avoid `unwrap` which is not safe in general. Here it is safe because in the previous line you have checked `is_some`. But generally better not to be dependent on a previous line if there is a simpler way like what I said `ensure!(collection_owner == Some(origin), Error::<T>::NoPermission);`

crml/nft/src/lib.rs
jordy25519: when you can't setup an 'ensure then store' situation then we need to ensure the changes rollback on failure

crml/nft/src/lib.rs
jordy25519: I ended up changing this to a double map

crml/nft/src/lib.rs
aliXsed: I see. I would still suggest `unwrap_or` or similar instead of playing with a hot burning ball. Who knows how the program would evolve next time someone is touching this code and if they make a mistake. The side effect of a mistake in a production level code shouldn't be a crash.

crml/nft/src/lib.rs
```diff
@@ -0,0 +1,651 @@
...
+
+			ensure!(!schema.is_empty(), Error::<T>::SchemaEmpty);
+			ensure!(schema.len() <= MAX_SCHEMA_FIELDS as usize, Error::<T>::SchemaMaxAttributes);
+
+			let mut set = BTreeSet::new();
+			for (name, type_id) in schema.iter() {
+				// Check the provided attribute types are valid
+				ensure!(NFTAttributeValue::is_valid_type_id(*type_id), Error::<T>::SchemaInvalid);
+				// Attribute names must be unique (future proofing for map lookups etc.)
+				ensure!(set.insert(name), Error::<T>::SchemaDuplicateAttribute);
```
aliXsed, line 232: I mentioned this on the previous revision of this file: I think we should allow the names for the attributes to be optional as a token schema creator may want to rely solely on the order of the fields and give a name to them in the application. They can save another bit of storage by that. If we do so, then we should handle the default or empty attribute name here as well.

crml/nft/src/lib.rs
aliXsed: But if they can cancel, they would loose their bid.

crml/nft/src/lib.rs
aliXsed: Is there any overhead with using #[transactional]?

holygits:
run benchmark

holygits:
@benchmark_bot

crml/nft/src/lib.rs
jordy25519: they can't cancel if there's a successful bid, would prefer to leave it up to the seller to do good price discovery. I don't think it's essential to allow this feature to get this module integrated into CENNZnet

crml/nft/src/lib.rs
jordy25519: Makes it too complicated for edge case at this stage, if it turns out this would be useful we can readdress it

crml/nft/src/lib.rs
jordy25519: yes

crml/nft/src/lib.rs
jordy25519: can you elaborate on this? the unsuccessful bid will still be logged as an event ~on the network~ in the block

crml/nft/src/lib.rs
jordy25519: it's much simpler this way, not much gain to optimise this

crml/nft/src/lib.rs
jordy25519: it would store a 0 byte when inserting `None` afaik

jordy25519:
run benchmark

crml/nft/src/lib.rs
aliXsed: The logged event is good. But when you drop a good offer, then there may come even worse offers after, which will be dropped too. So finding which one was the best is like going through all the logs. Apart from easily knowing what's the best offer so far, this suggestion would make sense if we want to let the seller reduce the reserve price incentivised by the fact that they can keep the best offer without having to go through another round of auction.

aliXsed: There is a strategy behind a couple of my suggestions on the NFT module: "Set and apply limitations only if they are absolutely necessary". This would allow flexibilty, scalabilty and new use cases that you haven't thought at the time of developing your code. Interestingly, most of the time, this makes the code be simpler too. Obviously it's not about allowing any fragility in your design or code. This policy would make sense when talking about safe interactions.

jordy25519: I get the idea but it's just guessing at this point. seller can set reserve price 0 and have no issues

crml/nft/src/lib.rs
aliXsed: So let's use it when avoiding an incomplete state is very hard or due to using other libraries out of our control. We seem to be able to re-arrange our implementation in this case and some other ones and get to the same guarantees with no extra cost. An example here is you check how much is actually reserved from the seller at the beginning of this function before proceeding with repatriation. Anyway, I can understand if you want to leave it to another round of refinement/improvement.

crml/nft/src/lib.rs
aliXsed: 0 would be like they have sold it with no control over the outcome. Anyway, the decision is yours.

PR #364, opened 2021-01-29
Feature/359 block authoring reward

### LINKED ISSUE #359: Reintroduce block reward incentives
Since there's no slashing if a validator misses its slot of authoring a block, the rewards to validators should remain proportional to the number of the blocks they have authored in an era. We need to take the staking reward points on events like note_author, note_uncle coming from the pallet_authorship into account when calculating rewards for each validator.

aliXsed:
We need to do the storage migration for the change that has happened to Payee and CurrentEraPointsEarned before closing #359.

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -135,28 +135,6 @@
 //!
 //! ## Usage
 //!
-//! ### Example: Rewarding a validator by id.
-//!
-//! ```
-//! use frame_support::{decl_module, dispatch};
-//! use frame_system::{self as system, ensure_signed};
-//! use crml_staking::{self as staking};
-//!
-//! pub trait Trait: staking::Trait {}
-//!
-//! decl_module! {
-//! 	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
-//! 		#[weight = 0]
-//! 		pub fn reward_myself(origin) -> dispatch::DispatchResult {
-//! 			let reported = ensure_signed(origin)?;
-//! 			<staking::Module<T>>::reward_by_ids(vec![(reported, 10)]);
-//! 			Ok(())
-//! 		}
-//! 	}
-//! }
-//! # fn main() { }
-//! ```
-//!
 //! ## Implementation Details
 //!
 //! ### Slot Stake
@@ -290,37 +268,6 @@ const STAKING_ID: LockIdentifier = *b"staking ";
 
 /// Counter for the number of eras that have passed.
 pub type EraIndex = u32;
-
-/// Counter for the number of "reward" points earned by a given validator.
-pub type Points = u32;
-
-/// Reward points of an era. Used to split era total payout between validators.
-#[derive(Encode, Decode, Default)]
-pub struct EraPoints {
-	/// Total number of points. Equals the sum of reward points for each validator.
-	total: Points,
-	/// The reward points earned by a given validator. The index of this vec corresponds to the
-	/// index into the current validator set.
-	individual: Vec<Points>,
-}
-
-impl EraPoints {
-	/// Add the reward to the validator at the given index. Index must be valid
-	/// (i.e. `index < current_elected.len()`).
-	fn add_points_to_index(&mut self, index: u32, points: Points) {
-		if let Some(new_total) = self.total.checked_add(points) {
-			self.total = new_total;
-			let new_size = (index as usize + 1).max(self.individual.len());
-			self.individual.resize(new_size, 0);
-			self.individual[index as usize] += points; // Addition is less than total
-		}
-	}
-	/// Get a list of individual points
-	#[cfg(any(feature = "std", test))]
-	pub fn individual_points(&self) -> Vec<Points> {
-		self.individual.clone()
-	}
-}
 
 /// Indicates the initial status of a staker (used by genesis config only).
 #[derive(RuntimeDebug)]
@@ -737,9 +684,6 @@ decl_storage! {
 
 		/// The session index at which the current era started.
 		pub CurrentEraStartSessionIndex get(fn current_era_start_session_index): SessionIndex;
-
-		/// Rewards for the current era. Using indices of current elected set.
-		CurrentEraPointsEarned get(fn current_era_points): EraPoints;
 
 		/// The amount of balance actively at stake for each validator slot, currently.
 		///
@@ -1695,45 +1639,6 @@ impl<T: Trait> Module<T> {
 		slashing::clear_stash_metadata::<T>(stash);
 	}
 
-	/// Add reward points to validators using their stash account ID.
-	///
-	/// Validators are keyed by stash account ID and must be in the current elected set.
-	///
-	/// For each element in the iterator the given number of points in u32 is added to the
-	/// validator, thus duplicates are handled.
-	///
-	/// At the end of the era each the total payout will be distributed among validator
-	/// relatively to their points.
-	///
-	/// COMPLEXITY: Complexity is `number_of_validator_to_reward x current_elected_len`.
-	/// If you need to reward lots of validator consider using `reward_by_indices`.
-	pub fn reward_by_ids(validators_points: impl IntoIterator<Item = (T::AccountId, u32)>) {
-		CurrentEraPointsEarned::mutate(|rewards| {
-			let current_elected = <Module<T>>::current_elected();
-			for (validator, points) in validators_points.into_iter() {
-				if let Some(index) = current_elected.iter().position(|elected| *elected == validator) {
-					rewards.add_points_to_index(index as u32, points);
-				}
-			}
-		});
-	}
-
-	/// Add reward points to validators using their validator index.
-	///
-	/// For each element in the iterator the given number of points in u32 is added to the
-	/// validator, thus duplicates are handled.
-	pub fn reward_by_indices(validators_points: impl IntoIterator<Item = (u32, u32)>) {
-		let current_elected_len = <Module<T>>::current_elected().len() as u32;
-
-		CurrentEraPointsEarned::mutate(|rewards| {
-			for (validator_index, points) in validators_points.into_iter() {
-				if validator_index < current_elected_len {
-					rewards.add_points_to_index(validator_index, points);
-				}
-			}
-		});
-	}
-
 	/// Ensures that at the end of the current session there will be a new era.
 	fn ensure_new_era() {
 		match ForceEra::get() {
@@ -1777,19 +1682,6 @@ impl<T: Trait> SessionManager<T::AccountId, Exposure<T::AccountId, BalanceOf<T>>
 impl<T: Trait> OnNewAccount<T::AccountId> for Module<T> {
 	fn on_new_account(stash: &T::AccountId) {
 		Self::kill_stash(stash);
-	}
-}
-
-/// Add reward points to block authors:
-/// * 20 points to the block producer for producing a (non-uncle) block in the relay chain,
-/// * 2 points to the block producer for each reference to a previously unreferenced uncle, and
-/// * 1 point to the producer of each referenced uncle block.
-impl<T: Trait + pallet_authorship::Trait> pallet_authorship::EventHandler<T::AccountId, T::BlockNumber> for Module<T> {
-	fn note_author(author: T::AccountId) {
-		Self::reward_by_ids(vec![(author, 20)]);
-	}
-	fn note_uncle(author: T::AccountId, _age: T::BlockNumber) {
-		Self::reward_by_ids(vec![(<pallet_authorship::Module<T>>::author(), 2), (author, 1)])
 	}
 }
 
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1613,20 +1613,6 @@ impl<T: Trait> Module<T> {
 			_ => ForceEra::put(Forcing::ForceNew),
 		}
 	}
-
-	#[cfg(test)]
-	pub fn payee(stash: &T::AccountId) -> RewardDestination<T::AccountId> {
-		let payee = T::Rewarder::payee(stash);
-		if payee == *stash {
-			return RewardDestination::Stash;
-		}
-		if let Some(controller) = Bonded::<T>::get(stash) {
-			if payee == controller {
-				return RewardDestination::Controller;
-			}
-		}
-		RewardDestination::Account(payee)
-	}
 }
 
 impl<T: Trait> pallet_session::SessionManager<T::AccountId> for Module<T> {
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1614,8 +1614,7 @@ impl<T: Trait> Module<T> {
 		}
 	}
 
-	#[cfg(test)]
-	fn payee(stash: &T::AccountId) -> RewardDestination<T::AccountId> {
+	pub fn payee(stash: &T::AccountId) -> RewardDestination<T::AccountId> {
 		let payee = T::Rewarder::payee(stash);
 		if payee == *stash {
 			return RewardDestination::Stash;
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1614,11 +1614,7 @@ impl<T: Trait> Module<T> {
 		}
 	}
 
-	/// Return the reward destination for a given stash id.
-	///
-	/// This is mostly a helper function for test purposes. There is a simpler interface,
-	/// HandlePayee, implemented by the rewards module which is suggested to look at first before
-	/// using this one in your code.   
+	#[cfg(test)]
 	pub fn payee(stash: &T::AccountId) -> RewardDestination<T::AccountId> {
 		let payee = T::Rewarder::payee(stash);
 		if payee == *stash {
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -233,7 +233,7 @@ mod multi_token_economy_tests;
 mod tests;
 
 pub mod rewards;
-pub use rewards::StakerRewardPayment;
+pub use rewards::{HandlePayee, StakerRewardPayment};
 
 mod slashing;
 pub use slashing::REWARD_F1;
@@ -594,11 +594,8 @@ pub trait Trait: frame_system::Trait {
 	type SessionInterface: self::SessionInterface<Self::AccountId>;
 
 	/// Handles payout for validator rewards
-	type Rewarder: StakerRewardPayment<
-		AccountId = Self::AccountId,
-		Balance = BalanceOf<Self>,
-		BlockNumber = Self::BlockNumber,
-	>;
+	type Rewarder: StakerRewardPayment<AccountId = Self::AccountId, Balance = BalanceOf<Self>, BlockNumber = Self::BlockNumber>
+		+ HandlePayee<AccountId = Self::AccountId>;
 
 	/// Extrinsic weight info
 	type WeightInfo: WeightInfo;
@@ -648,9 +645,6 @@ decl_storage! {
 		pub Ledger get(fn ledger):
 			map hasher(twox_64_concat) T::AccountId
 			=> Option<StakingLedger<T::AccountId, BalanceOf<T>>>;
-
-		/// Where the reward payment should be made. Keyed by stash.
-		pub Payee get(fn payee): map hasher(twox_64_concat) T::AccountId => RewardDestination<T::AccountId>;
 
 		/// The map from (wannabe) validator stash key to the preferences of that validator.
 		pub Validators get(fn validators):
@@ -886,12 +880,12 @@ decl_module! {
 			// you actually validate/nominate and remove once you unbond __everything__.
 			<Bonded<T>>::insert(&stash, &controller);
 
-			// controller destination is just a custom account after all
-			if let RewardDestination::Controller = payee {
-				<Payee<T>>::insert(&stash, RewardDestination::Account(controller.clone()));
-			} else {
-				<Payee<T>>::insert(&stash, payee);
-			}
+			let id = match payee {
+				RewardDestination::Stash => stash.clone(),
+				RewardDestination::Controller => controller.clone(),
+				RewardDestination::Account(account) => account,
+			};
+			T::Rewarder::set_payee(&stash, &id);
 
 			let stash_balance = T::Currency::free_balance(&stash);
 			let value = value.min(stash_balance);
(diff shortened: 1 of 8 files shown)


KarishmaBothara:
Went through the code changes... makes sense and looks good to me as non rust expert... I will however let Jordy approve it...

crml/staking/src/lib.rs
```diff
@@ -1688,59 +1599,33 @@ impl<T: Trait> Module<T> {
...
-	}
-
 	/// Ensures that at the end of the current session there will be a new era.
 	fn ensure_new_era() {
 		match ForceEra::get() {
 			Forcing::ForceAlways | Forcing::ForceNew => (),
 			_ => ForceEra::put(Forcing::ForceNew),
 		}
 	}
+
```
jordy25519: is this for tests only? if so please mark it that way with `#[cfg(test)]` or refactor tests

crml/staking/src/rewards/mod.rs
```diff
@@ -168,8 +188,9 @@ where
 		}
 
 		let total_payout = Self::calculate_next_reward_payout();
+		let num_of_validators = validator_commission_stake_map.len();
```
jordy25519: nit: `validator_count`

crml/staking/src/rewards/mod.rs
```diff
@@ -272,6 +312,32 @@ where
...
+			Payee::<T>::insert(stash, payee);
+		}
+	}
+
+	/// Remove the corresponding stash-payee from the look up. Do no operations if stash not found.
+	fn remove_payee(stash: &Self::AccountId) {
+		Payee::<T>::remove(stash);
+	}
+
+	/// Return the reward destination for the given stash account.
```
```suggestion
	/// Return the reward account for the given stash account.
```

crml/staking/src/lib.rs
```diff
@@ -1438,37 +1376,10 @@ impl<T: Trait> Module<T> {
 		let validator_commission_stake_map = elected_validator_stashes
 			.iter()
 			.map(|validator_stash| {
-				// Get a version of `Exposure` which maps to preferred payment account _instead of_ stash
```
jordy25519, line 1441: :tada:

crml/staking/src/rewards/mod.rs
```diff
@@ -188,21 +209,40 @@ where
...
+				let validator_reward_points = era_reward_points
+					.individual
+					.get(validator)
+					.map(|points| *points)
+					.unwrap_or_else(|| Zero::zero());
+
+				// This is how much validator + nominators are entitled to.
+				let validator_total_payout = if total_reward_points.is_zero() {
+					// When no authorship points are recorded, divide the payout equally
+					//let total_payout_share = validator_payout / BalanceOf::<T>::from(validator_commission_stake_map.len() as u32);
```
jordy25519: remove this line?

crml/staking/src/rewards/mod.rs
```diff
@@ -825,13 +919,20 @@ mod tests {
 					.as_tuple()
 				})
 				.collect();
+			let note_author_to_all = || {
+				for i in 0..validators_number {
+					Rewards::note_author((i + 1) * 10);
```
```suggestion
				for i in 1..=validators_number {
					Rewards::note_author(i * 10);
```

crml/staking/src/rewards/mod.rs
```diff
@@ -1090,4 +1191,46 @@ mod tests {
...
 	}
+
+	#[test]
+	fn reward_from_authorship_event_handler_works() {
+		ExtBuilder::default().build().execute_with(|| {
+			use pallet_authorship::EventHandler;
+
+			assert_eq!(<pallet_authorship::Module<TestRuntime>>::author(), 11);
+
+			<Module<TestRuntime>>::note_author(11);
```
jordy25519: nit: consistency 
```suggestion
			Rewards::note_author(11);
```

crml/staking/src/rewards/types.rs
```diff
@@ -42,3 +43,30 @@ pub trait StakerRewardPayment {
...
+	/// Return the reward destination for the given stash account.
+	fn payee(stash: &Self::AccountId) -> Self::AccountId;
+}
+
+/// Counter for the number of "reward" points earned by a given validator.
+pub type RewardPoint = u32;
+
+/// Reward points of an era. Used to split era total payout between validators.
+///
+/// This points will be used to reward validators and their respective nominators.
```
jordy25519:
```suggestion
/// These points will be used to reward validators and their respective nominators.
```

jordy25519 (commented):
Quick first pass, looks good so far 💪
It will need a storage migration to ensure the current payee storage is copied to the rewards module after runtime upgrade or similar solution
Will check tests next

crml/staking/src/lib.rs
aliXsed: Other than in the staking unit tests, it's called in a runtime integration test where the test part of the staking module is not available. In addition to that, I think this is a useful enquiry for the staking module. So I suggest we leave it as it is.

jordy25519: ok, could you leave a comment about its use for test purpose only. with the new (handle payee) trait it seems reward module should be the authority on this for any other modules in terms of public API not this (staking) module

crml/staking/src/rewards/types.rs
```diff
@@ -42,3 +43,30 @@ pub trait StakerRewardPayment {
 	/// i.e calling `enqueue_reward_payouts` would distribute this total value among stakers.
 	fn calculate_next_reward_payout() -> Self::Balance;
 }
+
+pub trait HandlePayee {
+	/// The system account ID type
+	type AccountId;
+
+	/// (Re-)set the payment target for a stash account.
+	/// If payee is not different from stash, do no operations.
```
jordy25519, line 52: nit: the trait doesn't enforce this '/// If payee is not different from stash, do no operations. ' could do something like return a bool to signal whether changes were made

crml/staking/src/rewards/types.rs
```diff
@@ -42,3 +43,30 @@ pub trait StakerRewardPayment {
...
+	fn set_payee(stash: &Self::AccountId, payee: &Self::AccountId);
+	/// Remove the corresponding stash-payee from the look up. Do no operations if stash not found.
```
jordy25519, line 54: same here: 'Do no operations if stash not found'

crml/staking/src/rewards/mod.rs
```diff
@@ -272,6 +311,32 @@ where
 	}
 }
 
+impl<T: Trait> HandlePayee for Module<T> {
+	type AccountId = T::AccountId;
+
+	/// (Re-)set the payment target for a stash account.
+	/// If payee is not different from stash, do no operations.
+	fn set_payee(stash: &Self::AccountId, payee: &Self::AccountId) {
```
jordy25519, line 317: needs some unit tests

crml/staking/src/tests.rs
```diff
@@ -208,45 +207,31 @@ fn rewards_should_work() {
...
 		assert_eq!(Session::current_index(), 1);
-		<Module<Test>>::reward_by_ids(vec![(11, 50)]);
-		<Module<Test>>::reward_by_ids(vec![(11, 50)]);
+
+		Rewards::reward_by_ids(vec![(11, 50)]);
+		Rewards::reward_by_ids(vec![(11, 50)]);
 		// This is the second validator of the current elected set.
-		<Module<Test>>::reward_by_ids(vec![(21, 50)]);
-		// This must be no-op as it is not an elected validator.
-		<Module<Test>>::reward_by_ids(vec![(1001, 10_000)]);
```
jordy25519, line 226: did you mean to remove this?

crml/staking/src/tests.rs
```diff
@@ -208,45 +207,31 @@ fn rewards_should_work() {
 			},
 		);
 
-		<Payee<Test>>::insert(&2, RewardDestination::Stash);
-		assert_eq!(Staking::payee(2), RewardDestination::Stash);
-		assert_eq!(Staking::payee(11), RewardDestination::Account(10));
+		assert_eq!(Staking::payee(&2), RewardDestination::Stash);
+		assert_eq!(Staking::payee(&11), RewardDestination::Controller);
+
+		Rewards::new_fiscal_era();
```
jordy25519: this would be preferred as a unit test, fiscal era start should be a by product
```suggestion
		start_era(0)
```

crml/staking/src/tests.rs
```diff
@@ -267,26 +252,31 @@ fn multi_era_reward_should_work() {
 		assert_ok!(Staking::set_payee(Origin::signed(10), RewardDestination::Controller));
 
 		// Compute now as other parameter won't change
-		let total_payout_0 = current_total_payout::<Balances>();
+		Rewards::new_fiscal_era();
```
jordy25519:
```suggestion
		start_era(0);
```
```

File: crml/staking/src/tests.rs [hunk 15]
```diff
@@ -267,26 +252,31 @@ fn multi_era_reward_should_work() {
...
 		start_session(2);
 		start_session(3);
 
+		Staking::on_initialize(System::block_number() + 1);
+
 		assert_eq!(Staking::current_era(), 1);
 		assert_eq!(Balances::total_balance(&10), init_balance_10 + total_payout_0);
 
-		start_session(4);
+		start_session(5);
```
jordy25519, line 268: wondering why do the sessions need to change here vs. before?

crml/staking/src/lib.rs
```diff
@@ -1688,59 +1599,38 @@ impl<T: Trait> Module<T> {
...
 			_ => ForceEra::put(Forcing::ForceNew),
 		}
 	}
+
+	/// Return the reward destination for a given stash id.
+	///
+	/// This is mostly a helper function for test purposes. There is a simpler interface,
+	/// HandlePayee, implemented by the rewards module which is suggested to look at first before
+	/// using this one in your code.   
+	pub fn payee(stash: &T::AccountId) -> RewardDestination<T::AccountId> {
```
jordy25519: on 2nd thought, this should be marked `#[cfg(test)]`.  It's only used here in integration tests: https://github.com/cennznet/cennznet/blob/29b9a214588f704c17b90a554cb198adff17c87b/runtime/tests/staking_reward.rs#L139 which should be replaced with `Rewards::payee(stash)`

crml/staking/src/lib.rs
```diff
@@ -1688,59 +1599,38 @@ impl<T: Trait> Module<T> {
...
+		let payee = T::Rewarder::payee(stash);
+		if payee == *stash {
+			return RewardDestination::Stash;
+		}
+		if let Some(controller) = Bonded::<T>::get(stash) {
+			if payee == controller {
+				return RewardDestination::Controller;
+			}
+		}
+		RewardDestination::Account(payee)
```
jordy25519: this is the existing behaviour and should remain, controller always needs 2 look ups so we phased it out already on v37 by disallowing it from being created in the first place.
```suggestion
		let payee = T::Rewarder::payee(stash);
		if payee == *stash {
			return RewardDestination::Stash;
		}
		RewardDestination::Account(payee)
```

crml/staking/src/tests.rs
```diff
@@ -208,45 +207,31 @@ fn rewards_should_work() {
 			},
 		);
 
-		<Payee<Test>>::insert(&2, RewardDestination::Stash);
-		assert_eq!(Staking::payee(2), RewardDestination::Stash);
-		assert_eq!(Staking::payee(11), RewardDestination::Account(10));
+		assert_eq!(Staking::payee(&2), RewardDestination::Stash);
+		assert_eq!(Staking::payee(&11), RewardDestination::Controller);
```
jordy25519: ~change assertion back to `RewardDestination::Account(10)` re discussion above (RewardDestination::Controller is phased out)~

crml/staking/src/tests.rs
```diff
@@ -3226,20 +3181,22 @@ fn payout_to_any_account_works() {
 		bond_nominator(1234, 1337, 100, vec![11]);
 
 		// Update payout location
-		assert_ok!(Staking::set_payee(Origin::signed(1337), RewardDestination::Account(42)));
+		assert_ok!(Staking::set_payee(Origin::signed(1337), RewardDestination::Account(41)));
```
jordy25519: why change the account values here, seems ok as it was?

crml/staking/src/mock.rs
```diff
@@ -605,26 +539,22 @@ pub(crate) fn bond_nominator(stash: AccountId, ctrl: AccountId, val: Balance, ta
...
-pub fn advance_session() {
-	let current_index = Session::current_index();
-	start_session(current_index + 1);
-}
+pub fn start_session(index: SessionIndex) {
+	assert!(Session::current_index() <= index);
 
-pub fn start_session(session_index: SessionIndex) {
-	// Compensate for session delay
-	let session_index = session_index + 1;
```
jordy25519, line 615: what was the reason for changing this? now there is lot's of house keeping in the tests, not sure if it was relevant or this PR

runtime/tests/staking_reward.rs
```diff
@@ -211,19 +211,13 @@ fn current_era_transaction_rewards_storage_update_works() {
...
 #[test]
-fn elected_validators_receive_equal_transaction_fee_reward() {
+fn elected_validators_receive_transaction_fee_according_to_authorship_points() {
 	// Make some txs
 	// Start a new era to payout last eras validators
 	// Check payouts happen as expected and total issuance is maintained
 	let validators = make_authority_keys(6);
 	let initial_balance = 100_000_000 * DOLLARS;
 	let staked_amount = initial_balance / validators.len() as Balance;
-	let transfer_amount = 50;
```
jordy25519: it's important we still test actual transaction fees are recorded and distributed here, maybe you forgot to add this stuff back into the test

runtime/tests/staking_reward.rs
```diff
@@ -271,42 +253,36 @@ fn elected_validators_receive_equal_transaction_fee_reward() {
...
+					);
+				} else {
+					assert_eq!(
+						RewardCurrency::free_balance(&stash),
+						initial_balance + per_validator_reward_era_1
+					);
+				}
 			}
 
-			// treasury gets it's cut
```
jordy25519, line 290: should still check treasury cut here

crml/staking/src/rewards/types.rs
aliXsed: Returning a bool would be confusing as a no-operation case is just to referring to the performance of the operation. It's not saying for example that the payee is not set or the operation has failed.

crml/staking/src/rewards/types.rs
aliXsed: Same as my comment above

crml/staking/src/rewards/mod.rs
aliXsed: Done

crml/staking/src/tests.rs
aliXsed: Yes. As this functionality is moved to the rewards module, there we don't have access to the listed of the elected validators. So we don't check it. However, as reward_by_ids is only called by note_author and note_uncle which are called by the authorship modules only for the elected validators, this should be okay practically.

crml/staking/src/tests.rs
alexsednz: Sorry which discussion?

crml/staking/src/tests.rs
aliXsed: start_era(0) couldn't do the job because the era 0 and session 0 is already configured and thus start_era(0) does nothing. I moved the very first new_fiscal_era to the genesis mocker for the tests instead.

crml/staking/src/mock.rs
aliXsed: Because it has wrongly coupled the concept of a block number to session number. The right way of changing to a particular session number is through rotating sessions. This was inconsistent with the tests that we have in the runtime modules as well. The session_index + 1 that was needed should have indicated that this wasn't correct.

aliXsed: There wasn't a lot of housekeeping and I have already done that too. There are a few places where we need to start a block as well which is better to be done as a step in the test to make it clear what's happning.

crml/staking/src/lib.rs
aliXsed: This function is only used for tests. There's no need to optimise it. RewardDestination::Controller is a better reflection of the RewardDestination enum and says more than RewardDestination::Account().

crml/staking/src/lib.rs
aliXsed: Alright I will do that

crml/staking/src/tests.rs
aliXsed: This could stay as before but the next one which is triggering a new era should be start_session(6). The era length is 3, so sessions `0, 1, 2 => era 0` and sessions 3, 4, 5 => era 1, session 6 => era 2. This was working before because start_session was wrong.

crml/staking/src/tests.rs
aliXsed: Account 42 is not created in the genesis config. When paying rewards we do deposit_into_existing which means if the account does exist we pay the reward. It was okay before because in the function make_payout for the mocker we had instead done deposit_creating. But now instead of the mock we are using the actual Rewards code.

runtime/tests/staking_reward.rs
aliXsed: We do test applying an extrinsic and its effect on the transaction fee pot in current_era_transaction_rewards_storage_update_works. Here the challenge that I had was that we do need to start a block for the amortised payments to happen. However, starting a new block affects the authorship reward points and the distribution of the next rewards. To keep the whole test simple and solid, I just allowed the authorship to happen in a controlled way only on Executive::initialize_block.

runtime/tests/staking_reward.rs
aliXsed: With the new arrangement of this test, the tx_fee is zero. I can add an individual test for the treasury cut and keep this one simple. I think simpler single-purpose tests are better maintainable and can help us more. So I prefer adding another test personally. What's your opionion?

crml/staking/src/lib.rs
jordy25519: it's not optimise, it tests something which doesn't happen in reality now

runtime/tests/staking_reward.rs
jordy25519: this was testing that transaction fees from a block are paid out to stakers. It doesn't make sense to remove this, you need to make another test

runtime/tests/staking_reward.rs
jordy25519: I'd say keep the original behaviour of the test and check authorship points also

runtime/tests/staking_reward.rs
aliXsed: I will add another test for checking the transaction fee is paid to the stakers and the treasury gets its cut.

runtime/tests/staking_reward.rs
aliXsed: The problem I had in this test was stemmed from Executive::apply_extrinsic not triggering note_author. So as said above, I'm going to add a separate test for this similar to the previous test here which is not checking for points at the same time.

crml/staking/src/rewards/mod.rs
```diff
@@ -1090,4 +1204,46 @@ mod tests {
 			);
 		});
 	}
+
+	#[test]
+	fn reward_from_authorship_event_handler_works() {
+		ExtBuilder::default().build().execute_with(|| {
+			use pallet_authorship::EventHandler;
```
jordy25519: can remove this import

crml/staking/src/lib.rs
aliXsed: Even RewardDestination::Account(payee) doesn't happen in reality since we started to handle payees in the rewards module. It's just an interface that we have now made it limited to tests. The staking module gets the reward destination through the RewardDestination enum and it's nice to give it back in the same way.

crml/staking/src/tests.rs
```diff
@@ -794,14 +777,14 @@ fn reward_destination_controller_is_replaced_with_account() {
 			Staking::minimum_bond(),
 			RewardDestination::Controller
 		));
-		assert_eq!(Staking::payee(stash), RewardDestination::Account(controller));
+		assert_eq!(Staking::payee(&stash), RewardDestination::Controller);
```
jordy25519: no need to change this

crml/staking/src/tests.rs
```diff
@@ -794,14 +777,14 @@ fn reward_destination_controller_is_replaced_with_account() {
...
 
 		// explicit set controller as payee
 		assert_ok!(Staking::set_payee(
 			Origin::signed(controller),
 			RewardDestination::Controller
 		));
-		assert_eq!(Staking::payee(stash), RewardDestination::Account(controller));
+		assert_eq!(Staking::payee(&stash), RewardDestination::Controller);
```
jordy25519: same

crml/staking/src/mock.rs
jordy25519: I think it's harder to understand the tests now, its easier if it always does the same thing but now the individual tests sometimes do session increment and sometimes block number increment and sometimes start another era and it's not clear why. It looks upstream like this now: https://github.com/paritytech/substrate/blob/9c5da5cc68e710d4d44b6bbeea6fb5a5e49b7134/frame/staking/src/mock.rs#L617-L628 which is easier to follow

crml/staking/src/tests.rs
jordy25519: ok this seems like an issue in the reward function since we don't limit the reward destination to existing accounts only

crml/staking/src/lib.rs
jordy25519: Ok RewardDestination is just a legacy thing for us at this point. We accept it as an input to some staking calls but it is never stored. The tests should assert on the stored AccountId directly

crml/staking/src/tests.rs
jordy25519: check the thing we actually store
```suggestion
		assert_eq!(<<Staking as Trait>::Rewarder as HandlePayee>::payee(&11), 11)
```

crml/staking/src/mock.rs
aliXsed: start_era and start_sessions together are used in about 100 times in our tests. However, there are only 12 places where we needed to start a block as well. The reason for starting a block can be mentioned as a comment in the test and I am happy to do that. In our staking module design, when an era starts, the rewards are enqueued. However, the rewards are only paid on subsequent blocks. Mixing these different functions (starting and authoring a block + starting a session) is not giving us the fine control of the environment that we need for tests. Starting a block affects the authorship as well, which we should do on clear intentions.
At the end of the day, start_era and start_session are helper functions inside the test. You can do lots of things in a function and name it whatever you want. However, best design practices recommend 1. Good appropriate names 2. [Single responsibility](https://en.wikipedia.org/wiki/Single-responsibility_principle)

crml/staking/src/tests.rs
aliXsed: Do you mean in Staking::set_payee and Staking::bond, we should check if the destination account exists? This is a good idea. However, we still need to change 42 to 41 in this test.

crml/staking/src/rewards/mod.rs

crml/staking/src/tests.rs
aliXsed: Done

crml/staking/src/tests.rs
aliXsed: The reward function issue is fixed according to our chat on slack.

jordy25519: I think this (test) can change back now after the `deposit_creating` change

aliXsed: Done

jordy25519 (approved):

PR #366, opened 2021-02-04
Feature/365 rewards rpc

### LINKED ISSUE #365: Staking rewards custom RPC
Add a custom RPC to the staking module which returns the currently accrued payout for a given stash id.

aliXsed:
Closes #365. The implementation is reusing the same logic for calculating and enqueuing the actual payouts at the end of an era. Since that logic is being tested well already, The function to calculate the accrued rewards is also already tested to a good level. I will, however, add some tests specific to this function.


crml/staking/rpc/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+
+		if at.is_some() {
+			return Err(RpcError {
+				code: ErrorCode::ServerError(Error::UnsupportedError.into()),
+				message: "Unsupported query when block hash is given.".into(),
+				data: None,
+			});
+		}
+
+		let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
```
jordy25519, line 98: We should support `at` after the runtime upgrade block, maybe check runtime version and fail for pre v38

crml/staking/src/rewards/mod.rs
```diff
@@ -307,6 +276,26 @@ where
 	fn calculate_next_reward_payout() -> Self::Balance {
 		TransactionFeePot::<T>::get().saturating_add(Self::target_inflation_per_staking_era())
 	}
+
+	/// Calculate the next reward payout (as of accrued right now) for the given payee.
```
```suggestion
	/// Calculate the next reward payout (accrued as of right now) for the given payee.
```

crml/staking/src/rewards/types.rs
```diff
@@ -42,6 +42,11 @@ pub trait StakerRewardPayment {
 	/// Calculate the value of the next reward payout as of right now.
 	/// i.e calling `enqueue_reward_payouts` would distribute this total value among stakers.
 	fn calculate_next_reward_payout() -> Self::Balance;
+	/// Calculate the next reward payout (as of accrued right now) for the given payee.
```
jordy25519:
```suggestion
	/// Calculate the next reward payout (accrued as of right now) for the given payee.
```

crml/staking/src/rewards/mod.rs
```diff
@@ -335,6 +324,71 @@ impl<T: Trait> HandlePayee for Module<T> {
...
+impl<T: Trait> Module<T>
+where
+	BalanceOf<T>: FixedPointOperand,
+{
+	/// Calculate all payouts of the current era as of right now. Then filter out those not relevant
+	/// validator-exposure sets by calling the "filter" function and execute the on_each_payout
+	/// function for each of non filtered payouts.
+	/// Return the total rewards calculated for the stakers at the time of this call paired with the
+	/// reward the development cut.
+	fn on_all_payouts<F, A>(
```
jordy25519: it's not really a hook 'on_all_payouts', seems to be still`calculate_payouts` with some additional config
would be good to include a doc comment example on usage 
```
/// # Example
``
let filter = ...
let on_payout = ...
let (staker_rewards, treasury_reward) = on_all_payouts<F, A>(filter, on_payout);
``
```

crml/staking/src/rewards/mod.rs
```diff
@@ -187,61 +187,30 @@ where
...
+			validator_commission_stake_map,
+			|_, _| false,
+			|account, payout| {
+				Payouts::<T>::mutate(|p| p.push_back((account.clone(), payout, era)));
+			},
+		);
 
-		// Deduct development fund take %
-		let development_fund_payout = Self::development_fund_take() * fee_payout;
+		// Consume the reward points what are used in the previous stage
```
```suggestion
		// Reset the reward points storage for the next era
```

crml/staking/src/rewards/mod.rs
```diff
@@ -307,6 +276,26 @@ where
...
+	/// Calculate the next reward payout (as of accrued right now) for the given payee.
+	fn payee_next_reward_payout(
+		payee: &Self::AccountId,
+		validator_commission_stake_map: &[(Self::AccountId, Perbill, Exposure<Self::AccountId, Self::Balance>)],
+	) -> Self::Balance {
+		let mut payee_cut: Self::Balance = Zero::zero();
+
+		let _ = Self::on_all_payouts(
+			validator_commission_stake_map,
+			|validator, exposure| payee != validator && !exposure.others.iter().any(|x| &x.who == payee),
```
jordy25519: validators always nominate themselves so I guess we don't need the `payee != validator` clause

crml/staking/src/rewards/mod.rs
```diff
@@ -335,6 +324,71 @@ impl<T: Trait> HandlePayee for Module<T> {
...
+	/// Calculate all payouts of the current era as of right now. Then filter out those not relevant
+	/// validator-exposure sets by calling the "filter" function and execute the on_each_payout
+	/// function for each of non filtered payouts.
+	/// Return the total rewards calculated for the stakers at the time of this call paired with the
+	/// reward the development cut.
+	fn on_all_payouts<F, A>(
+		validator_commission_stake_map: &[(T::AccountId, Perbill, Exposure<T::AccountId, BalanceOf<T>>)],
+		filter: F,
+		mut on_each_payout: A,
+	) -> (BalanceOf<T>, BalanceOf<T>)
```
```suggestion
	fn calculate_payouts_filtered<F, A>(
		validator_commission_stake_map: &[(T::AccountId, Perbill, Exposure<T::AccountId, BalanceOf<T>>)],
		filter: Option<F>,
	) -> (BalanceOf<T>, BalanceOf<T>, Iterator<Item=(AccountId, Balance))
```

This API could be improved by leaving iteration of payouts for the caller to handle.
Means caller doesn't need to know the details of this function and easier to follow logic at the call site:
```
let (_, _, payouts) = calculate_payouts_filtered(stake_map, None);
payouts.foreach(|account, amount| {
  // do whatever
});

let (_, _, payouts) = calculate_payouts_with_filtered(stake_map, Some(|v, e| ... ));
payouts.foreach(|account, amount| {
  // do whatever
});
```

crml/staking/src/rewards/mod.rs
```diff
@@ -1228,6 +1282,90 @@ mod tests {
 		})
 	}
 
+	#[test]
+	fn add_reward_points_fns_works() {
```
jordy25519, line 1285: can you add a test for validator checking reward?
recommend using some var names for the stash account IDs instead of `21` and `11` etc.

crml/staking/src/lib.rs
```diff
@@ -1322,6 +1322,22 @@ impl<T: Trait> Module<T> {
 			.unwrap_or_default()
 	}
 
+	/// Calculate the next accrued payout for a stash id (a payee) without affecting the storage.
+	pub fn accrued_payout(stash: &T::AccountId) -> BalanceOf<T> {
+		let validator_commission_stake_map = Self::current_elected()
```
jordy25519, line 1382: could be worth caching this value per era/session

crml/staking/src/rewards/mod.rs
```diff
@@ -187,61 +187,30 @@ where
...
 
-		// Deduct development fund take %
-		let development_fund_payout = Self::development_fund_take() * fee_payout;
+		// Consume the reward points what are used in the previous stage
+		<CurrentEraRewardPoints<T>>::kill();
+
+		// track historic era fee amounts
+		Self::note_fee_payout(TransactionFeePot::<T>::take());
 
 		// implementation note: imbalances have the side affect of updating storage when `drop`ped.
```
jordy25519: just noticed we can remove this comment now

crml/staking/rpc/src/lib.rs
aliXsed: For runtime versions pre v38 if we query this rpc we get:
`{"jsonrpc":"2.0","error":{"code":1,"message":"Unable to accrued payout.","data":"Execution(Wasmi(Function(\"Module doesn\\'t have export StakingApi_accrued_payout\")))"},"id":"1"}
`
And it seems the node operation proceeds normally.

crml/staking/src/rewards/mod.rs
aliXsed: Calculate is a passive verb for what it does. It's able to call a mut fn on every payout. As I have mentioned in the doc, it calculates all the payouts, filter some and execute a function on every one of them. This is a more generalised version of what we had inside `enqueue_reward_payouts` so I can reuse the same logic for calculating payee_next_reward_payout. I will add an example to the doc comment, will be open to new name suggestions.

crml/staking/src/rewards/mod.rs
aliXsed: Looks like even though it's conceptually true that a validator nominate itself, they would not appear in the `other` field of the exposure: https://github.com/cennznet/cennznet/blob/develop/crml/staking/src/lib.rs#L1541-L1545

crml/staking/src/rewards/mod.rs
aliXsed: I have refactored this function according to your suggestion except for Option<F> which could make the implementation harder to read with not much performance improvement.

crml/staking/src/rewards/mod.rs
aliXsed: This test, in particular, is just moved in this PR. However, regarding your suggestion for a validator checking reward, it seems to me it's covered in the test calculate_accrued_reward. There stash id 1 is a validator.

crml/staking/src/rewards/mod.rs

crml/staking/src/lib.rs
aliXsed: Can you elaborate on your suggestion here?

jordy25519: so everytime the RPC is called it will run this calculation, the value of `validator_commission_stake_map` won't change unless era changes or session to be safe e.g. incase of slash.
Seems could be nicer for the node to cache it

aliXsed: It would then mean adding another storage item to the staking module, right? Can we do this optimisation later based on the actual load of this RPC on our nodes?

jordy25519: Yes not worth blocking the PR over but could see it getting intensive for full nodes: we'd try this: https://substrate.dev/docs/en/knowledgebase/learn-substrate/off-chain-features#off-chain-indexing not onchain

crml/staking/src/rewards/mod.rs
jordy25519: test is not very nice with all the ids instead of var names but that's leftover from before

aliXsed: Updated this test

crml/staking/rpc/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+where
+	Block: BlockT,
+	C: 'static + ProvideRuntimeApi<Block> + HeaderBackend<Block>,
+	C::Api: StakingRuntimeApi<Block, AccountId>,
+	AccountId: Codec,
+{
+	fn accrued_payout(&self, payee: AccountId, at: Option<<Block as BlockT>::Hash>) -> Result<u64> {
+		let api = self.client.runtime_api();
+
+		if at.is_some() {
```
jordy25519, line 90: We should support `at` so this check can be removed

jordy25519 (approved):
lgtm! apart from the rpc `at` thing

crml/staking/rpc/src/lib.rs
aliXsed: I created https://github.com/cennznet/cennznet/issues/370 for this part.

PR #367, opened 2021-02-11
Configure BondingDuration and SlashDeferDuration for 24 hour eras

### LINKED ISSUE #362: Adjust CENNZnet bonding duration
The current value is 24 * 28 eras which with 1-hour eras could be as long as 28 days. Now that we have changed the era to be 24 hours long, we need to readjust this parameter.

jordy25519:
\+ Driveby clean up to remove `BlockBonding`

closes #362


crml/staking/src/lib.rs
```diff
@@ -814,8 +814,6 @@ decl_error! {
 		NotSortedAndUnique,
 		/// Cannot nominate the same account multiple times
 		DuplicateNominee,
-		// TODO remove the following error when https://github.com/cennznet/cennznet/issues/297
-		BondingNotEnabled,
```
aliXsed, line 878: Didn't we want to use the sudo command to enable bonding?

jordy25519: it's already live on mainnet, it won't be needed after (and including) the next update

aliXsed: So please remove the corresponding storage as well:
`BlockBonding get(fn block_bonding) config(): bool;`
In fact, we should kill the storage in the runtime upgrade as well. I can actually do that in https://github.com/cennznet/cennznet/pull/368 if you prefer.

aliXsed: Now that https://github.com/cennznet/cennznet/pull/368 is merged, can you also please do `inner::BlockBonding::kill();` after merging develop into your branch?

aliXsed (approved):

PR #369, opened 2021-02-15
Make sylo modules readonly

### LINKED ISSUE #347: Legacy sylo modules readonly
Sylo have moved their app logic to their new Oya network.
We can make the sylo modules readonly (i.e. remove their transaction interfaces).
They still  need some of the query interfaces on mainnet to allow user migration.

jordy25519:
Keeps the storage items for js API discoverability but removes all extrinsics, events, errors

at somepoint we should issue a kill tx to remove these from state once Sylo is done with them

closes #347


runtime/src/lib.rs
```diff
@@ -613,16 +600,8 @@ construct_runtime!(
 		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Module, Storage} = 19,
 		Historical: session_historical::{Module} = 20,
 		Cennzx: crml_cennzx::{Module, Call, Storage, Config<T>, Event<T>} = 21,
-		// TODO: these should all be in one module
-		SyloGroups: sylo_groups::{Module, Call, Storage} = 22,
```
aliXsed, line 619: Is this necessary?

runtime/src/lib.rs
```diff
@@ -613,16 +600,8 @@ construct_runtime!(
...
-		SyloE2EE: sylo_e2ee::{Module, Call, Storage} = 23,
-		SyloDevice: sylo_device::{Module, Call, Storage} = 24,
-		SyloInbox: sylo_inbox::{Module, Call, Storage} = 25,
-		SyloResponse: sylo_response::{Module, Call, Storage} = 26,
-		SyloVault: sylo_vault::{Module, Call, Storage} = 27,
-		SyloPayment: sylo_payment::{Module, Call, Storage} = 28,
-		Attestation: prml_attestation::{Module, Call, Storage, Event<T>} = 29,
-		Rewards: crml_staking_rewards::{Module, Call, Storage, Config, Event<T>} = 30,
```
jordy25519: I think I need to re-include these for the `Module + Storage`

aliXsed (approved):
LGTM! We should just remember to involve Sylo in testing this particular change on Nikau before releasing it to Azalea.

PR #372, opened 2021-02-16
Fix set_payee

aliXsed:
Fix the case where the payee has been set to an account other than the stash and now needs to be set back to stash.

Also, Improve the current unit test to be able to catch the above issue.


KarishmaBothara (approved):
lgtm as non rust expert

jordy25519 (approved):

PR #375, opened 2021-02-17
Prerelease/1.3.0

aliXsed:
As it says on the tin ^

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -80,10 +80,11 @@ services:
     - ./data/node-5:/mnt/data
     command:
       - --chain=/cennznet/genesis/rata.raw.json
-      - --name=full-${HOSTNAME:-localhost}-5
+      - --validator
+      - --rata
       - --base-path=/mnt/data
-      - --node-key=8967bcbae338641d1102ff4465fda281f0bf6e9842702f3a992650d4855f5cd0
-      - --ws-external
-      - --rpc-external
+      - --name=rata-${HOSTNAME:-localhost}-5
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -80,10 +80,11 @@ services:
...
+      - --validator
+      - --rata
       - --base-path=/mnt/data
-      - --node-key=8967bcbae338641d1102ff4465fda281f0bf6e9842702f3a992650d4855f5cd0
-      - --ws-external
-      - --rpc-external
+      - --name=rata-${HOSTNAME:-localhost}-5
+      - --node-key=0000000000000000000000000000000000000000000000000000000000000006
+      - --bootnodes=/dns4/node-5/tcp/30333/p2p/12D3KooWCjJYJPjLJmHH9Q7NGzQRZvN28xwZ4Rjsazf14sCr3uH8
     ports:
```
jordy25519: can you change this back, it was a full node + bootnode, now it's a validator with a different node-key
also I don't think `--chain=/cennznet/genesis/rata.raw.json` and `--rata` should be used together probably keep `--chain` like the other nodes

jordy25519 (approved):
lgtm apart from the docker-compose change

docker-compose.yml
aliXsed: I guess --rata was supposed to say which account is linked to this validator similar to other nodes (--alice --bob, ...), For having a full node I can remove (--validator --rata). But the issue that we had for the upgrade was related to the line for the bootnodes which wasn't specified for this one. How do you make this node a boot node?

PR #380, opened 2021-02-23
Get payee from stash id

### LINKED ISSUE #379: Fix accruedPayout rpc when reward destination is not stash
accruedPayout only returns the payout correctly if the stash and the reward account are the same.

aliXsed:
Closes #379

### DIFF
--- a/crml/staking/src/rewards/mod.rs
+++ b/crml/staking/src/rewards/mod.rs
@@ -273,19 +273,19 @@ where
 		TransactionFeePot::<T>::get().saturating_add(Self::target_inflation_per_staking_era())
 	}
 
-	/// Calculate the next reward payout (accrued as of right now) for the given payee.
+	/// Calculate the next reward payout (accrued as of right now) for the given stash id.
 	fn payee_next_reward_payout(
-		payee: &Self::AccountId,
+		stash: &Self::AccountId,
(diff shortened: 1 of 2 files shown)


jordy25519:
Please add a unit test to show the fixed behaviour

I think part of the confusion is the language used throughout RPC is `payee` when it's really the `stash` account, would recommend this is reviewed also

real issue is here: exposure requires stash
https://github.com/cennznet/cennznet/blob/565fd6be74ff025ae9efdd78740a19540f491d95/crml/staking/src/rewards/mod.rs#L283-L291

crml/staking/src/rewards/mod.rs
```diff
@@ -273,19 +273,19 @@ where
...
 		let mut payee_cut: Self::Balance = Zero::zero();
 
 		let (_, _, payouts) =
 			Self::calculate_payouts_filtered(validator_commission_stake_map, |validator, exposure| {
-				payee != validator && !exposure.others.iter().any(|x| &x.who == payee)
+				stash != validator && !exposure.others.iter().any(|x| &x.who == stash)
 			});
 		payouts.into_iter().for_each(|(account, payout)| {
-			if &account == payee {
+			if account == Self::payee(stash) {
```
jordy25519: should query this once outside the loop and reuse the variable here

jordy25519 (approved):
lgtm apart from the one comment

crml/staking/src/rewards/mod.rs

PR #391, opened 2021-03-11
Chore/update to plug 3.0

aliXsed:
This is the first brick in creating Cennznet 2.0. Next stage would be "enabling smart contracts". However we need to merge this with develop first.


Cargo.toml
```diff
@@ -89,9 +89,6 @@ zeroize = { opt-level = 3 }
 
 [features]
 default = []
-quick_staking = [
```
jordy25519: keep this please, it's for building a client with quick staking sessions/eras

cli/Cargo.toml
```diff
@@ -15,77 +15,75 @@ futures = { version = "0.3.1", features = ["compat"] }
...
 [build-dependencies]
-substrate-build-script-utils = { rev = "2.0.5", git = "https://github.com/plugblockchain/plug-blockchain" }
+substrate-build-script-utils = { rev = "v3.0.0-rc4", git = "https://github.com/plugblockchain/plug-blockchain" }
 
 [features]
 default = []
-# Uses integration test config for quicker staking eras + epochs
-quick_staking = [
-	"cennznet-runtime/integration_config"
-]
```
jordy25519: keep pls

crml/cennzx/src/mock.rs
```diff
@@ -25,87 +25,99 @@ pub const TRADE_ASSET_B_ID: AssetId = 3;
...
 	type OnKilledAccount = ();
 	type SystemWeightInfo = ();
+	type SS58Prefix = ();
 }
 
-impl prml_generic_asset::Trait for Test {
+parameter_types! {
+		pub const TreasuryModuleId: ModuleId = ModuleId(*b"py/trsry");
+}
+pub struct TransferImbalanceToTreasury;
```
jordy25519, line 98: could you add a comment on why this is implemented/used for this test runtime

crml/transaction-payment/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+//! Runtime API definition for transaction payment module.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::Codec;
+use sp_runtime::traits::MaybeDisplay;
+
+pub use crml_transaction_payment::{FeeDetails, InclusionFee, RuntimeDispatchInfo};
+
+// TODO Fix conflicting implementations of trait `parity_scale_codec::Decode` for type `pallet_transaction_payment::RuntimeDispatchInfo<_>`:
```
jordy25519: will you open an issue or address in this PR? I think these changes should be added to `transaction-payment/src/types.rs`

runtime/src/lib.rs
```diff
@@ -126,28 +137,66 @@ pub fn native_version() -> NativeVersion {
...
+	fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance>) {
+		if let Some(fees) = fees_then_tips.next() {
+			Rewards::note_transaction_fees(fees.peek());
+			if let Some(tips) = fees_then_tips.next() {
+				Rewards::note_transaction_fees(tips.peek());
+			}
+		}
+	}
+}
+
```
jordy25519: should go to impls.rs

runtime/src/lib.rs
```diff
@@ -126,28 +137,66 @@ pub fn native_version() -> NativeVersion {
...
 // Configure modules to include in the runtime.
 
-const AVERAGE_ON_INITIALIZE_WEIGHT: Perbill = Perbill::from_percent(10);
+/// We assume that ~10% of the block weight is consumed by `on_initialize` handlers.
+/// This is used to limit the maximal weight of a single extrinsic.
+const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
+/// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used
+/// by  Operational  extrinsics.
+const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
+/// We allow for 2 seconds of compute with a 6 second average block time.
```
jordy25519:
```suggestion
/// We allow for 2 seconds of compute with a 5 second average block time.
```

primitives/src/types.rs
```diff
@@ -64,12 +64,17 @@ pub type Timestamp = u64;
...
+
+/// Block ID.
+pub type BlockId = generic::BlockId<Block>;
+
 /// The outer `FeeExchange` type. It is versioned to provide flexibility for future iterations
 /// while maintaining backward compatibility.
 #[derive(PartialEq, Eq, Clone, Encode, Decode, Debug)]
 pub enum FeeExchange<AssetId, Balance> {
 	/// A V1 FeeExchange
-	#[codec(compact)]
```
jordy25519, line 72: why remove this?

crml/staking/src/lib.rs
```diff
@@ -1901,7 +1901,8 @@ decl_module! {
 		/// # </weight>
 		#[weight = T::WeightInfo::reap_stash(1_u32)]
 		fn reap_stash(_origin, stash: T::AccountId) {
-			ensure!(T::Currency::total_balance(&stash).is_zero(), Error::<T>::FundedTarget);
+			let at_minimum = T::Currency::total_balance(&stash) == T::Currency::minimum_balance();
+			ensure!(at_minimum, Error::<T>::FundedTarget);
```
jordy25519: shouldn't this be
```suggestion
			let at_minimum = T::Currency::total_balance(&stash) <= T::Currency::minimum_balance();
			ensure!(at_minimum, Error::<T>::FundedTarget);
```

crml/staking/src/rewards/mod.rs
```diff
@@ -569,81 +563,86 @@ mod tests {
...
+		type PalletInfo = PalletInfo;
+		type AccountData = prml_generic_asset::AccountData<AssetId>;
 		type OnNewAccount = ();
 		type OnKilledAccount = ();
 		type SystemWeightInfo = ();
+		type SS58Prefix = ();
 	}
 
-	impl prml_generic_asset::Trait for TestRuntime {
+	pub struct TransferImbalanceToTreasury;
```
jordy25519, line 615: could you add a comment on why this is implemented/used for this test runtime

jordy25519 (commented):
Generally lgtm.

This should not merge to `develop` until after NFT launch, recommend using another trunk for that in the meantime.

Some things to check before a '2.0' branch gets deployed live would be (besides smart contracts):
- Review the bounties  + tips config makes sense for mainnet
- Fix decoding historic tx fee amounts

crml/transaction-payment/rpc/runtime-api/src/lib.rs
aliXsed: We have this related issue open here: https://github.com/plugblockchain/plug-blockchain/issues/180. I will check if your suggested solution there would address the issue here.

primitives/src/types.rs
aliXsed: It is removed as part of fixing cennzx build issues. I need to put it back and chack what was the actual build issue.

Cargo.toml
aliXsed: I don't find such a feature in the recent code.

aliXsed: Sure, During merge it was somehow recognised as old code.

cli/Cargo.toml
aliXsed: Will do

crml/staking/src/lib.rs
aliXsed: This code is the same as Plug (and Substrate) and possibly written according to the recent behaviour of the Balances module. Balances wouldn't allow the balance to go below the minimum. For us and with generic asset, I'm not sure it's still the case. So probably safer to check "<=".

primitives/src/types.rs
aliXsed: `error: Invalid attribute on variant, only `#[codec(skip)]` and `#[codec(index = $u8)]` are accepted.`

crml/transaction-payment/rpc/runtime-api/src/lib.rs
aliXsed: Fixed it according to the suggestion.

jordy25519: can clean up this file now

primitives/src/types.rs
jordy25519: Ok I think it's safe to ignore, seems like `enum` doesn't support 'compact' in codec 2.0. (or rather it's already compact)

crml/transaction-payment/rpc/runtime-api/src/lib.rs
aliXsed: The clean up is done as well.

PR #392, opened 2021-03-14
offline phragmén

### LINKED ISSUE #346: Remove finality tracker module
'Finality tracker' is unused and adding intrinsics to the chain we can remove it after some small updates in Plug

### LINKED ISSUE #385: New era delays
New eras are very slow at the moment due to election process, logging reward events also seems to be quite detrimental.
Problem is compounding because our session time is < 10 minutes, resulting in few nodes being flagged as offline
which can trigger a new era again (and around we go).

[block data](https://docs.google.com/spreadsheets/d/1pGyLje8kC8zgIFveF8sYtzuI0MJOTinHORIcjNQHN2I/edit?usp=sharing) from the first occurrence

Fixes:
- [x] Offchain Phragmén
- [x] Split rewards over more blocks
- [ ] Increase session length (help to reduce offline sensitivity)

---
Relevant discussions on dynamic epoch duration changes
https://github.com/galacticcouncil/HydraDX-node/issues/191#issuecomment-810257218
https://github.com/paritytech/substrate/pull/8491#pullrequestreview-624108828

jordy25519:
Changes:
- staking module updated with offline phragmen changes up to 9e2d7b74742f7b1db7e09a00d423f613787d19a8
Screen Shot 2021-03-15 at 9 51 43

  - adds a limit to the number of nominators which can be rewarded per validator (128)
  - elections will run offchain in the last session of an era, new nominations/active stake changes close during this time
  - 'current era' is now the 'scheduled era' which will occur next, 'active era' is the real 'current era'
  - forced eras now accrue tx fee rewards for the next natural era. inflation reward doesn't carry over. (needs inflation curve to do this properly) 
- rewards scheduling updated to follow 'lazy payouts' but at scheduled intervals / consume less i/o

Adds:
- `cargo build --features=quick_staking` for shorter epoch duration and sessions per era

Removes:
- sp-finality-tracker module was unused and is removed here also (closes #346)

mostly resolves #385


crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+		.map_err(|_| OffchainElectionError::PoolSubmissionFailed)
+}
+
+/// Get a random number of iterations to run the balancing.
+///
+/// Uses the offchain seed to generate a random number.
+pub fn get_balancing_iters<T: Trait>() -> usize {
+	match T::MaxIterations::get() {
+		0 => 0,
+		max @ _ => {
```
aliXsed, line 136: `max @ _` -> `max`

crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+			Ok(next) => return next,
+		}
+		step = step / 2;
+		current_weight = weight_with(voters);
+	}
+
+	// Time to finish.
+	// We might have reduced less than expected due to rounding error. Increase one last time if we
+	// have any room left, the reduce until we are sure we are below limit.
+	while voters + 1 <= max_voters && weight_with(voters + 1) < max_weight {
```
aliXsed, line 201: `voters + 1 <= max_voters` -> `voters < max_voters`

crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+	if size.nominators < 1 {
+		return size.nominators;
+	}
+
+	let max_voters = size.nominators.max(1);
+	let mut voters = max_voters;
+
+	// helper closures.
+	let weight_with = |voters: u32| -> Weight {
+		W::submit_solution_better(size.validators.into(), size.nominators.into(), voters, winners_len)
```
aliXsed, line 161: consider removing `.into()` for nominators: `size.nominators`

crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+	let compact = trim_to_weight::<T, _>(maximum_allowed_voters, compact, &nominator_index)?;
+
+	// re-compute the score. We re-create what the chain will do. This is a bit verbose and wastes
+	// CPU time, but it is necessary to ensure that the score that we claim is the same as the one
+	// calculated by the chain.
+	let score = {
+		let compact = compact.clone();
+		let assignments = compact.into_assignment(nominator_at, validator_at).unwrap();
+		let staked =
+			sp_npos_elections::assignment_ratio_to_staked(assignments.clone(), <Module<T>>::slashable_balance_of_fn());
```
aliXsed, line 371: redundant clone for `assignments`

crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+	let mut staked = sp_npos_elections::assignment_ratio_to_staked(assignments, <Module<T>>::slashable_balance_of_fn());
+
+	// reduce
+	if do_reduce {
+		reduce(&mut staked);
+	}
+
+	// Convert back to ratio assignment. This takes less space.
+	let low_accuracy_assignment =
+		sp_npos_elections::assignment_staked_to_ratio_normalized(staked).map_err(|e| OffchainElectionError::from(e))?;
```
aliXsed, line 339: `sp_npos_elections::assignment_staked_to_ratio_normalized(staked).map_err(OffchainElectionError::from)?;`

crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+
+	// compact encode the assignment.
+	let compact = CompactAssignments::from_assignment(low_accuracy_assignment, nominator_index, validator_index)
+		.map_err(|e| OffchainElectionError::from(e))?;
```
aliXsed, line 343: Same as in line 339

crml/staking/src/offchain_election.rs
```diff
@@ -0,0 +1,537 @@
...
+	// potentially reduce the size of the compact to fit weight.
+	let maximum_allowed_voters = maximum_compact_len::<T::WeightInfo>(winners.len() as u32, size, maximum_weight);
+
+	crate::log!(
+		debug,
+		"💸 Maximum weight = {:?} // current weight = {:?} // maximum voters = {:?} // current votes = {:?}",
+		maximum_weight,
+		T::WeightInfo::submit_solution_better(
+			size.validators.into(),
+			size.nominators.into(),
```
aliXsed, line 354: `size.nominators.into()` -> `size.nominators`

crml/staking/src/rewards/mod.rs
```diff
@@ -298,31 +270,118 @@ where
...
+
+		let mut remainder = stakers_cut;
+		let payouts = validators
+			.iter()
+			.map(|validator| {
+				let validator_reward_points = era_reward_points
+					.individual
+					.get(validator)
+					.map(|points| *points)
+					.unwrap_or_else(|| Zero::zero());
```
aliXsed: `.unwrap_or_else(|| Zero::zero());` -> `.unwrap_or_else(Zero::zero);`

crml/staking/src/rewards/mod.rs
```diff
@@ -298,31 +270,118 @@ where
...
+		let total_reward_points = era_reward_points.total;
+
+		let mut remainder = stakers_cut;
+		let payouts = validators
+			.iter()
+			.map(|validator| {
+				let validator_reward_points = era_reward_points
+					.individual
+					.get(validator)
+					.map(|points| *points)
```
aliXsed: `.map(|points| *points)` -> `copied()`

crml/staking/src/lib.rs
```diff
@@ -626,15 +796,27 @@ impl Default for Forcing {
...
+		Releases::V2
+	}
+}
+
+impl From<u32> for Releases {
+	fn from(val: u32) -> Self {
+		match val {
+			0 => Releases::V0,
+			1 => Releases::V1,
+			2 | _ => Releases::V2,
```
aliXsed, line 818: `2 | _` to `_`

crml/staking/src/lib.rs
```diff
@@ -1349,202 +1809,441 @@ decl_module! {
...
+		/// # </weight>
+		#[weight = T::WeightInfo::set_history_depth(*_era_items_deleted)]
+		fn set_history_depth(origin,
+			#[compact] new_history_depth: EraIndex,
+			#[compact] _era_items_deleted: u32,
+		) {
+			ensure_root(origin)?;
+			if let Some(current_era) = Self::current_era() {
+				HistoryDepth::mutate(|history_depth| {
+					let last_kept = current_era.checked_sub(*history_depth).unwrap_or(0);
```
aliXsed, line 1877: `current_era.checked_sub(*history_depth).unwrap_or(0)` -> `current_era.saturating_sub(*history_depth)`

crml/staking/src/lib.rs
```diff
@@ -1349,202 +1809,441 @@ decl_module! {
...
+					let new_last_kept = current_era.checked_sub(new_history_depth).unwrap_or(0);
```
aliXsed, line 1878: `current_era.saturating_sub(new_history_depth)`

crml/staking/src/lib.rs
```diff
@@ -1349,202 +1809,441 @@ decl_module! {
...
+		///    minimized (to ensure less variance)
+		///
+		/// # <weight>
+		/// The transaction is assumed to be the longest path, a better solution.
+		///   - Initial solution is almost the same.
+		///   - Worse solution is retraced in pre-dispatch-checks which sets its own weight.
+		/// # </weight>
+		#[weight = T::WeightInfo::submit_solution_better(
+			size.validators.into(),
+			size.nominators.into(),
```
aliXsed, line 1960: Apparently `.into` is unnecessary here

crml/staking/src/lib.rs
```diff
@@ -1349,202 +1809,441 @@ decl_module! {
...
+		/// Note that this must pass the [`ValidateUnsigned`] check which only allows transactions
+		/// from the local node to be included. In other words, only the block author can include a
+		/// transaction in the block.
+		///
+		/// # <weight>
+		/// See `crate::weight` module.
+		/// # </weight>
+		#[weight = T::WeightInfo::submit_solution_better(
+			size.validators.into(),
+			size.nominators.into(),
```
aliXsed, line 1994: Unnecessary `.into`

crml/staking/src/lib.rs
```diff
@@ -1349,202 +1809,441 @@ decl_module! {
...
+		if let Some(current_era) = Self::current_era() {
+			// Initial era has been set.
+			let current_era_start_session_index = Self::eras_start_session_index(current_era).unwrap_or_else(|| {
+				frame_support::print("Error: start_session_index must be set for current_era");
+				0
+			});
 
-				// prune out everything that's from before the first-kept index.
-				let n_to_prune = bonded.iter().take_while(|&&(era_idx, _)| era_idx < first_kept).count();
+			let era_length = session_index.checked_sub(current_era_start_session_index).unwrap_or(0); // Must never happen.
```
aliXsed, line 2133: `session_index.saturating_sub(current_era_start_session_index)`

crml/staking/src/lib.rs
```diff
@@ -1349,202 +1809,441 @@ decl_module! {
...
 
-		maybe_new_validators
+	/// Select a new validator set from the assembled stakers and their role preferences. It tries
+	/// first to peek into [`QueuedElected`]. Otherwise, it runs a new on-chain phragmen election.
+	///
+	/// If [`QueuedElected`] and [`QueuedScore`] exists, they are both removed. No further storage
+	/// is updated.
+	fn try_do_election() -> Option<ElectionResult<T::AccountId, BalanceOf<T>>> {
+		// an election result from either a stored submission or locally executed one.
+		let next_result = <QueuedElected<T>>::take().or_else(|| Self::do_on_chain_phragmen());
```
aliXsed, line 2177: `|| Self::do_on_chain_phragmen()` -> `Self::do_on_chain_phragmen`

crml/staking/src/lib.rs
```diff
@@ -1565,98 +2264,447 @@ impl<T: Trait> Module<T> {
...
+		ensure!(
+			compact_assignments.unique_targets().len() == winners.len(),
+			Error::<T>::OffchainElectionBogusWinnerCount,
+		);
+
+		// Check that the number of presented winners is sane. Most often we have more candidates
+		// than we need. Then it should be `Self::validator_count()`. Else it should be all the
+		// candidates.
+		let snapshot_validators_length = <SnapshotValidators<T>>::decode_len()
+			.map(|l| l as u32)
```
aliXsed, line 2355: map l to l?

crml/staking/src/lib.rs
```diff
@@ -1565,98 +2264,447 @@ impl<T: Trait> Module<T> {
...
+		// check the winner length only here and when we know the length of the snapshot validators
+		// length.
+		let desired_winners = Self::validator_count().min(snapshot_validators_length);
+		ensure!(
+			winners.len() as u32 == desired_winners,
+			Error::<T>::OffchainElectionBogusWinnerCount
+		);
+
+		let snapshot_nominators_len = <SnapshotNominators<T>>::decode_len()
+			.map(|l| l as u32)
```
aliXsed, line 2373: Same as in line 2351

crml/staking/src/lib.rs
```diff
@@ -1565,98 +2264,447 @@ impl<T: Trait> Module<T> {
...
+			compact_assignments.unique_targets().len() == winners.len(),
+			Error::<T>::OffchainElectionBogusWinnerCount,
+		);
+
+		// Check that the number of presented winners is sane. Most often we have more candidates
+		// than we need. Then it should be `Self::validator_count()`. Else it should be all the
+		// candidates.
+		let snapshot_validators_length = <SnapshotValidators<T>>::decode_len()
+			.map(|l| l as u32)
+			.ok_or_else(|| Error::<T>::SnapshotUnavailable)?;
```
aliXsed, line 2356: `ok_or(Error::<T>::SnapshotUnavailable)?`

crml/staking/src/lib.rs
```diff
@@ -1565,98 +2264,447 @@ impl<T: Trait> Module<T> {
...
+		// length.
+		let desired_winners = Self::validator_count().min(snapshot_validators_length);
+		ensure!(
+			winners.len() as u32 == desired_winners,
+			Error::<T>::OffchainElectionBogusWinnerCount
+		);
+
+		let snapshot_nominators_len = <SnapshotNominators<T>>::decode_len()
+			.map(|l| l as u32)
+			.ok_or_else(|| Error::<T>::SnapshotUnavailable)?;
```
aliXsed, line 2374: `ok_or(Error::<T>::SnapshotUnavailable)?`

crml/staking/src/lib.rs
```diff
@@ -1733,13 +2866,20 @@ impl<T: Trait> Convert<T::AccountId, Option<T::AccountId>> for StashOf<T> {
...
 pub struct ExposureOf<T>(sp_std::marker::PhantomData<T>);
 
 impl<T: Trait> Convert<T::AccountId, Option<Exposure<T::AccountId, BalanceOf<T>>>> for ExposureOf<T> {
 	fn convert(validator: T::AccountId) -> Option<Exposure<T::AccountId, BalanceOf<T>>> {
-		Some(<Module<T>>::stakers(&validator))
+		if let Some(active_era) = <Module<T>>::active_era() {
+			Some(<Module<T>>::eras_stakers(active_era.index, &validator))
+		} else {
+			None
+		}
```
aliXsed, line 2886: `<Module<T>>::active_era().map(|active_era| <Module<T>>::eras_stakers(active_era.index, &validator))`

crml/staking/src/lib.rs
jordy25519: it's type conversion from usize to u32,  length is an `Option<usize>`

jordy25519:
most of this code in staking came verbatim from upstream. have applied the fixes where possible. we could do a clippy pass over this whole project as another thing

crml/staking/src/lib.rs
aliXsed: I get that. In theory when we can do something as cheap as `unwrap()? as u32`,  using a map and a closure is not a good choice. In practice, possibly the compiler is optimizing both to the same binary. I guess this change also comes from the upstream. So in that case preferring what is there makes the future merge easier.

crml/staking/src/lib.rs
```diff
@@ -883,24 +1127,158 @@ decl_module! {
...
+		fn on_runtime_upgrade() -> Weight {
+			match StorageVersion::get().into() {
+				// upgraded!
+				Releases::V2 => Zero::zero(),
+				// needs upgrade
+				Releases::V1 => {
+					migration::upgrade_v1_to_v2::<T>();
+					T::MaximumBlockWeight::get()
+				},
+				// won't occur, no live networks on this version...
```
aliXsed, line 1174: If this's not going to happen, then we need to have a log or error or report of a failure here.

aliXsed:
@holygits What has been the reason for substantial change in the way amortised rewards work for an urgent update of the staking module?

crml/staking/src/rewards/mod.rs
```diff
@@ -87,19 +88,20 @@ decl_event!(
...
 		/// Development fund % take for reward payouts, parts-per-billion
 		pub DevelopmentFundTake get(fn development_fund_take) config(): Perbill;
 		/// Accumulated transaction fees for reward payout
 		pub TransactionFeePot get(fn transaction_fee_pot): BalanceOf<T>;
 		/// Historic accumulated transaction fees on reward payout
 		pub TransactionFeePotHistory get(fn transaction_fee_pot_history): VecDeque<BalanceOf<T>>;
-		/// Remaining reward amount of the eras which are not fully processed yet
-		pub QueuedEraRewards get(fn queued_era_rewards): VecDeque<BalanceOf<T>>;
 		/// Where the reward payment should be made. Keyed by stash.
+		// TODO: migrate to blake2 to prevent trie unbalancing
```
aliXsed, line 99: As far as I know, Blake2 is necessary for what users have control over and can affect the storage. AccountId here is coming from User's account which are already issued by the system to be balanced.

crml/staking/src/rewards/mod.rs
```diff
@@ -146,6 +148,18 @@ decl_module! {
 			ensure_root(origin)?;
 			ForceFiscalEra::put(true);
 		}
+
+		fn on_initialize(now: T::BlockNumber) -> Weight {
+			// payouts are scheduled to happen on quantized block intervals such that only blocks where
+			//  `now` % `BlockPayoutInterval` == 0 will every have a payout. see `schedule_reward_payouts`
+			if (now % T::BlockPayoutInterval::get()).is_zero() {
+				if let Some((validator_stash, amount)) = ScheduledPayouts::<T>::take(now) {
+					return T:: ScheduledPayoutRunner::run_payout(&validator_stash, amount, Self::scheduled_payout_era())
```
aliXsed, line 157: This is not a good load balancing when one validator has got many more nominators than another one.

jordy25519:
> @holygits What has been the reason for substantial change in the way amortised rewards work for an urgent update of the staking module?

couple reasons:
1) calculating all the payouts upfront adds to how expensive the election block is to compute
2) the vec approach to payouts wasn't very storage efficient, I don't think benchmarks reflect the actual storage cost on chain when the db is much more occupied.
this will read and decode every single payout
https://github.com/cennznet/cennznet/blob/1acfaa070ca49cbbdf3472cd33e49e36c5f88a34/crml/staking/src/rewards/mod.rs#L223
this will encode and write all the payouts we didn't use
https://github.com/cennznet/cennznet/blob/1acfaa070ca49cbbdf3472cd33e49e36c5f88a34/crml/staking/src/rewards/mod.rs#L266
this will read all payouts each time and thrash the cache
https://github.com/cennznet/cennznet/blob/1acfaa070ca49cbbdf3472cd33e49e36c5f88a34/crml/staking/src/rewards/mod.rs#L198

moving to a map structure for payouts was necessary, keeping it close to lazy payouts made sense also
so basically it's amortised, lazy payouts now

crml/staking/src/rewards/mod.rs
jordy25519: users can generate stash accounts as they like so the cost to unbalance this is fees + minimum bond, which makes it not urgent imo. `Bonded` is blake hashed upstream and we'll migrate to that in this PR also

aliXsed:
> > @holygits What has been the reason for substantial change in the way amortised rewards work for an urgent update of the staking module?
> 
> couple reasons:
> 
>     1. calculating all the payouts upfront adds to how expensive the election block is to compute
> 
>     2. the vec approach to payouts wasn't very storage efficient, I don't think benchmarks reflect the actual storage cost on chain when the db is much more occupied.
>        this will read and decode every single payout
>        https://github.com/cennznet/cennznet/blob/1acfaa070ca49cbbdf3472cd33e49e36c5f88a34/crml/staking/src/rewards/mod.rs#L223
>        
>        this will encode and write all the payouts we didn't use
>        https://github.com/cennznet/cennznet/blob/1acfaa070ca49cbbdf3472cd33e49e36c5f88a34/crml/staking/src/rewards/mod.rs#L266
>        
>        this will read all payouts each time and thrash the cache
>        https://github.com/cennznet/cennznet/blob/1acfaa070ca49cbbdf3472cd33e49e36c5f88a34/crml/staking/src/rewards/mod.rs#L198
> 
> 
> moving to a map structure for payouts was necessary, keeping it close to lazy payouts made sense also
> so basically it's amortised, lazy payouts now

process_reward_payouts has been happening on the second block of every era. So as it's not at the same time with calculating the election result, we can actually check how much of an effect we have from the current process_reward_payouts on the network. Do we have any measurements of that already? The benchmarking wasn't suggesting that even with 1000 payouts we gonna have a significant slow down. But even 1000 is configurable and can be reduced.

runtime/src/impls.rs
```diff
@@ -16,38 +16,60 @@
...
+	type Balance = Balance;
 
-impl CurrencyToVoteHandler {
-	fn factor() -> Balance {
-		(<StakingAssetCurrency<Runtime>>::total_issuance() / u64::max_value() as Balance).max(1)
-	}
-}
+	/// Feed exposure info from staking to run reward calculations
+	/// This is called by Rewards on_initialize at scheduled intervals
+	fn run_payout(validator_stash: &Self::AccountId, amount: Self::Balance, payout_era: EraIndex) -> Weight {
```
aliXsed, line 56: Instead of adding on_initialize to the rewards module, it could be added to the staking module. Then there staking module could interact with Rewards internally. With such a design, there wouldn't be any need to bring the implementation of staking/rewards functions to runtime.

crml/staking/src/rewards/mod.rs
```diff
@@ -61,10 +58,12 @@ pub trait Trait: frame_system::Trait {
 	type TreasuryModuleId: Get<ModuleId>;
 	/// The number of historical eras for which tx fee payout info should be retained.
 	type HistoricalPayoutEras: Get<u16>;
-	/// The reward payouts would be split among several blocks when their number exceeds this threshold.
-	type PayoutSplitThreshold: Get<u32>;
+	/// The gap between subsequent payouts (in number of blocks)
+	type BlockPayoutInterval: Get<Self::BlockNumber>;
```
aliXsed, line 62: It seems block spacing couples the proper behaviour of the rewards to HISTORY_DEPTH in the staking module in a sneaky way. At the time when we calculate and process payouts, ErasStakersClipped and ErasValidatorPrefs should still have the required data. However the history may have removed.

jordy25519: history depth is for lazy payouts so that reward payouts can happen some time in future eras.
in this implementation we could remove it and have `LastEraStakers` only which seems much nicer solution.
From a public API perspective there's nothing wrong here, but we happen to know the internal details of the staking module and that it does it's own garbage cleaning so the storage could disappear if payout era is < current era + history depth.
we have integration tests for this reason

runtime/src/impls.rs
jordy25519: payout scheduling is something the staking module doesn't need to know about.
It belongs more in the rewards module but needed to pass a trait to share the staking exposure storage,
this would create a dependency cycle from staking needing 'T:Rewarder' and rewards needing 'T::LastExposureInfo', this design would've compiled fine but in the end decided to avoid the cycle and
use a higher-level component with knowledge of both modules

jordy25519:
added block data to the original issue: https://docs.google.com/spreadsheets/d/1pGyLje8kC8zgIFveF8sYtzuI0MJOTinHORIcjNQHN2I/edit#gid=369491099
we could still use some io/storage metrics to add. benchmarks don't show the real delay on a live network because the trie db in test is basically empty in comparison to a live network

jordy25519:
will address benchmarking in a follow up PR

PR #411, opened 2021-04-19
Nft benchmarks

jordy25519:
Add benchmarking
Some weights could be modified to scale with the input to avoid over-charging

### DIFF
--- a/crml/nft/src/benchmarking.rs
+++ b/crml/nft/src/benchmarking.rs
@@ -1,0 +1,153 @@
+/* Copyright 2019-2021 Centrality Investments Limited
+*
+* Licensed under the LGPL, Version 3.0 (the "License");
+* you may not use this file except in compliance with the License.
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+* You may obtain a copy of the License at the root of this project source code,
+* or at:
+*     https://centrality.ai/licenses/gplv3.txt
+*     https://centrality.ai/licenses/lgplv3.txt
+*/
+
+//! NFT benchmarking.
+
+#![cfg(feature = "runtime-benchmarks")]
+
+use super::*;
+
+use frame_benchmarking::{account, benchmarks, whitelisted_caller};
+use frame_system::RawOrigin;
+use sp_runtime::Permill;
+
+use crate::types::MAX_ENTITLEMENTS;
+use crate::{Module as Nft, MAX_COLLECTION_ID_LENGTH};
+
+// Create a collection for benchmarking
+// Returns the collection id, schema, and royalties schedule
+fn setup_collection<T: Trait>(creator: T::AccountId) -> (CollectionId, NFTSchema, RoyaltiesSchedule<T::AccountId>) {
+	// Collection id with max length
+	let collection_id = [1_u8; MAX_COLLECTION_ID_LENGTH as usize].to_vec();
+
+	// Schema with max. attributes
+	let schema = (0..MAX_SCHEMA_FIELDS)
+		.map(|i| {
+			let mut v = b"probably-a-decent-upper-length-for-an-attribte-name".to_vec();
+			v.push(i as u8);
+			(v, NFTAttributeValue::String(Default::default()).type_id())
+		})
+		.collect::<NFTSchema>();
+
+	// Royalties with max. entitled addresses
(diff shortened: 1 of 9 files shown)


crml/nft/src/types.rs
```diff
@@ -101,18 +101,9 @@ impl NFTAttributeValue {
 	}
 }
 
-/// A single data field of an NFT
```
jordy25519, line 104: unused

crml/nft/src/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+fn setup_collection<T: Trait>(creator: T::AccountId) -> (CollectionId, NFTSchema, RoyaltiesSchedule<T::AccountId>) {
+	// Collection id with max length
+	let collection_id = [1_u8; MAX_COLLECTION_ID_LENGTH as usize].to_vec();
+
+	// Schema with max. attributes
+	let schema = (0..MAX_SCHEMA_FIELDS)
+		.map(|i| {
+			let mut v = b"probably-a-decent-upper-length-for-an-attribte-name".to_vec();
+			v.push(i as u8);
+			(v, NFTAttributeValue::String(Default::default()).type_id())
```
aliXsed, line 40: I would suggest a String longer than Default::default() for benchmarking here. Something that would satisfy T::MaxAttributeLength::get()

crml/nft/src/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	create_token {
+		let creator: T::AccountId = whitelisted_caller();
+		let owner: T::AccountId = account("owner", 0, 0);
+
+		let (collection_id, schema, royalties) = setup_collection::<T>(creator.clone());
+		let _ = <Nft<T>>::create_collection(RawOrigin::Signed(creator.clone()).into(), collection_id.clone(), schema.clone(), Some(royalties.clone())).expect("created collection");
+		// all attributes max. length
+		let attributes = schema.iter().map(|_| NFTAttributeValue::String([1_u8; 140_usize].to_vec())).collect::<Vec<NFTAttributeValue>>();
```
aliXsed, line 102: Could you already make schema like this in setup_collection?

aliXsed: If 140 is a maximum len that should be respected in the module, could you please add 140 as a defined const in the module and "use" it here? If it should be coming from T::MaxAttributeLength::get(), then you can use T::MaxAttributeLength::get() directly here.

crml/nft/src/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+		let collection_id = setup_token::<T>(owner.clone());
+		let new_owner: T::AccountId = account("new_owner", 0, 0);
+		let token_id = T::TokenId::from(0_u32);
+
+	}: _(RawOrigin::Signed(owner.clone()), collection_id.clone(), token_id, new_owner.clone())
+	verify {
+		assert_eq!(<Nft<T>>::token_owner(&collection_id, token_id), new_owner);
+	}
+
+	burn {
```
aliXsed, line 122: Looks like the expense of burn would be variable based on the number of tokens of a particular id that an owner has collected:
`			<CollectedTokens<T>>::mutate(&collection_id, current_owner, |tokens| {
				tokens.retain(|t| t != &token_id)
			});
`
Can you make the benchmark take that count into account or test a worst case/average scenario?

crml/nft/src/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+
+	cancel_sale {
+		let owner: T::AccountId = account("owner", 0, 0);
+		let collection_id = setup_token::<T>(owner.clone());
+		let token_id = T::TokenId::from(0_u32);
+		let payment_asset = 16_000;
+		let reserve_price = 1_000_000 * 10_000; // 1 million 4dp asset
+		let duration = T::BlockNumber::from(100_u32);
+
+		let _ = <Nft<T>>::auction(RawOrigin::Signed(owner.clone()).into(), collection_id.clone(), token_id, payment_asset, reserve_price, Some(duration)).expect("listed ok");
```
aliXsed, line 205: Again here according to the following code:
`			<CollectedTokens<T>>::mutate(&collection_id, current_owner, |tokens| {
				tokens.retain(|t| t != &token_id)
			});
`
It seesm we should have a number of auctions scheduled to make it either the worst or average case scenario.

aliXsed:
It lgtm! just left a few suggestions.

crml/nft/src/benchmarking.rs
jordy25519: needs a const here, added an assert as a workaround

PR #413, opened 2021-04-20
Add smart contract to cennznet

aliXsed:
This is the first step but a complete step on its own. Next PR would bring some contract integration tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
   push:
     branches:
     - develop
+    - trunk/*
 
 jobs:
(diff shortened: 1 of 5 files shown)


jordy25519 (approved):

.github/workflows/build.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
     - develop
+    - trunk/*
   push:
     branches:
     - develop
```
jordy25519, line 10: need to add it here also (push), otherwise it only builds once when PR opens

PR #415, opened 2021-04-21
Move benchmark files to the right destination

aliXsed:
NOTE: This may look too simple to warrant an individual script for the current structure. However with 2.0, every result should go to their own folders while being touched to include the trait definition too. For example later we need to change this script to move "crml_cennzx.rs" to "crml/cennzx/src/weights" while adding the following snippet to the produced file:
`pub trait WeightInfo {
	fn buy_asset() -> Weight;
	fn sell_asset() -> Weight;
	fn add_liquidity() -> Weight;
	fn remove_liquidity() -> Weight;
	fn set_fee_rate() -> Weight;
}`

### DIFF
--- a/scripts/use-benchmark-results.sh
+++ b/scripts/use-benchmark-results.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+echo Utilising benchmark results...
+mv *.rs "runtime/src/weights/"


scripts/use-benchmark-results.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo Utilising benchmark results...
+mv *.rs "runtime/src/weights/"
```
jordy25519, line 4: I this is all that's needed I think we can instead do `--output=runtime/src/weights` on the benchmark command

aliXsed: That was the thing, yes. Not needed for this branch right now. But when merging 2.0, this would do a more usefull job. But if you change the benchmark command to do `--output=runtime/src/weights` then when merging 2.0, we need to change that script too.

aliXsed:
@holygits @holygits I realised the pipeline script is going to be held in the same repo under cennznet. Then we can do anything that I have descriped here in that script. So I'm going to close this.

PR #417, opened 2021-04-22
Include benchmarking host functions

aliXsed:
The cofiguration which doesn't have the benchmarking host function deos not work in 2.0 any more and throws: `Cannot create a runtime: Instantiation("Instantiation: Export ext_benchmarking_commit_db_version_1 not found") ` So I'm removing it.

### DIFF
--- a/cli/src/service.rs
+++ b/cli/src/service.rs
@@ -35,18 +35,6 @@ use crate::rpc as node_rpc;
 use cennznet_primitives::types::Block;
 use cennznet_runtime::RuntimeApi;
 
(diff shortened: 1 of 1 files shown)


jordy25519:
lgtm but can you add some details e.g. error output to accompany 'does not work anymore'?

aliXsed:
> lgtm but can you add some details e.g. error output to accompany 'does not work anymore'?

@holygits I updated the PR comment.

jordy25519 (approved):

Issue #437, opened 2021-05-10
NFT semi-fungibles support


jordy25519:
Helpful to have batch methods to reduce cost and better API ergonomics
- [x] batch create tokens
- [x] batch fixed price sales (cc #434)
- [x] batch auctions (cc #434)
- [x] batch transfer tokens

miketamis:
Heres what I'm thinking, we make it so that every NFT has a unique identifier from every other NFT (even outside its collection). Collections are a NFT prototype, (for everything, metaurl, royalties, etc). This means that you can operate on an NFT with just its TokenId.

```
CollectionOwner: CollectionId => Option<AccountId>
Balances: (TokenId, AccountId) => Balance
TokenCollection: TokenId => CollectionId
CollectionTokens: (CollectionId, TokenId) => ()
TokenAttributes: (TokenId) => Vec<NFTAttributeValue>
```

jordy25519:
Yeah this makes alot of sense. Splitting the collection and token id definitely created some problems for semi-fungible support

this will use the hash of `(CollectionId, TokenId)` as the key which would be unique in the module:
`Balances: ((CollectionId, TokenId), AccountId) => Balance`

PR #448, opened 2021-05-24
 Add series royalties again 🙈

### LINKED ISSUE #434: NFT bundle sales
Allow tokens to be sold in bundles

We mostly want to support bulk semi-fungible sales with this i.e sell 10x some token.
Selling groups of tokens which are unrelated will create problems for royalties

Ideas:
- Allow token creator to set whether their token can be sold in a bundle or not
- Allow bundles sales within the same collection only
- Take flat fee or % fee of royalty whichever is larger

jordy25519:
Adds:
- Tokens/series can have individual royalties again. Set some checks to prevent bundle sales when the royalties are different
 
Fixes:
- Spellchecks + `CollectionMetdata` -> `CollectionMetadata`

closes #434 #437

### DIFF
--- a/crml/nft/src/benchmarking.rs
+++ b/crml/nft/src/benchmarking.rs
@@ -72,6 +72,7 @@ fn setup_token<T: Trait>(owner: T::AccountId) -> CollectionId {
 		Some(owner.clone()),
 		attributes.clone(),
 		None,
+		None,
 	)
 	.expect("created token");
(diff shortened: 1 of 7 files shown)


miketamis (approved):
Looks good, clean solution to allow NFTs to be bundled.

jordy25519:
run benchmark

PR #453, opened 2021-05-28
Chore/435 merge nft

### LINKED ISSUE #435: NFT module to 2.0 trunk
Merge the NFT module into the 2.0 trunk

aliXsed:
Closes #435


crml/nft/src/benchmarking.rs
```diff
@@ -0,0 +1,306 @@
...
+		let listing_id = <Nft<T>>::next_listing_id();
+		let _ = <Nft<T>>::auction(RawOrigin::Signed(owner.clone()).into(), token_id, PAYMENT_ASSET, PRICE, Some(duration)).expect("listed ok");
+
+	}: _(RawOrigin::Signed(owner.clone()), listing_id)
+	verify {
+		assert!(<Nft<T>>::listings(listing_id).is_none());
+	}
+}
+
+#[cfg(test)]
```
jordy25519, line 229: I think there's a new macro to generate all these tests auto-magically (not sure if we got it in 3.0.0-rc5)

runtime/src/lib.rs
```diff
@@ -720,7 +736,8 @@ construct_runtime!(
 		Rewards: crml_staking_rewards::{Module, Call, Storage, Config, Event<T>} = 29,
 		Bounties: pallet_bounties::{Module, Call, Storage, Event<T>} = 30,
 		Tips: pallet_tips::{Module, Call, Storage, Event<T>} = 31,
-		Contracts: pallet_contracts::{Module, Call, Config<T>, Storage, Event<T>},
+		Contracts: pallet_contracts::{Module, Call, Config<T>, Storage, Event<T>} = 32,
+		Nft: crml_nft::{Module, Call, Storage, Event<T>} = 33,
```
jordy25519: Let's keep NFT at `30`, as it already is in runtime v40

runtime/tests/staking_reward.rs
```diff
@@ -353,6 +353,13 @@ fn elected_validators_receive_transaction_fee_reward() {
 
 			// end era 1, reward payouts are scheduled
 			start_active_era(2);
+
```
jordy25519, line 356: I don't think this is testing the same thing

runtime/src/lib.rs

runtime/tests/staking_reward.rs
aliXsed: It's testing the treasury cut the right way. The magic number +6 that was in the previous version is sensitive to the aspect of the logic that was not clearly communicated by the test why they should be maintained and how the exact +6 imbalance should be produced. (In the recent test it turned out to be +5, the remainder of a division between 6 validators.)

crml/nft/src/benchmarking.rs
aliXsed: Let's keep it the same as all the other tests in the current version and upgrade the tests as a separate step.

runtime/tests/staking_reward.rs
jordy25519: Adding new check is fine but removing old one isn't, this actually documents the behaviour that payouts don't get split equally by the payout process (i.e. a remainder exists which must go to the treasury)

aliXsed: Restored the check.

jordy25519 (approved):

PR #470, opened 2021-07-27
NFT Generic RPC calls

JasonTulp:
Creates generic RPC calls for NFT,
1: getCollectionInfo: (collection_name, owner, royalties)


jordy25519:
superseded #480

PR #478, opened 2021-08-06
Add proposal votes RPC

jordy25519:
```bash
 curl \
  -H "Content-Type: application/json" \
  -d '{"id":1,"jsonrpc":"2.0","method":"governance_getProposalVotes","params":[]}' \
   localhost:9933
```

```
{
  "jsonrpc": "2.0",
  "result": [
    {
      "proposal_id": 0,
      "votes": [
        [
          "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
          true
        ],
        [
          "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
          true
        ]
      ]
    },
    {
      "proposal_id": 1,
      "votes": [
        [
          "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty",
          true
        ],
        [
          "5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y",
          null
        ]
      ]
    }
  ],
  "id": 1
}
```

### DIFF
--- a/cli/src/rpc.rs
+++ b/cli/src/rpc.rs
@@ -119,6 +119,7 @@ where
 	C::Api: crml_staking_rpc::StakingRuntimeApi<Block, AccountId>,
 	C::Api: crml_transaction_payment_rpc::TransactionPaymentRuntimeApi<Block, Balance>,
 	C::Api: crml_generic_asset_rpc::GenericAssetRuntimeApi<Block, AssetId, Balance, AccountId>,
+	C::Api: crml_governance_rpc::GovernanceRuntimeApi<Block, AccountId>,
 	P: TransactionPool + 'static,
 	SC: SelectChain<Block> + 'static,
 	B: sc_client_api::Backend<Block> + Send + Sync + 'static,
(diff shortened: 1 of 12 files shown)


cli/src/rpc.rs
```diff
@@ -195,7 +197,8 @@ where
 	io.extend_with(CennzxApi::to_delegate(Cennzx::new(client.clone())));
 	io.extend_with(NftApi::to_delegate(Nft::new(client.clone())));
 	io.extend_with(StakingApi::to_delegate(Staking::new(client.clone())));
-	io.extend_with(GenericAssetApi::to_delegate(GenericAsset::new(client)));
+	io.extend_with(GenericAssetApi::to_delegate(GenericAsset::new(client.clone())));
+	io.extend_with(GovernanceApi::to_delegate(Governance::new(client)));
```
KarishmaBothara, line 201: do we need to clone here as well?
io.extend_with(GovernanceApi::to_delegate(Governance::new(client.clone())));

jordy25519: the last one in the order gets to "move" the ownership so no clone needed

KarishmaBothara (approved):

PR #485, opened 2021-08-27
Added get_collection_listing RPC call and tests

JasonTulp:
Added a JSON RPC to the NFT module that returns all the currently listed Auctions or Fixed price sales splitting the result up with pagination and using a cursor relating to a listing_id for navigation


crml/nft/src/lib.rs
```diff
@@ -967,4 +969,46 @@ impl<T: Config> Module<T> {
...
+		collection_id: CollectionId,
+		cursor: u128,
+		limit: u16,
+	) -> (Option<u128>, Vec<(ListingId, Listing<T>)>) {
+		let mut listing_ids = OpenCollectionListings::iter_prefix(collection_id)
+			.map(|(listing_id, _)| listing_id)
+			.collect::<Vec<u128>>();
+		listing_ids.sort();
+		let last_id = listing_ids.last().copied();
+		let mut highest_cursor: u128 = 0;
```
jordy25519, line 982: nicer to assign this after the response is built:

```rust
let new_cursor = match (response.last(), listing_ids.last()) {
  (Some((highest_id, _)), Some(last_id)) if highest_id < last_id => Some(highest_id + 1)
  _ => None,
};
```

jordy25519 (commented):
lgtm in general! it's worth writing a couple test cases for the cursor edge cases

crml/nft/src/tests.rs
```diff
@@ -1781,3 +1781,159 @@ fn get_collection_info() {
...
+		let collection_id = setup_collection(owner);
+		let name = b"test-collection".to_vec();
+		let cursor: u128 = 0;
+		let limit: u16 = 1000;
+
+		assert!(has_event(RawEvent::CreateCollection(
+			collection_id,
+			name.clone(),
+			owner
+		)));
```
jordy25519: don't need to assert this in every test. can add this assert in `setup_collection` (if it's not already there)

jordy25519 (commented):
looks good in general.
would add a test where `cursor` is higher than expected.

jordy25519 (approved):

PR #516, opened 2021-10-17
Changed Permill to Perquintill for calculation

JasonTulp:
Fixed rounding error by changing type to perQuintill for calculations


JasonTulp:
Add test that will break it without this change

JasonTulp:
Not actually a bug

Issue #525, opened 2021-11-16
Bridging for ERC721


jordy25519:
Provide support for bridging ERC721 tokens as CENNZnet NFTs

Some of the pieces:
- Mapping from (collection, series) to erc721 contract
- 'collection base uri' should be removed/replaced by 'series metadata uri'
- contract factory to issue new ERC721 contracts for CENNZnet native NFTs
~- serial number should update to u256 to match 1:1 with ERC271 token addresses~
we can maintain a mapping from Ethereum token address to serial number (cheaper than making all serial numbers U256 by default)
- peg creates new collection/series per token address upon bridging

constructing a token Uri should be `seriesBaseUri/tokenId`

## conceptual mapping
![nft-erc721 drawio](https://user-images.githubusercontent.com/5133901/142932844-fda941f7-e056-455a-93f1-b8623922b62b.png)

cc #442

Issue #442, opened 2021-05-14
NFT metadata URI standards 


jordy25519:
Define a standard for constructing token metadata URIs depending on the base URI protocol
Building on https://eips.ethereum.org/EIPS/eip-1155#metadata

1) if collection base URI is `http(s)://`
format: `${collectionUri}/${tokenId}` without `0x` prefix
example: `nft.example.com/8e31838a80773cf4c25053051f651acd5c8d0184f3194b3e99804da5162fec0f`
`MetadataCID` is optional at token creation

2) if collection base URI is `ipfs://`
format: `${collectionUri}/${MetadataCID}`
example: `ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`
`MetadataCID` must be given at token creation
where `MetadataCID` is the CID of the `metadata.json` object on IPFS

3) Similar to 2) but for Arweave stored

Module would add a storage
`TokenMetadataCID: map  TokenId -> Option<Vec<u8>>`

Module could define an RPC method to construct this for clients
`api.rpc.nft.tokenMetadataUri(tokenId)` similar to ERC1155

Makes sense to make a new attribute type `CID` to support this

dhjjjer:
Hey, man! May you explain can I upload metadata to this smart contract? https://etherscan.io/address/0xe77da975fb81e7ae50a89be981e60d31e4387a29#code

JasonTulp:
> Hey, man! May you explain can I upload metadata to this smart contract? https://etherscan.io/address/0xe77da975fb81e7ae50a89be981e60d31e4387a29#code

Hi dhjjjer, on Ethereum metadata is handled by each marketplace, such as OpenSea through custom API endpoints. [Opensea Metadata Standards](https://docs.opensea.io/docs/metadata-standards) This process is a whole lot easier on CENNZnet as you can define metadata as attributes when you mint an NFT [Minting an NFT on CENNZnet](https://wiki.cennz.net/#/CENNZnet-API/NFT-API?id=minting-an-nft)

jordy25519:
## Metadata Schemes
When creating an NFT series a metadata scheme must be provided.
Initially there will be two supported schemes.

## 1) HTTP(S)
An HTTP API provides a JSON document given a valid serial number from the series.
`https://<host>/<path>+/<serialNumber>`

e.g. `https://api.example.com/tokens/123`

## 2) IPFS directory
[...]

PR #535, opened 2021-12-13
Update NFT data model

jordy25519:
### Changes
Change NFT data model to more closely resemble the ERC721 structure for future cross-chain compatibility.
- 'Collections' are a grouping of series
- 'Series' are equivalent to ERC721 home contracts
- 'Serial numbers' are equivalent to ERC721 token address

**SFT removal**
Removes the concept of SFT/copies/unique enforced by the module, all tokens are NFTs by default.
It is much simpler to leave to the token metadata to denote/manage copies.

**Metadata Reference Schemes**
`MetadataBaseUri` renamed to `MetadataScheme`
Users can choose to store token metadata using 1 of 2 reference schemes,  either:
1) https. token URIs assembled like `https://api.example.com/metadata/<serialNumber>.json`
2) IPFS directory. token URIs assembled like: `ipfs://<directoryCID/<serialNumber>.json`
see: https://github.com/cennznet/cennznet/issues/442#issuecomment-992086784

### Removes
- mint_unique function (not needed)
- SeriesAttributes (doesn't make sense anymore, could be reintroduced as TokenAttributes)
- CollectionBaseUri

## TODO:
- [x] Add migration/storage clean up functions
- [ ] refactor OpenCollectionListings -> OpenSeriesListings, since grouping by SeriesId will become more standard

partially addresses #525 & #442 (minus a token_uri RPC method)

### DIFF
--- a/crml/nft/src/lib.rs
+++ b/crml/nft/src/lib.rs
@@ -36,6 +36,7 @@ use cennznet_primitives::types::{AssetId, Balance, BlockNumber};
 use crml_support::MultiCurrency;
 use frame_support::{
 	decl_error, decl_event, decl_module, decl_storage, ensure,
+	storage::IterableStorageDoubleMap,
 	traits::{ExistenceRequirement, Get, Imbalance, WithdrawReasons},
 	transactional,
 	weights::Weight,
@@ -243,7 +244,7 @@ decl_storage! {
 		/// Block numbers where listings will close. Value is `true` if at block number `listing_id` is scheduled to close.
 		pub ListingEndSchedule get(fn listing_end_schedule): double_map hasher(twox_64_concat) T::BlockNumber, hasher(twox_64_concat) ListingId => bool;
 		/// Version of this module's storage schema
-		StorageVersion build(|_: &GenesisConfig| Releases::V0 as u32): u32;
+		StorageVersion build(|_: &GenesisConfig| Releases::V2 as u32): u32;
 	}
 }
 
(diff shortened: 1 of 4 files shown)


crml/nft/src/lib.rs
```diff
@@ -667,20 +582,14 @@ decl_module! {
 				return Err(Error::<T>::NoCollection.into());
 			}
 
-			ensure!(attributes.len() as u32 <= MAX_SCHEMA_FIELDS, Error::<T>::SchemaMaxAttributes);
-			let max_attribute_length = T::MaxAttributeLength::get() as usize;
-			for attribute in attributes.iter() {
-				ensure!(attribute.len() <= max_attribute_length, Error::<T>::MaxAttributeLength);
-			}
-
 			// Check we can issue the new tokens
```
JasonTulp, line 585: Will we need to validate the metadata_scheme passed in?

crml/nft/src/lib.rs
```diff
@@ -689,50 +598,35 @@ decl_module! {
 					Error::<T>::ActivationTimeInvalid
 				);
 				SeriesMassDrops::insert(collection_id, series_id, mass_drop);
-			} else {
```
JasonTulp, line 692: Does this mean people will be able to create an empty series and mint additional in that series?

JasonTulp (approved):
lgtm! Looks a lot tidier and I think this will remove a lot of confusion around minting

crml/nft/src/lib.rs
jordy25519: yep, couldn't think of a good reason to prevent this action vs. allow it. since with this you can configure the series then add tokens after

PR #549, opened 2021-12-21
Update/4.0.0 dev staking

jordy25519:
Update staking module to 4.0.0-dev

other;
- ran `cargo fmt --all` hence the unrelated file changes
- merged `develop`

### DIFF
--- a/crml/staking/src/lib.rs
+++ b/crml/staking/src/lib.rs
@@ -1966,7 +1966,7 @@ decl_module! {
 		///   - Initial solution is almost the same.
 		///   - Worse solution is retraced in pre-dispatch-checks which sets its own weight.
 		/// # </weight>
-		#[weight = T::WeightInfo::submit_solution_better(
+		#[weight = T::WeightInfo::submit_unsigned(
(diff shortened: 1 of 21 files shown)


crml/staking/src/lib.rs
```diff
@@ -686,7 +678,8 @@ where
 	T::ValidatorIdOf: Convert<<T as frame_system::Config>::AccountId, Option<<T as frame_system::Config>::AccountId>>,
 {
 	fn disable_validator(validator: &<T as frame_system::Config>::AccountId) -> Result<bool, ()> {
-		<pallet_session::Module<T>>::disable(validator)
+		// CHECK
+		Ok(<pallet_session::Module<T>>::disable(validator))
```
JasonTulp, line 682: Could we change the return type from Result here?

JasonTulp (approved):
Looks good!

crml/staking/src/lib.rs
jordy25519: the trait comes from substrate so we just follow what's required there

PR #594, opened 2022-03-07
Pay out priority fee to validator

JasonTulp:
Added implementation of On EVM Charge Transaction to payout priority fee to validator


runtime/src/impls.rs
```diff
@@ -223,6 +225,44 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for EthereumFindAuthor<F> {
...
+		<() as OnChargeEVMTransaction<T>>::withdraw_fee(who, fee)
+	}
+
+	fn correct_and_deposit_fee(who: &H160, corrected_fee: U256, already_withdrawn: Self::LiquidityInfo) {
+		<() as OnChargeEVMTransaction<T>>::correct_and_deposit_fee(who, corrected_fee, already_withdrawn)
+	}
+
+	fn pay_priority_fee(tip: U256) {
+		let controller = T::AddressMapping::into_account_id(<pallet_evm::Pallet<T>>::find_author());
+		let staking_ledger = <crml_staking::Module<T>>::ledger(&controller);
```
jordy25519: can get the proper account to pay the validator with this:
`let pay_to = Rewards::payee(stash)`

runtime/src/impls.rs
```diff
@@ -223,6 +225,44 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for EthereumFindAuthor<F> {
...
+	fn withdraw_fee(who: &H160, fee: U256) -> Result<Self::LiquidityInfo, pallet_evm::Error<T>> {
+		<() as OnChargeEVMTransaction<T>>::withdraw_fee(who, fee)
+	}
+
+	fn correct_and_deposit_fee(who: &H160, corrected_fee: U256, already_withdrawn: Self::LiquidityInfo) {
+		<() as OnChargeEVMTransaction<T>>::correct_and_deposit_fee(who, corrected_fee, already_withdrawn)
+	}
+
+	fn pay_priority_fee(tip: U256) {
+		let controller = T::AddressMapping::into_account_id(<pallet_evm::Pallet<T>>::find_author());
```
jordy25519: this address will not be the controller. The result of this function is going to get the controller convert to 20 byte then to a new address altogether `c -> e -> c'`

some ideas to solve this are to add a function to rewards pallet so `c` can withdraw from `c'` but the manual withdraws aren't the best

JasonTulp: Ah true, that makes it complicated. I'll have a think on some solutions

jordy25519: to give some context: `pallet_evm::Pallet<T>>::find_author()` should return a 20 byte address that can be uniquely tied to the validator. 

contracts or EVM tooling will treat this address as the 'miner'
so it will show up in evm block info and when a contract transfers to `block.coinbase`it will also pay this address.

runtime/src/impls.rs
```diff
@@ -197,32 +200,59 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for EthereumFindAuthor<F> {
...
-				H160::from_slice(&AsRef::<[u8; 32]>::as_ref(controller)[..20])
+				H160::from_slice(&AsRef::<[u8; 32]>::as_ref(stash)[..20])
 			} else {
 				H160::default()
 			}
 		})
 	}
 }
 
+/// Implements the transaction payment for a pallet implementing the `Currency`
```
jordy25519: need to update this comment

runtime/src/impls.rs
```diff
@@ -197,32 +200,58 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for EthereumFindAuthor<F> {
...
+		let pre_runtime_digests = digest.logs.iter().filter_map(|d| d.as_pre_runtime());
+		if let Some(author_index) = Babe::find_author(pre_runtime_digests) {
+			if let Some(stash) = <pallet_session::Pallet<T>>::validators().get(author_index as usize) {
+				let pay_to = Rewards::payee(stash);
+				let _ = <T as pallet_evm::Config>::Currency::deposit_into_existing(
+					&pay_to,
+					tip.low_u128().unique_saturated_into(),
+				);
+			} else {
+				log::debug!("Error processing priority fee, validator not found");
```
jordy25519: `log::error!` or `log::warn!`
same for the log below

jordy25519:
May need rebase

runtime/tests/staking_reward.rs
```diff
@@ -724,3 +788,62 @@ fn accrued_payout_nominators() {
...
+			let extrinsic = fp_self_contained::CheckedExtrinsic::<_, _, SignedExtra, _> {
+				signed: fp_self_contained::CheckedSignature::SelfContained(source),
+				function: Call::Ethereum(call),
+			};
+			use frame_support::weights::GetDispatchInfo as _;
+			let dispatch_info = extrinsic.get_dispatch_info();
+			assert_ok!(extrinsic.apply::<Runtime>(&dispatch_info, 0));
+
+			let validators = <pallet_session::Pallet<Runtime>>::validators();
+			let actual_priority_fee = 384; // Actual priority fee based on used gas
```
jordy25519: add some details on the calculation so its not a magic number

jordy25519 (approved):

jordy25519 (approved):

PR #602, opened 2022-03-17
Client syncing fixes

jordy25519:
**Fixes**:
- `frontierGenesisBlockNumber` property added to chainSpec files. This is a block number which designates the oldest block the frontier syncing task will try to operate on. Without this value the 2.1.X client will try to map every block all the way back to genesis causing stalls during update or syncing from snapshots.
- drive by fix to the deposit expiry check. It was too aggressive banning some deposits slightly earlier.

**Changes**:
- Update substrate and frontier commits to include the frontier syncing fix and prevent an unnecessary db migration
- Bump cli to 2.1.0-rc7 and runtime to v53

mainnet evm update at block 12,179,483 (7 days from now)
nikau evm update at block 2,226,331 (2 hours from now)


JasonTulp (approved):
lgtm!

PR #606, opened 2022-03-22
Custom Peg delays for Withdrawals and Deposit claims -dev

### LINKED ISSUE #600: custom peg delays

JasonTulp:
- Adds ClaimDelay which is set per asset_id and sets a minimum balance and delay
 - Changes deposits and withdrawal execution to check whether they are more than the ClaimDelay balance. They will be executed immediately if they are below the specified balance, but delayed if higher
 
Closes #600


crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+			.take(max_claims.into())
+		{
+			removed += 1;
+			match pending_claim {
+				PendingClaim::Deposit((deposit_claim, tx_hash)) => {
+					Self::process_deposit_claim(deposit_claim, tx_hash);
+				}
+				PendingClaim::Withdrawal(withdrawal_message) => {
+					// At this stage it is assumed that a mapping between erc20 to asset id exists for this token
+					let asset_id = Self::erc20_to_asset(withdrawal_message.token_address).unwrap_or_default();
```
jordy25519: better to fail here than use default asset Id as that has unintended consequences i.e. turns into an Eth withdraw. doesn't seem possible for it to fail but better to be defensive
```suggestion
										let asset_id = Self::erc20_to_asset(withdrawal_message.token_address).unwrap_or_default();
```

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+		);
+		let beneficiary: T::AccountId = T::AccountId::decode(&mut &claim.beneficiary.0[..]).unwrap();
+		match event_claim_id {
+			Ok(claim_id) => Self::deposit_event(<Event<T>>::Erc20Claim(claim_id, beneficiary)),
+			_ => {
+				// There was an error submitting an event claim. Could be that the bridge is down
+				// In this case, delay the deposit claim by 120 blocks
+				let claim_id = NextClaimId::get();
+				let claim_block =
+					<frame_system::Pallet<T>>::block_number().saturating_add(T::BlockNumber::from(FAILED_CLAIM_DELAY));
```
jordy25519: should pass `FAILED_CLAIM_DELAY` this value in the Config trait so that it can be set to match `T::SessionsPerEra` clearly in runtime/src/lib

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+				ClaimSchedule::<T>::insert(claim_block, claim_id, PendingClaim::Deposit((claim.clone(), tx_hash)));
+				NextClaimId::mutate(|c| *c += 1);
```
jordy25519: need to check if claimId will overflow before scheduling (also insert)
```suggestion
				NextClaimId::insert(claim_id + 1);
```

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+		let event_claim_id = T::EthBridge::submit_event_claim(
+			&Self::contract_address().into(),
+			&T::DepositEventSignature::get().into(),
+			&tx_hash,
+			&EthAbiCodec::encode(&claim),
+		);
+		let beneficiary: T::AccountId = T::AccountId::decode(&mut &claim.beneficiary.0[..]).unwrap();
+		match event_claim_id {
+			Ok(claim_id) => Self::deposit_event(<Event<T>>::Erc20Claim(claim_id, beneficiary)),
+			_ => {
```
jordy25519: nit: makes it clear its an error condition. could also log the err
```suggestion
			Err(_) => {
```

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+				let withdrawal_hash: T::Hash = T::Hashing::hash(&mut (message.clone(), proof_id).encode());
+				WithdrawalDigests::<T>::insert(proof_id, withdrawal_hash);
+				Self::deposit_event(<Event<T>>::Erc20Withdraw(
+					proof_id,
+					asset_id,
+					amount,
+					message.beneficiary,
+				));
+			}
+			_ => {
```
jordy25519: Same here, clear its an error and helps to log the err
```suggestion
			Err(_) => {
```

crml/erc20-peg/src/lib.rs
```diff
@@ -187,21 +233,37 @@ decl_module! {
 					ExistenceRequirement::KeepAlive,
 				)?;
 			} else {
-				let _imbalance = T::MultiCurrency::withdraw(&origin, asset_id, amount, WithdrawReasons::TRANSFER, frame_support::traits::ExistenceRequirement::KeepAlive)?;
+				let _imbalance = T::MultiCurrency::withdraw(
+					&origin,
+					asset_id,
+					amount,
+					WithdrawReasons::empty(),
```
jordy25519:
```suggestion
					WithdrawReasons::TRANSFER,
```

crml/erc20-peg/src/lib.rs
```diff
@@ -187,21 +233,37 @@ decl_module! {
...
+						// Store withdrawal to be claimed later
+						let claim_id = NextClaimId::get();
+						let claim_block = <frame_system::Pallet<T>>::block_number().saturating_add(delay);
+						ClaimSchedule::<T>::insert(claim_block, claim_id, PendingClaim::Withdrawal(message));
+						NextClaimId::mutate(|c| *c += 1);
+						Self::deposit_event(<Event<T>>::Erc20WithdrawalDelayed(claim_id, claim_block, amount, beneficiary));
+					}
+				},
+				None => Self::process_withdrawal(message, asset_id),
+			}
```
jordy25519: nit:  would refactor this so it looks for the delay case and everything else by default is: `Self::process_withdrawal(message, asset_id),`

```
if let Some((min_amont, delay)) = claim_delay {
  if min_amount < amount {
    // do delay stuff..
    return
  }
}
// do normal case...
```

crml/erc20-peg/src/lib.rs
```diff
@@ -154,17 +186,31 @@ decl_module! {
...
 
-			let event_claim_id = T::EthBridge::submit_event_claim(
-					&Self::contract_address().into(),
-					&T::DepositEventSignature::get().into(),
-					&tx_hash,
-					&EthAbiCodec::encode(&claim),
-			)?;
-
-			Self::deposit_event(<Event<T>>::Erc20Claim(event_claim_id, origin));
+			let asset_id = Self::erc20_to_asset(claim.token_address);
```
jordy25519, line 243: nit: same as comment below, clearer to have delay case and not delay case
```
if let Some((min_amount, delay)) = claim_delay {
  if min_amount < amount {
    // do delay stuff..
    return
  }
}
// do normal case...
```

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+					amount,
+					message.beneficiary,
+				));
+			}
+			_ => {
+				// There was an error generating an event proof. Could be that the bridge is down
+				// In this case, delay the withdrawal by 120 blocks
+				let claim_id = NextClaimId::get();
+				let claim_block =
+					<frame_system::Pallet<T>>::block_number().saturating_add(T::BlockNumber::from(FAILED_CLAIM_DELAY));
```
jordy25519: same re `T::FailedClaimDelay` should be on Config
and check overflow on claimId + insert

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +295,104 @@ decl_module! {
...
+			Self::deposit_event(<Event<T>>::ClaimDelaySet(asset_id, min_balance, delay));
+		}
 	}
 }
 
 impl<T: Config> Module<T> {
+	/// Process claims at a block after a delay
+	fn process_claims_at(now: T::BlockNumber, max_claims: u8) -> u8 {
+		let mut removed = 0_u8;
+		for (block, claim_id, pending_claim) in ClaimSchedule::<T>::iter()
```
jordy25519: not 100% about the internals of this `iter()` i.e. how many entries it decodes and in what order for the filtering. The iter order is based on the encoding of the key (block.encode) which isn't necessarily the same as the block ordering.

crml/erc20-peg/src/lib.rs
```diff
@@ -187,21 +236,37 @@ decl_module! {
...
-			let withdrawal_hash: T::Hash = T::Hashing::hash(&mut (message, event_proof_id).encode());
-			WithdrawalDigests::<T>::insert(event_proof_id, withdrawal_hash);
 
-			Self::deposit_event(<Event<T>>::Erc20Withdraw(event_proof_id, asset_id, amount, beneficiary));
+			let claim_delay: Option<(Balance, T::BlockNumber)> = Self::claim_delay(asset_id);
+			if let Some((min_amount, delay)) = claim_delay {
+				if min_amount <= amount {
+					// Store withdrawal to be claimed later
+					let claim_id = NextClaimId::get();
+					ensure!(claim_id.checked_add(One::one()).is_some(), Error::<T>::NoAvailableIds);
```
jordy25519: double up

crml/erc20-peg/src/lib.rs
```diff
@@ -154,17 +193,27 @@ decl_module! {
...
-			let event_claim_id = T::EthBridge::submit_event_claim(
-					&Self::contract_address().into(),
-					&T::DepositEventSignature::get().into(),
-					&tx_hash,
-					&EthAbiCodec::encode(&claim),
-			)?;
-
-			Self::deposit_event(<Event<T>>::Erc20Claim(event_claim_id, origin));
+			let asset_id = Self::erc20_to_asset(claim.token_address);
+			if asset_id.is_some() {
```
jordy25519, line 244: I don't think this check is required, for deposits it will allocate a new generic asset id on success

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +298,119 @@ decl_module! {
...
+		);
+		let beneficiary: T::AccountId = T::AccountId::decode(&mut &claim.beneficiary.0[..]).unwrap();
+		match event_claim_id {
+			Ok(claim_id) => Self::deposit_event(<Event<T>>::Erc20Claim(claim_id, beneficiary)),
+			Err(_) => {
+				// There was an error submitting an event claim. Could be that the bridge is down
+				// In this case, delay the deposit claim by 120 blocks
+				let claim_id = NextClaimId::get();
+				if !claim_id.checked_add(One::one()).is_some() {
+					log::error!("📌 No available claim ids");
```
jordy25519: now that I think about it these could be made into events which is better for visibility

crml/erc20-peg/src/lib.rs
JasonTulp: This is required to get the claim_delay as it uses unwrap, if the erc_to_asset mapping isn't set, then it just assumes there is no claim_delay (Given we only set delays for a small number of asset_ids)

JasonTulp: Potentially we'd need to put a check in set_claim_delay to make sure a mapping already exists before adding a delay for it

crml/erc20-peg/src/lib.rs
```diff
@@ -66,6 +72,16 @@ decl_storage! {
...
 		/// Metadata for well-known erc20 tokens (symbol, decimals)
 		Erc20Meta get(fn erc20_meta): map hasher(twox_64_concat) EthAddress => Option<(Vec<u8>, u8)>;
+		/// Map from asset_id to minimum amount and delay
+		ClaimDelay get(fn claim_delay): map hasher(twox_64_concat) AssetId => Option<(Balance, T::BlockNumber)>;
+		/// Map from claim id to claim
+		PendingClaims get(fn pending_claims): map hasher(twox_64_concat) ClaimId => Option<PendingClaim>;
+		/// Map from block number to claims scheduled for that block
+		ClaimSchedule get(fn claim_schedule): map hasher(twox_64_concat) T::BlockNumber => Option<Vec<ClaimId>>;
+		/// The blocks with claims that are ready to be processed
+		ReadyBlocks get(fn ready_blocks): Option<Vec<T::BlockNumber>>;
```
jordy25519: wouldn't use the Option here, empty vec is fine as a default

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,57 @@ decl_module! {
...
 		fn deposit_event() = default;
 
+		/// Check and process outstanding claims
+		fn on_initialize(now: T::BlockNumber) -> Weight {
+			let mut weight: u64 = 10_000_000;
+			if ClaimSchedule::<T>::contains_key(now) {
+				match Self::ready_blocks() {
+					Some(mut blocks) => {
+						// TODO Check that ready_blocks doesn't overflow
+						blocks.push(now);
```
jordy25519: try to use `ReadyBlocks<T>::append(now)` as it does no decoding, it just adds the new block to storage

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,57 @@ decl_module! {
...
+			let initial_read_cost = 10_000_000;
+			// Ensure we have enough weight to perform the initial read
+			if remaining_weight <= initial_read_cost {
+				return 0;
+			}
+			// Check if there are any blocks pending, else return early
+			let ready_blocks = Self::ready_blocks();
+			if !ready_blocks.is_some() {
+				return initial_read_cost;
+			}
```
jordy25519: Use decode_len for an efficient existence check, it won't decode the whole vec
`ReadyBlocks::decode_len().is_none()`

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,57 @@ decl_module! {
...
+			let mut ready_blocks: Vec<T::BlockNumber> = ready_blocks.unwrap();
+			// Process as many claims as we can
+			let weight_each: Weight = 50_000_000;
+			let max_claims = (remaining_weight / weight_each).saturated_into::<u8>();
+
+			let mut processed_claims: u8 = 0;
+
+			if ready_blocks[0] <= now {
```
jordy25519: don't think this check is needed, since it should be ready if its in ready_blocks

crml/erc20-peg/src/lib.rs
```diff
@@ -66,6 +72,16 @@ decl_storage! {
 		AssetIdToErc20 get(fn asset_to_erc20): map hasher(twox_64_concat) AssetId => Option<EthAddress>;
 		/// Metadata for well-known erc20 tokens (symbol, decimals)
 		Erc20Meta get(fn erc20_meta): map hasher(twox_64_concat) EthAddress => Option<(Vec<u8>, u8)>;
+		/// Map from asset_id to minimum amount and delay
+		ClaimDelay get(fn claim_delay): map hasher(twox_64_concat) AssetId => Option<(Balance, T::BlockNumber)>;
+		/// Map from claim id to claim
+		PendingClaims get(fn pending_claims): map hasher(twox_64_concat) ClaimId => Option<PendingClaim>;
+		/// Map from block number to claims scheduled for that block
+		ClaimSchedule get(fn claim_schedule): map hasher(twox_64_concat) T::BlockNumber => Option<Vec<ClaimId>>;
```
jordy25519: wouldn't use the Option here, empty vec is fine as a default

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,57 @@ decl_module! {
...
+
+			let mut processed_claims: u8 = 0;
+
+			if ready_blocks[0] <= now {
+				while processed_claims < max_claims && !ready_blocks.is_empty() {
+					if ClaimSchedule::<T>::contains_key(ready_blocks[0]) {
+						processed_claims += Self::process_claims_at(ready_blocks[0], max_claims - processed_claims);
+					} else {
+						ready_blocks.remove(0);
+					}
```
jordy25519: try to avoid checking storage every loop (contains_key)
`.remove(0)` can be expensive since it shifts all elements left

something like `put(ready_blocks[completed_count:]` may be nicer

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +324,150 @@ decl_module! {
...
+			ClaimDelay::<T>::insert(asset_id, (min_balance, delay));
+			Self::deposit_event(<Event<T>>::ClaimDelaySet(asset_id, min_balance, delay));
+		}
 	}
 }
 
 impl<T: Config> Module<T> {
+	/// Process claims at a block after a delay
+	fn process_claims_at(now: T::BlockNumber, max_claims: u8) -> u8 {
+		let claim_ids = Self::claim_schedule(now);
```
jordy25519: this function would be cleaner if it was given a list of claims to process vs. block number

crml/erc20-peg/src/lib.rs
```diff
@@ -156,51 +156,69 @@ decl_module! {
...
-				return initial_read_cost;
+			// Check that there are blocks in ready_blocks
+			if ReadyBlocks::<T>::decode_len().is_none() {
+				return 0;
 			}
-			let mut ready_blocks: Vec<T::BlockNumber> = ready_blocks.unwrap();
+
 			// Process as many claims as we can
 			let weight_each: Weight = 50_000_000;
 			let max_claims = (remaining_weight / weight_each).saturated_into::<u8>();
```
jordy25519:
```suggestion
			let max_claims = (remaining_weight - initial_read_cost / weight_each).saturated_into::<u8>();
```

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,75 @@ decl_module! {
 
 		fn deposit_event() = default;
 
+		/// Check and process outstanding claims
+		fn on_initialize(now: T::BlockNumber) -> Weight {
+			let mut weight: u64 = 10_000_000;
```
jordy25519: There's a constant to calculate the cost of various storage reads in frame_support weights, also weight.rs files have examples

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,75 @@ decl_module! {
...
+				}
+			}
+			// Update ready blocks
+			if completed_blocks_count != 0 {
+				if completed_blocks_count >= ready_blocks.len() as u8 {
+					let new_ready_blocks: Vec<T::BlockNumber> = vec![];
+					ReadyBlocks::<T>::put(new_ready_blocks);
+				} else {
+					ReadyBlocks::<T>::put(ready_blocks.split_off(completed_blocks_count.into()));
+				}
```
jordy25519: can this be a one-liner: `ReadyBlocks::<T>::put(&ready_blocks[completed_blocks_count..])` ?

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,60 @@ decl_module! {
...
+
+		/// Check and process outstanding claims
+		fn on_idle(_now: T::BlockNumber, remaining_weight: Weight) -> Weight {
+			let initial_read_cost = 10_000_000;
+			// Ensure we have enough weight to perform the initial read
+			if remaining_weight <= initial_read_cost {
+				return 0;
+			}
+			// Check that there are blocks in ready_blocks
+			if ReadyBlocks::<T>::decode_len().is_none() {
```
jordy25519: could check Some(0) here as well

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,60 @@ decl_module! {
...
+				return 0;
+			}
+
+			// Process as many claims as we can
+			let weight_each: Weight = 50_000_000;
```
jordy25519: something like:
```suggestion
			let weight_each: Weight = T::WeightInfo::withdraw()
```

crml/erc20-peg/src/lib.rs
```diff
@@ -125,6 +152,60 @@ decl_module! {
...
+			let max_claims = ((remaining_weight - initial_read_cost) / weight_each).saturated_into::<u8>();
+			let mut ready_blocks: Vec<T::BlockNumber> = Self::ready_blocks();
+			// Total claims processed in this block
+			let mut processed_claim_count: u8 = 0;
+			// Count of blocks where all claims have been processed
+			let mut processed_block_count: u8 = 0;
+
+			for block in ready_blocks.iter() {
```
jordy25519, line 191: would add some unit tests for the processed claim and processed block handing

crml/erc20-peg/src/lib.rs
```diff
@@ -66,6 +72,16 @@ decl_storage! {
 		AssetIdToErc20 get(fn asset_to_erc20): map hasher(twox_64_concat) AssetId => Option<EthAddress>;
 		/// Metadata for well-known erc20 tokens (symbol, decimals)
 		Erc20Meta get(fn erc20_meta): map hasher(twox_64_concat) EthAddress => Option<(Vec<u8>, u8)>;
+		/// Map from asset_id to minimum amount and delay
+		ClaimDelay get(fn claim_delay): map hasher(twox_64_concat) AssetId => Option<(Balance, T::BlockNumber)>;
+		/// Map from claim id to claim
+		PendingClaims get(fn pending_claims): map hasher(twox_64_concat) ClaimId => Option<PendingClaim>;
+		/// Map from block number to claims scheduled for that block
+		ClaimSchedule get(fn claim_schedule): map hasher(twox_64_concat) T::BlockNumber => Vec<ClaimId>;
```
jordy25519: We should use the language 'delay' for this so its clear on intention
```suggestion
		DelayedClaims get(fn pending_claims): map hasher(twox_64_concat) ClaimId => Option<DelayedClaim>;
		/// Map from block number to claims scheduled for that block
		DelayedClaimSchedule get(fn delayed_claim_schedule): map hasher(twox_64_concat) T::BlockNumber => Vec<ClaimId>;
```

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +327,125 @@ decl_module! {
...
+			match pending_claim {
+				PendingClaim::Deposit((deposit_claim, tx_hash)) => {
+					Self::process_deposit_claim(deposit_claim, tx_hash);
+				}
+				PendingClaim::Withdrawal(withdrawal_message) => {
+					// At this stage it is assumed that a mapping between erc20 to asset id exists for this token
+					let asset_id = Self::erc20_to_asset(withdrawal_message.token_address);
+					if asset_id.is_some() {
+						Self::process_withdrawal(withdrawal_message, asset_id.unwrap());
+					} else {
```
jordy25519:
```suggestion
					if let Some(asset_id) = asset_id {
						Self::process_withdrawal(withdrawal_message, asset_id);
					} else {
```

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +327,125 @@ decl_module! {
...
+			&EthAbiCodec::encode(&claim),
+		);
+		let beneficiary: T::AccountId = T::AccountId::decode(&mut &claim.beneficiary.0[..]).unwrap();
+		match event_claim_id {
+			Ok(claim_id) => Self::deposit_event(<Event<T>>::Erc20Claim(claim_id, beneficiary)),
+			Err(_) => {
+				// There was an error submitting an event claim. Could be that the bridge is down
+				// In this case, delay the deposit claim by 120 blocks
+				Self::delay_claim(
+					T::BlockNumber::from(T::FailedClaimDelay::get()),
```
holygits, line 379: may need some max retry attempts to prevent the indefinite retrying thing from happening

crml/erc20-peg/src/lib.rs
```diff
@@ -46,6 +52,8 @@ pub trait Config: frame_system::Config {
 	type PegPalletId: Get<PalletId>;
 	/// The EVM event signature of a deposit
 	type DepositEventSignature: Get<[u8; 32]>;
+	/// Failed claim delay length
+	type FailedClaimDelay: Get<u32>;
```
JasonTulp: Need to remove this

jordy25519 (approved):

crml/erc20-peg/src/lib.rs
```diff
@@ -233,10 +332,111 @@ decl_module! {
...
+	fn process_deposit_claim(claim: Erc20DepositEvent, tx_hash: H256) {
+		let event_claim_id = T::EthBridge::submit_event_claim(
+			&Self::contract_address().into(),
+			&T::DepositEventSignature::get().into(),
+			&tx_hash,
+			&EthAbiCodec::encode(&claim),
+		);
+		let beneficiary: T::AccountId = T::AccountId::decode(&mut &claim.beneficiary.0[..]).unwrap();
+		match event_claim_id {
+			Ok(claim_id) => Self::deposit_event(<Event<T>>::Erc20Claim(claim_id, beneficiary)),
```
KarishmaBothara, line 379: The event deposited earlier had a different claimer id, not beneficiary as second param..
https://github.com/cennznet/cennznet/blob/35fff7139af05f541fefd1167fa3525e1ee5e901/crml/erc20-peg/src/lib.rs#L129

PR #610, opened 2022-03-24
NFT migration support

jordy25519:
- Add storage migration for existing NFT listings (adds marketplace_id field)
- Reinstate seriesMetdataURI to support read-only queries
- Add migration function to allow collection owners to move to the new seriesMetadataScheme/tokenUri


JasonTulp (approved):
lgtm!

Issue #607, opened 2022-03-23
Release pipeline improvements


jordy25519:
Bunch of improvements to our CI:
- [x] release pipeline should upload the new compressed wasm blob
- [ ] prune docker images (build server fills up 200GB+)
- [ ] add the spec_version to the wasm filename for clarity (e.g. `cennznet-<version#>.compact.wasm`
- [ ] add try-runtime tests for release PRs
- [ ] add evm e2e tests for release PRs
- [x] code coverage (comment `check coverage` on a PR)

PR #616, opened 2022-03-29
Chore/compress wasm blob

jordy25519:
Bring the 2.1.1 version bump in
Ensure the wasm blob is the compressed version
cc #607


JasonTulp (approved):
lgtm!

Issue #578, opened 2022-02-10
GA refactors


jordy25519:
1) refactor staking asset Id and spending asset Id as constant types (reduce storage reads)
These are not likely to ever change dynamically

```rust
pub struct StakingAssetIdAuthority<T>(sp_std::marker::PhantomData<T>);

impl<T: Config> AssetIdAuthority for StakingAssetIdAuthority<T> {
	type AssetId = T::AssetId;
	fn asset_id() -> Self::AssetId {
		<Module<T>>::staking_asset_id()
	}
}
```

2) Locks will never be used besides staking so we should track a single locked amount instead of `iter().sum()` all the time

3) The `reserve()` api should handle checking lock balance properly so we don't need code like this https://github.com/cennznet/cennznet/pull/628/files/1b1c17d904ce676b9dacdafe1125fe0d10fe2deb#diff-9a45c7441ed9787c51c9858f35c52438a0c97b286a3ef14ff770a586b9a711d2R1028-R1039 to check the locks

4) Reserve asset ID separation has not proven useful should be removed

5) Review Permissions enum

jordy25519:
evm will be doing lots of spending assetId queries

jordy25519:
related #641

PR #628, opened 2022-04-14
NFT Offer System

JasonTulp:
Adds the ability to create offers on an NFT. 
The token owner can choose to accept or reject the offer.
The person who creates the offer can cancel the offer.

### DIFF
--- a/crml/nft/src/lib.rs
+++ b/crml/nft/src/lib.rs
@@ -97,6 +97,7 @@ decl_event!(
 		CollectionNameType = CollectionNameType,
 		Permill = Permill,
 		MarketplaceId = MarketplaceId,
+		OfferId = OfferId,
 	{
 		/// A new token collection was created (collection, name, owner)
 		CreateCollection(CollectionId, CollectionNameType, AccountId),
@@ -126,6 +127,14 @@ decl_event!(
 		Bid(CollectionId, ListingId, Balance),
 		/// An account has been registered as a marketplace (account, entitlement, marketplace_id)
 		RegisteredMarketplace(AccountId, Permill, MarketplaceId),
+		/// An offer has been made on an NFT (offer_id, amount, asset_id, marketplace_id, buyer)
+		OfferMade(OfferId, Balance, AssetId, Option<MarketplaceId>, AccountId),
+		/// An offer has been accepted (offer_id)
+		OfferAccepted(OfferId),
+		/// An offer has been rejected (offer_id)
+		OfferRejected(OfferId),
+		/// An offer has been cancelled (offer_id)
+		OfferCancelled(OfferId),
 	}
 );
 
@@ -171,7 +180,13 @@ decl_error! {
 		/// The Series name has been set
 		NameAlreadySet,
 		/// The metadata path is invalid (non-utf8 or empty)
-		InvalidMetadataPath
+		InvalidMetadataPath,
+		/// No offer exists for the given OfferId
+		InvalidOffer,
+		/// The signer is not the buyer
+		NotBuyer,
+		/// The signer owns the token and can't make an offer
+		IsTokenOwner,
 	}
 }
 
@@ -223,6 +238,12 @@ decl_storage! {
 		pub ListingWinningBid get(fn listing_winning_bid): map hasher(twox_64_concat) ListingId => Option<(T::AccountId, Balance)>;
 		/// Block numbers where listings will close. Value is `true` if at block number `listing_id` is scheduled to close.
 		pub ListingEndSchedule get(fn listing_end_schedule): double_map hasher(twox_64_concat) T::BlockNumber, hasher(twox_64_concat) ListingId => bool;
+		/// Map from offer_id to the information related to the offer
+		pub Offers get(fn offers): map hasher(twox_64_concat) OfferId => Option<Offer<T::AccountId>>;
+		/// Maps from token_id to a vector of offer_ids on that token
+		pub TokenOffers get(fn token_offers): map hasher(twox_64_concat) TokenId => Vec<OfferId>;
+		/// The next available offer_id
+		pub NextOfferId get(fn next_offer_id): OfferId;
 		/// Version of this module's storage schema
 		StorageVersion build(|_: &GenesisConfig| Releases::V2 as u32): u32;
 	}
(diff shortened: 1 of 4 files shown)


crml/nft/src/lib.rs
```diff
@@ -974,6 +999,152 @@ decl_module! {
...
+				Ok(())
+			} else {
+				Err(Error::<T>::InvalidOffer.into())
+			}
+		}
+
+		/// Rejects an offer on a token
+		/// Caller must be token owner
+		#[weight = 0]
+		fn reject_offer (
```
jordy25519: not sure this is needed so would leave out
can set the offer fee/weight higher to deter bad offers

crml/nft/src/lib.rs
```diff
@@ -974,6 +999,152 @@ decl_module! {
...
+				let _ = T::MultiCurrency::ensure_can_withdraw(&origin, asset_id, amount, WithdrawReasons::RESERVE, balance_after_bid)?;
+			}
+
+			// try lock funds
+			T::MultiCurrency::reserve(&origin, asset_id, amount)?;
+
+			let mut token_offers = Self::token_offers(token_id);
+			token_offers.push(offer_id);
+			TokenOffers::insert(token_id, token_offers);
+
```
jordy25519: `TokenOffers::append(offer_id)`

crml/nft/src/lib.rs
```diff
@@ -974,6 +999,152 @@ decl_module! {
...
 		}
+
+		/// Create an offer on a token
+		/// Locks funds until offer is accepted, rejected or cancelled
+		#[weight = 0]
+		fn make_offer (
+			origin,
+			token_id: TokenId,
+			amount: Balance,
+			asset_id: AssetId,
```
jordy25519, line 1011: would recommend this become an enum so this api doesn't need to change when we add new offer types, also storage won't need migrating either e.g.
```rust
enum OfferType {
   Simple(AssetId, Balance),
   // future
   Deadline(AssetId, Balance, Deadline),
   Trade(Vec<TokenId>),
}
```

crml/nft/src/lib.rs
```diff
@@ -974,6 +999,152 @@ decl_module! {
...
+						>= seller_balance.saturating_add(for_seller),
+					Error::<T>::InternalPayment
+				);
+				Self::do_transfer_unchecked(token_id.0, token_id.1, &vec![token_id.2], &origin, &offer.buyer)?;
+
+				// Clean storage
+				Offers::<T>::remove(offer_id);
+				let mut token_offers = Self::token_offers(token_id);
+				token_offers.retain(|&x| x != offer_id);
+				TokenOffers::insert(token_id, token_offers);
```
jordy25519: `mutate` is good for this kind of read and update operation

crml/nft/src/lib.rs
```diff
@@ -974,6 +999,152 @@ decl_module! {
...
+					let entitlements = royalties_schedule.entitlements.clone();
+					for (who, entitlement) in entitlements.into_iter() {
+						let royalty = entitlement * offer.amount;
+						let _ = T::MultiCurrency::repatriate_reserved(&offer.buyer, offer.asset_id, &who, royalty)?;
+						for_seller -= royalty;
+					}
+				}
+
+				let seller_balance = T::MultiCurrency::free_balance(&origin, offer.asset_id);
+				let _ = T::MultiCurrency::repatriate_reserved(&offer.buyer, offer.asset_id, &origin, for_seller)?;
```
jordy25519: can we find a way to reuse this code block across payments?

JasonTulp: Yeah was thinking that, will separate out into a separate function

crml/nft/src/lib.rs
```diff
@@ -974,6 +1001,130 @@ decl_module! {
 				None => Err(Error::<T>::NotForFixedPriceSale.into()),
 			}
 		}
+
+		/// Create an offer on a token
+		/// Locks funds until offer is accepted, rejected or cancelled
+		#[weight = 0]
```
jordy25519: set weights + use `#[transactional]` on all these methods that can fail partway through

crml/nft/src/lib.rs
```diff
@@ -974,6 +1001,130 @@ decl_module! {
...
+
+			// ensure the token_id is not currently in an auction
+			if let Some(TokenLockReason::Listed(listing_id)) = Self::token_locks(token_id) {
+				match Self::listings(listing_id) {
+					Some(Listing::<T>::Auction(_)) => return Err(Error::<T>::TokenOnAuction.into()),
+					None | Some(Listing::<T>::FixedPrice(_)) => (),
+				}
+			}
+			// check user has the required funds to make this offer
+			let balance = T::MultiCurrency::free_balance(&origin, asset_id);
```
jordy25519, line 1028: some context:
this is needed to check balance locks since part of free_balance could actually be locked (only applies to CENNZ)

`reserve()` will fail correctly if user has too low balance for every other currency

added mention of this: https://github.com/cennznet/cennznet/issues/578

crml/nft/src/lib.rs
```diff
@@ -974,6 +1001,130 @@ decl_module! {
...
+			offer_id: OfferId,
+		) -> DispatchResult {
+			let origin = ensure_signed(origin)?;
+			if let Some(offer_type) = Self::offers(offer_id) {
+				match offer_type {
+					OfferType::Simple(offer) => {
+						let token_id = offer.token_id;
+						ensure!(Self::token_owner((token_id.0, token_id.1), token_id.2) == origin, Error::<T>::NoPermission);
+
+						let royalties_schedule = Self::check_bundle_royalties(&vec![token_id], offer.marketplace_id)?;
```
jordy25519, line 1095: why `check_bundle_royalties` here?

crml/nft/src/lib.rs
```diff
@@ -974,6 +1001,130 @@ decl_module! {
...
+							vec![offer.token_id],
+							offer.amount,
+							royalties_schedule,
+						) {
+							// offer settlement failed despite our prior validations.
+							// release winning bid funds
+							log!(error, "🃏 bid settlement failed: {:?}", err);
+							T::MultiCurrency::unreserve(&offer.buyer, offer.asset_id, offer.amount);
+
+						} else {
```
jordy25519: simply take payment or fail should do
```suggestion
						Self::process_token_payment(
							&offer.buyer,
							&origin,
							offer.asset_id,
							vec![offer.token_id],
							offer.amount,
							royalties_schedule,
						)?;
```

crml/nft/src/lib.rs
```diff
@@ -974,6 +1001,130 @@ decl_module! {
...
+							// offer settlement success
+							// Clean storage
+							Offers::<T>::remove(offer_id);
+							TokenOffers::mutate(token_id, |offers| offers.retain(|&x| x != offer_id));
```
jordy25519: `offers.position(offer_id).map(|idx| offers.remove(idx))` or `offers.binary_search(offer_id).map(|idx| offers.remove(idx))` so it doesn't iterate all the offer_ids

crml/nft/src/lib.rs
```diff
@@ -1220,38 +1378,44 @@ impl<T: Config> Module<T> {
 
 		removed
 	}
-	/// Settle an auction listing (guaranteed to be atomic).
+	/// Settle an auction listing or accepted offer
+	/// (guaranteed to be atomic).
 	/// - transfer funds from winning bidder to entitled royalty accounts and seller
 	/// - transfer ownership to the winning bidder
```
jordy25519, line 1372: keep the #[transactional]`

crml/nft/src/types.rs
```diff
@@ -31,6 +31,26 @@ use variant_count::VariantCount;
...
+	pub buyer: AccountId,
+	pub marketplace_id: Option<MarketplaceId>,
+}
+
+#[derive(Decode, Encode, Debug, Clone, PartialEq, TypeInfo)]
+pub enum OfferType<AccountId> {
+	Simple(SimpleOffer<AccountId>),
+	// future
+	Deadline,
+	Trade,
```
jordy25519: better to leave out if they're unused

crml/nft/src/lib.rs
JasonTulp: check_bundle_royalties returns a royalty schedule based on the token_id and marketplace_id, which can then be passed into the process_token_payment function

crml/nft/src/lib.rs
```diff
@@ -974,6 +999,120 @@ decl_module! {
...
+			offer_id: OfferId,
+		) -> DispatchResult {
+			let origin = ensure_signed(origin)?;
+			if let Some(offer_type) = Self::offers(offer_id) {
+				match offer_type {
+					OfferType::Simple(offer) => {
+						ensure!(offer.buyer == origin, Error::<T>::NotBuyer);
+						T::MultiCurrency::unreserve(&origin, offer.asset_id, offer.amount);
+						Offers::<T>::remove(offer_id);
+						TokenOffers::remove(offer.token_id);
```
jordy25519: does this remove the element _at_ index `offer.token_id` or does it remove the element _with_ value `offer.token_id`?

JasonTulp: Good catch! It should be the same as accepting an offer. I'll change it and add some tests in for it

jordy25519 (approved):

PR #630, opened 2022-05-03
Multi-currency EVM gas fees

JasonTulp:
Adds eth-wallet payment asset preference
Adds withdraw_fee exchange


JasonTulp:
Moved to #635 we are no longer using fee_preferences in eth_wallet

PR #631, opened 2022-05-03
evm multicurrency precompiles

JasonTulp:
Adds precompiles for setting multicurrency fee preferences.
Connects to eth-wallet pallet and lets users set fee preferences through the EVM


jordy25519 (commented):
lgtm!

Issue #632, opened 2022-05-03
OffendingValidators storage persisted across eras


jordy25519:
`OffendingValidators` validators storage is being persisted across eras
This means any new offence is triggering a new era (whereas it would normally require 17% / 9 validator failures in a single era to trigger)

```js
// era 8359
> (await api.query.staking.offendingValidators()).toString()
[[5,true], [9,true], [10,false], [11,true], [13,false], [16,true], [19,false], [27,true], [30,true], [32,true], [39,true], [47,true], [49,true]]

// era 8340
// blockHash: 0x1cbccdf566266b3d76471d8c951fc28f39eec9b9221fccc6c69bea8e450b3106
> (await api.query.staking.offendingValidators.at('0x1cbccdf566266b3d76471d8c951fc28f39eec9b9221fccc6c69bea8e450b3106')).toString()
[[5,true], [9,true], [11,true], [16,true], [27,true], [30,true], [32,true], [39,true], [47,true], [49,true]]
```

<img width="1039" alt="Screen Shot 2022-05-04 at 11 43 14" src="https://user-images.githubusercontent.com/5133901/166597748-079ebadc-0cda-4267-bc44-4b193e252e59.png">

PR #634, opened 2022-05-05
Adjust inflation payout based on era duration

### LINKED ISSUE #393: inflation curve
use the inflation curve to accurately model target stake %, APY, and handle shorter era durations

jordy25519:
Related to missed payouts as part of #632 

Changes:
- Adjust inflation payouts based on era duration
- Forced era payouts are always scheduled to be paid
- Scheduling uses `on_idle` instead of predetermined block numbers

Other:
- spiked into using inflation curve but values aren't quite right. closes #393 as a result

TODO:
- [x] fix integration tests
- [ ] update rewards section in wiki


JasonTulp (approved):
lgtm!

runtime/tests/staking_reward.rs
```diff
@@ -285,96 +295,36 @@ fn era_transaction_fees_collected() {
...
 
 			// At the start of the next era, transaction rewards should be cleared
-			assert!(Rewards::calculate_total_reward().transaction_fees.is_zero());
+			assert!(Rewards::calculate_total_reward(era_duration_ms)
+				.transaction_fees
+				.is_zero());
 		});
 }
 
 #[test]
```
jordy25519, line 328: accruing doesn't happen anymore, payouts happen as normal

runtime/src/impls.rs
```diff
@@ -608,6 +609,7 @@ mod tests {
 	}
 
 	#[test]
+	#[ignore]
```
jordy25519, line 612: this test isn't really useful and adds ~2minutes to CI

JasonTulp (approved):
Looks good!

PR #639, opened 2022-05-10
Base64 encode the message used by eth wallet signning

KarishmaBothara:
This encode the eth call data before signing so that any eth wallet shows some meaningful data. 
Screen Shot 2022-05-11 at 9 54 55 AM

Payload from API side
```
let payload = cennznet.registry.createType('EthWalletCall', { call: call, nonce });
payload = `data:application-octet;base64,${base64Encode(payload.toHex())}`;
const signature = await ethereum.request({ method: 'personal_sign', params: [payload, address] });

```

### DIFF
--- a/crml/eth-wallet/src/lib.rs
+++ b/crml/eth-wallet/src/lib.rs
@@ -26,6 +26,9 @@ use sp_std::prelude::*;
 /// Ethereum-compatible signatures (eth_sign API call).
 pub mod ethereum;
 
+/// Prefix to be appended to payload before verifying signature
+const PREFIX_PAYLOAD_DATA: &str = "data:application/octet-stream;base64,";
+
 /// The module's configuration trait.
 pub trait Config: frame_system::Config {
 	/// The overarching event type.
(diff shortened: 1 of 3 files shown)


crml/eth-wallet/src/lib.rs
```diff
@@ -99,9 +99,10 @@ decl_module! {
 			// convert to CENNZnet address
 			let account = T::AddressMapping::into_account_id(eth_address);
 			let nonce = <frame_system::Pallet<T>>::account_nonce(account.clone());
-			let message = &(&call, nonce).encode()[..];
-
-			if let Some(_public_key) = ecrecover(&signature, message, &eth_address) {
+			let prefix = "data:application-octet;base64,";
```
jordy25519: why include this?
- should it be `application/octet-stream`?
- use a constant

ken-futureverse: ah yip, should be `data:application/octet-stream;base64,`

@holygits it's a convention when you base64 encode stuff, so the user know the type of the encoded message, also as a little bonus, if you cut n paste that in the browser, it will understand and decode the message too.

Do you reckon it adds too much of tx size overall?

jordy25519 (approved):
lgtm!

crml/eth-wallet/src/lib.rs
jordy25519: cool thing is the prefix is added on both sides (not sent) so it doesn't increase the tx size at all

ken-futureverse:
woop!! nicely done @KarishmaBothara !!

PR #648, opened 2022-05-18
Add gas fee for state oracle request

jordy25519:
Consume some specified gas before allowing state oracle request

Other:
- Share the scale_to_4dp function via crml/support (renamed to scale_wei_to_4dp)


JasonTulp (approved):
lgtm!

PR #649, opened 2022-05-18
Fix missing error metadata for some pallets

jordy25519:
These pallets were not creating error info in the metadata


crml/token-approvals/src/lib.rs
```diff
@@ -78,6 +78,7 @@ decl_storage! {
 decl_module! {
 	/// The module declaration.
 	pub struct Module<T: Config> for enum Call where origin: T::Origin {
+		type Error = Error<T>;
```
jordy25519, line 81: this one doesn't get called externally but included for consistency

PR #650, opened 2022-05-19
Ensure relayed returndata is 32 bytes

jordy25519:
Returndata must fit into 32 bytes or reported as invalid.
Restricting the byte length ensures costs are fixed and mitigates DoS potential


.github/workflows/coverage.yml
```diff
@@ -15,16 +15,19 @@ jobs:
     steps:
```
jordy25519, line 15: unrelated fix to coverage workflow

JasonTulp (approved):
lgtm!

PR #651, opened 2022-05-19
Add eth_call to EthereumRpcClient

jordy25519:
Adds eth_call RPC method to the EthereumRpcClient + some tests


JasonTulp (approved):
looks good!

PR #654, opened 2022-05-22
Make requests expire after ChallengePeriod blocks

jordy25519:
Ensure a request will expire if no response is submitted

Other:
- Store request input with the request (had separate storage before for optimisation but wasn't helpful in the end)


JasonTulp (approved):
lgtm!

PR #661, opened 2022-05-25
Add timestamp to state oracle callback

jordy25519:
Changes:
- state oracle callback now includes the block timestamp of the block where the request was fulfilled
- requires relayers submit the block timestamp aswell as block number

Other:
- Does a sensibility check on the submitted timestamp is +/- 30 seconds from the CENNZnet request timestamp


jordy25519:
check coverage

ken-futureverse:
sweet, let me know once this is one docker and i'll start adding in the timestamp

JasonTulp (approved):
lgtm!

PR #664, opened 2022-05-31
Add eth call notarization

jordy25519:
Adds:
- Adds process for eth bridge to make eth_calls and return the result (guts of the state oracle challenge protocol)

Other:
- LatestOrNumber using `u32` for no apparent reason, switched to `u64` to reduce all the casting
- REQUEST_TTL_MS down to 1.5 seconds from 2.5

TODO:
- [ ] tests
- handling the call result, slashing, issuing callbacks etc. is for a future PR


jordy25519:
superseded #667

PR #665, opened 2022-06-01
Relayer bond

JasonTulp:
Adds Relayer bond and unbond mechanics with constant bond amount
Changes call response so that it require a relayer to have a bond

### DIFF
--- a/crml/eth-state-oracle/src/lib.rs
+++ b/crml/eth-state-oracle/src/lib.rs
@@ -77,6 +77,8 @@ pub trait Config: frame_system::Config {
 		Balance = Balance,
 		FeeExchange = FeeExchange<AssetId, Balance>,
 	>;
+	/// Minimum bond amount for a relayer
+	type RelayerBondAmount: Get<Balance>;
 }
 
 decl_storage! {
@@ -88,7 +90,7 @@ decl_storage! {
 		/// Requests for remote 'eth_call's keyed by request Id
 		Requests get(fn requests): map hasher(twox_64_concat) RequestId => Option<CallRequest>;
 		/// Maps from account to balance bonded for relayer responses
-		RelayerBonds get(fn relayer_bonds): map hasher(twox_64_concat) T::AccountId => Option<Balance>;
+		RelayerBonds get(fn relayer_bonds): map hasher(twox_64_concat) T::AccountId => Balance;
 		/// Reported response details keyed by request Id
 		/// These are not necessarily valid until passed the challenge period
 		Responses get(fn responses): map hasher(twox_64_concat) RequestId => Option<CallResponse<T::AccountId>>;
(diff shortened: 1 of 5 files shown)


crml/eth-state-oracle/src/lib.rs
```diff
@@ -87,6 +87,8 @@ decl_storage! {
 		NextRequestId get(fn next_request_id): RequestId;
 		/// Requests for remote 'eth_call's keyed by request Id
 		Requests get(fn requests): map hasher(twox_64_concat) RequestId => Option<CallRequest>;
+		/// Maps from account to balance bonded for relayer responses
+		RelayerBonds get(fn relayer_bonds): map hasher(twox_64_concat) T::AccountId => Option<Balance>;
```
jordy25519: types like Balance have a sensible default `0` so don't really need to wrap in `Option`

crml/eth-state-oracle/src/lib.rs
```diff
@@ -241,6 +255,56 @@ decl_module! {
...
+			let origin = ensure_signed(origin)?;
+
+			// Check account doesn't already have a bond
+			if let Some(_bonded) = Self::relayer_bonds(&origin) {
+				// Account already has CPay bonded
+				return Err(Error::<T>::AlreadyBonded.into())
+			};
+
+			// check user has the requisite funds to make this bid
+			let balance = T::MultiCurrency::free_balance(&origin, T::MultiCurrency::staking_currency());
```
jordy25519: fee_currency

crml/eth-state-oracle/src/types.rs
```diff
@@ -18,6 +18,8 @@ pub use crml_support::{H160 as EthAddress, H256, U256};
 use scale_info::TypeInfo;
 use sp_std::prelude::*;
 
+pub const RELAYER_BOND_AMOUNT: Balance = 1_000_000_000;
```
jordy25519: instead of this use a  'runtime constant' so it can be configured by the runtime setup, see `ChallengePeriod` as an example

crml/eth-state-oracle/src/lib.rs
```diff
@@ -241,6 +257,55 @@ decl_module! {
...
+		#[weight = 500_000]
+		pub fn deposit_relayer_bond(origin) {
+			let origin = ensure_signed(origin)?;
+
+			// Check account doesn't already have a bond
+			if !Self::relayer_bonds(&origin).is_zero() {
+				// Account already has CPay bonded
+				return Err(Error::<T>::AlreadyBonded.into())
+			};
+
```
jordy25519: nit: these values tend to do a bit of work so best to put it into a local variable if planning to access more than once so its on the stack
```
T::RelayerBondAmount::get()
T::MultiCurrency::fee_currency()
```

jordy25519 (approved):

PR #668, opened 2022-06-02
Fix State Oracle Relayer Tests

JasonTulp:
Fixes tests that call submit_call_response


crml/eth-state-oracle/src/tests.rs
```diff
@@ -308,9 +318,12 @@ fn submit_call_response() {
 #[test]
 fn submit_call_response_request_should_exist() {
 	ExtBuilder::default().build().execute_with(|| {
+		let relayer: u64 = 1;
+		deposit_bond(relayer);
```
jordy25519, line 322: would be a good function to add to the ExtBuilder

jordy25519 (approved):

PR #671, opened 2022-06-03
Adding tests for offchain_try_eth_call

jordy25519:
try_eth_call considers delay between time a challenge is placed and when it is verified

Other:
- Add mock for setting specific timestamp
- removed block_hash from some mocks because it wasn't under test


JasonTulp (approved):
looks good!

Issue #676, opened 2022-06-09
clear invalidated events


jordy25519:
if an event claim is invalidated we should still flush it from PendingTxhashes (related #673)

PR #678, opened 2022-06-10
Some tidy up for relayer bonding

jordy25519:
the extra reserve checks is only if CENNZ is used (since it can have locks included in free balance)
+ the fix from develop for #676


JasonTulp (approved):
lgtm!

PR #687, opened 2022-06-16
Adds Controller account for NFT collections

JasonTulp:
Lets the collection owner add a controller account which can mint a series or additional tokens under a collection_id
This is useful for giving ERC721 contracts access to mint tokens in a series without passing ownership completely

### DIFF
--- a/crml/nft/src/lib.rs
+++ b/crml/nft/src/lib.rs
@@ -207,6 +207,8 @@ decl_storage! {
 	trait Store for Module<T: Config> as Nft {
 		/// Map from collection to owner address
 		pub CollectionOwner get(fn collection_owner): map hasher(twox_64_concat) CollectionId => Option<T::AccountId>;
+		/// Map from collection to controller address
+		pub CollectionController get(fn collection_controller): map hasher(twox_64_concat) CollectionId => Option<T::AccountId>;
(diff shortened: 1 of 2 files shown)


crml/nft/src/lib.rs
```diff
@@ -207,6 +207,8 @@ decl_storage! {
 	trait Store for Module<T: Config> as Nft {
 		/// Map from collection to owner address
 		pub CollectionOwner get(fn collection_owner): map hasher(twox_64_concat) CollectionId => Option<T::AccountId>;
+		/// Map from collection to controller address
+		pub CollectionController get(fn collection_controller): map hasher(twox_64_concat) CollectionId => Option<T::AccountId>;
```
jordy25519, line 211: would it be more useful to set controllers at the series level?

JasonTulp: I thought about this, however I think it makes more sense at a collection level as the goal is to allow for contracts to be able to act on a collection_owners behalf. One of the two functions introduced in the nft precompile is to mint a new series, so the controller account needs to have permission for this

jordy25519: conceptually a series should map to an erc721 contract which may have its own rules for minting otherwise need to make new collections each time

PR #701, opened 2022-07-11
Update rpc calls to latest substrate

JasonTulp:
Updated all RPC crates to use the latest version of substrate. This change moves to an updated json_rpsee crate


jordy25519 (approved):
lgtm 🦾

PR #706, opened 2022-07-19
Fix issue with sorting of council members affecting votes

JasonTulp:
Fixes issue where votes on existing proposals are changed when a new council member is added.


jordy25519 (approved):
looks good to me!

PR #716, opened 2023-01-15
Fix/rata local dev

justinfrevert:
This fixes the local docker compose to work with a recent cennznet image and separately updates the  Dockerfile to use a more recent debian image. This also alters the relationships in the dockerfile so the bootnode is alice instead of the RPC node. I haven't 100% confirmed whether an upstream change has definitively changed whether we need bootnodes and rpc nodes to be separate or something like this, but it's my current suspicion.

We still need to confirm that a runtime upgrade does not eventually brick with this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update && \
 ENV CARGO_HOME=/cennznet/.cargo
 RUN cargo build --release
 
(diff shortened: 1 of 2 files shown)


justinfrevert:
Trying to re-open to kickstart the github action

JasonTulp:
Closing to restart CI

